PR #81, opened 2026-06-30
Verifies ingest-on-startup and folded render on the openTUI surface

AmetAlvirde:
- Extracts the startup data path (--as-of → ingest → surface report → fold) out of app.ts's inline top-level code into a tested prepareStartup seam (mirrors the mountApp extraction in #67). app.ts behavior is preserved exactly.
- Adds startup.test.ts (Node): report wording, ingest+dedup surfacing, --as-of wiring, fail-loud paths.
- Adds smoke-startup-openTui.ts (pnpm smoke:startup, Bun): drives the real prepareStartup + mountApp through the openTUI test renderer against an on-disk store, asserting the five pnpm spine targets on the render surface — ingest+dedup counts, mutated current state, two as-of snapshots, tracer-parity fund value, and byte-identical restart survival.

## Review follow-ups (addressed)

- Temp-dir leak fix: `event-store.test.ts`, `report-fold.test.ts`, and `startup.test.ts` now `rm` their tracked `createdDirs` in `afterEach` instead of only resetting the array. Verified: after a clean run those prefixes leak zero dirs.
- Also applied the same cleanup to the pre-existing `review-file.test.ts` (outside this PR's original diff, but it leaked `numisma-review-*` dirs via the same pattern) — noted here per review.
- `startup.ts`: documented the deliberate graceful-degradation choice for an out-of-genesis-range `--as-of` on the app surface (the semantic guard stays in the engine fold per ADR-001 rather than being duplicated in the seam).
- `event-store.ts`: added a note that the atomic ingest is an O(n) rewrite (the price of rename-based crash-atomicity), not an O(1) `appendFile`.
- `fold.ts`: added a comment marking the latent first-position-wins t0/magnitude-guard divergence (unreachable today; one position per instrument).

Closes #76 
Closes #77 
Closes #78 
Closes #79 
Closes #80


packages/tui/src/event-store.test.ts
```diff
@@ -0,0 +1,409 @@
...
+}
+
+function markAapl(price: number, id = "mark-aapl") {
+  return { id, asOf: "2026-06-06", type: "PriceMarked", instrumentId: "aapl-usd", price };
+}
+
+const createdDirs: string[] = [];
+
+afterEach(() => {
+  createdDirs.length = 0;
```
AmetAlvirde, line 97: Temp dirs are tracked but never deleted — `afterEach` resets the array without `rm`-ing the directories. I confirmed it's live: `$TMPDIR` already holds 324 leaked `numisma-*` dirs. (`smoke-startup-openTui.ts` gets this right with `finally { rm(...) }`.)

```ts
afterEach(async () => {
  await Promise.all(createdDirs.map((dir) => rm(dir, { recursive: true, force: true })));
  createdDirs.length = 0;
});
```
`rm` is already imported here; the other two test files need the import added.

packages/tui/src/report-fold.test.ts
```diff
@@ -0,0 +1,138 @@
...
+      lots: [{ quantity: 1, cost: 100, tier: "c1" }],
+    },
+    decision: DECISION,
+  };
+}
+
+const createdDirs: string[] = [];
+
+afterEach(() => {
+  createdDirs.length = 0;
```
AmetAlvirde, line 93: Same temp-dir leak as in `event-store.test.ts`: `createdDirs` is populated but `afterEach` only resets the array, never removes the directories. Add `rm` to the import and `await Promise.all(createdDirs.map((dir) => rm(dir, { recursive: true, force: true })))` before resetting.

packages/tui/src/startup.test.ts
```diff
@@ -0,0 +1,191 @@
...
+      lots: [{ quantity: 1, cost: 100, tier: "c1" }],
+    },
+    decision: DECISION,
+  };
+}
+
+const createdDirs: string[] = [];
+
+afterEach(() => {
+  createdDirs.length = 0;
```
AmetAlvirde, line 86: Same temp-dir leak: `afterEach` resets `createdDirs` without deleting the dirs. Add `rm` to the import and `await Promise.all(createdDirs.map((dir) => rm(dir, { recursive: true, force: true })))` here.

packages/tui/src/startup.ts
```diff
@@ -0,0 +1,75 @@
...
+ * Throws on a malformed `--as-of` or a fail-loud ingest rejection so the host can
+ * exit non-zero with the durable log left unchanged — the report is emitted only
+ * after a successful ingest, so a rejected startup surfaces the error, not a count.
+ */
+export async function prepareStartup(
+  paths: EventStorePaths,
+  argv: string[],
+  deps: StartupDeps,
+): Promise<StartupPlan> {
+  const asOf = parseAsOfArg(argv);
```
AmetAlvirde, line 75: Low: `parseAsOfArg` validates only the `YYYY-MM-DD` shape; the `asOf < genesis.review.asOf` guard lives in `foldEvents`, reached lazily via the `loadData` thunk. So on the app path `pnpm dev --as-of 2020-01-01` emits the ingest report, enters the alternate screen, then renders mount-app's load-failure panel rather than exiting non-zero — inconsistent with the fail-loud framing (a *malformed* `--as-of` does exit before ingest). `pnpm spine`/`report` are fine since they await the fold in their try/catch. Either move the genesis-bound check earlier, or note the deliberate graceful-degradation choice for the app surface.

packages/tui/src/event-store.ts
```diff
@@ -0,0 +1,323 @@
...
+  return foldEvents(genesis, events, asOf);
+}
+
+/**
+ * Append events atomically: build the full next image of the log, write it to a
+ * sibling temp file, then `rename` over the log. rename(2) within a directory is
+ * atomic, so a crash mid-write leaves the prior log intact — a reader never sees a
+ * half-written or truncated final line.
+ */
+async function appendEvents(logPath: string, events: PortfolioEvent[]): Promise<void> {
```
AmetAlvirde, line 277: Nit: this reads the entire log, concatenates, and renames a full rewritten image on every ingest — deliberate and correct for crash-atomicity (rename(2)), but it's O(n) per ingest, not a true append. Worth a one-line note so a future reader doesn't read the name as an O(1) `appendFile`.

packages/engine/src/fold.ts
```diff
@@ -0,0 +1,649 @@
...
+  const latestMark = new Map<string, number>();
+  let usdMxn = genesis.review.usdMxn;
+  let latestAsOf = genesis.review.asOf;
+
+  // Seed a t0 baseline Close per genesis-held instrument so the FIRST PriceMarked
+  // already yields a 2-point journey (the journey builder skips single anchors).
+  // The anchor is the genesis `markPrice`, NOT cost: that keeps it coherent with
+  // the authoritative valuation, so no synthetic `markprice-close-mismatch` fires.
+  // Genesis-provided closes win — we only fill instruments that have none.
+  const seededInstruments = new Set(closes.map((close) => close.instrumentId));
```
AmetAlvirde, line 562: Nit (latent, not currently reachable): if two genesis positions share an instrument with *different* `markPrice`, this seeds the t0 close from the first position, while `buildEventReference`'s `noteClose` keeps the last (equal `asOf` overwrites) — so the fold's t0 anchor and the magnitude-guard's seed can diverge in that corner. Harmless today; worth a comment if multi-position-per-instrument ever becomes valid.

AmetAlvirde (commented):
Reviewed the full diff; suite passes (236 tests) and typecheck is clean. The persistence spine is cohesive and the ADR-001 boundary is respected. No correctness bugs on the production path — the idempotent append-then-archive recovery and the byte-for-byte "log unchanged on rejection" assertions are exactly right.

One real cleanup (the temp-dir leak, inline) plus a few low-severity notes below. Good to merge once the temp-dir cleanup is wired up.

PR #133, opened 2026-07-24
Harden durable-log ingest capture against a stale allowlist (#132)

AmetAlvirde:
Fixes the silent-failure class behind issue #132: the in-process durable-log capture skipped **every ingest since deploy** because a half-finished `checkpoint.json → head-digest.json` rename left accumulus's allowlist ignoring the real `head-digest.json`, so the single atomic `git add` aborted wholesale on the only-ignored pathspec — dropping the durable event log with it. ~17 days of real marks sat uncommitted, one `git reset --hard` from loss.

## What's in this PR (numisma, code only — no fund data)

- **`ingest-commit.ts` — per-file resilient staging (Q2):** each durable file gets its own `git add`; a file that refuses to stage warns loudly *by name* and is skipped, but the rest — the log above all — still commit. Keeps the allowlist-only guard (never `git add -A`).
- **`run-daily-fetch.sh` — post-check (Q3):** after `pnpm spine` + the step-3 backstop, assert the data-dir tree is clean. **Strict on the log** (dirty ⇒ non-zero exit ⇒ launchd shows a red job — the miss can never again be an unread stderr warning); **lenient on `head-digest.json`** (forensic breadcrumb: warn, don't fail).
- **Regression tests (Q4):** two cases over inline allowlist repos mirroring accumulus's strategy (decoupled — no read of the sibling repo). One locks the fixed state (head-digest lands); one reproduces the outage (head-digest only-ignored) and asserts the log still commits — **verified to go red on the old atomic-add code**.
- **Docs:** step-4 post-check documented in `docs/price-feed-ops.md`.

## Companion fix (accumulus data repo — the correctness gate, Q1)

Committed separately in `~/Dev/accumulus` (`70e8a26`), **not yet pushed** — carries no fund data into numisma:
- `.gitignore`: add `!/data/head-digest.json`, drop the stale `!/data/checkpoint.json`.
- `git rm --cached data/checkpoint.json` + remove the fossil from disk. `git ls-files data/` now reads exactly the four durable files.

## Verification

- Full suite green: **554 passed / 11 skipped**.
- New regression test confirmed **red on the old atomic-add code**, green with per-file staging.
- Q5 (shell backstop) kept as defense-in-depth; unchanged.

## Reliability review

Requested on this PR — focus on the silent-failure class (unattended job, unread stderr) and the "never lose the log" contract.

Closes #132

### DIFF
--- a/ops/price-feed/run-daily-fetch.sh
+++ b/ops/price-feed/run-daily-fetch.sh
@@ -104,4 +104,30 @@ fi
 #    surfaced to the scheduler too.
 echo "[$STAMP] prices:fetch clean — ingesting marks via pnpm spine"
 pnpm spine
+
+# 3) Persist the appended marks to the private data repo. `pnpm spine` appends the
+#    day's marks to events.jsonl but leaves that change UNCOMMITTED — a stray
+#    reset/checkout would silently lose real fund data (exactly how a multi-day
+#    backlog once piled up in the working tree). Commit the tracked data changes so
+#    the durable log is actually durable. Scoped to the data dir (never sweeps
+#    unrelated repo edits), idempotent (a no-new-marks run commits nothing), and it
+#    NEVER pushes — pushing to the remote stays a manual, reviewed step.
+DATA_DIR="${NUMISMA_DATA_DIR:-}"
+if [[ -z "$DATA_DIR" ]]; then
+  echo "[$STAMP] NUMISMA_DATA_DIR unset — skipping data-repo commit (durable log left uncommitted)."
+elif ! git -C "$DATA_DIR" rev-parse --git-dir >/dev/null 2>&1; then
+  echo "[$STAMP] WARNING: $DATA_DIR is not inside a git repo — skipping commit (durable log left uncommitted)."
+else
+  git -C "$DATA_DIR" add -u -- .
+  if git -C "$DATA_DIR" diff --cached --quiet -- .; then
+    echo "[$STAMP] no tracked data changes to commit (idempotent no-op run)."
+  else
+    git -C "$DATA_DIR" commit -q \
+      -m "data: daily price marks $STAMP" \
+      -m "Auto-committed by run-daily-fetch.sh after spine ingest. Not pushed (push is a manual, reviewed step)." \
+      -- .
+    echo "[$STAMP] committed durable-log changes (not pushed)."
+  fi
+fi
+
 echo "[$STAMP] price-feed daily run complete."
--- a/ops/price-feed/run-daily-fetch.sh
+++ b/ops/price-feed/run-daily-fetch.sh
@@ -112,13 +112,30 @@ pnpm spine
 #    the durable log is actually durable. Scoped to the data dir (never sweeps
 #    unrelated repo edits), idempotent (a no-new-marks run commits nothing), and it
 #    NEVER pushes — pushing to the remote stays a manual, reviewed step.
-DATA_DIR="${NUMISMA_DATA_DIR:-}"
-if [[ -z "$DATA_DIR" ]]; then
-  echo "[$STAMP] NUMISMA_DATA_DIR unset — skipping data-repo commit (durable log left uncommitted)."
-elif ! git -C "$DATA_DIR" rev-parse --git-dir >/dev/null 2>&1; then
+#
+#    DATA_DIR defaults to the SAME tree the in-process capture writes to
+#    (apps/tui/src/ingest-commit.ts derives ~/Dev/accumulus/data from os.homedir()
+#    when NUMISMA_DATA_DIR is unset), so this backstop AND the step-4 post-check
+#    always target the dir actually written to — never a no-op that leaves a failed
+#    capture uncommitted while the job stays green.
+#
+#    `git add -u` restages tracked modifications only, so a FIRST-TIME untracked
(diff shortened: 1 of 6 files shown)


AmetAlvirde (commented):
Reliability review (focus: silent-failure class + "never lose the log"). The core per-file `git add` change is sound and the regression tests genuinely reproduce #132. Three inline notes below; the first is the one I'd want resolved before merge.

ops/price-feed/run-daily-fetch.sh
```diff
@@ -104,4 +104,57 @@ fi
...
+
+# 3) Persist the appended marks to the private data repo. `pnpm spine` appends the
+#    day's marks to events.jsonl but leaves that change UNCOMMITTED — a stray
+#    reset/checkout would silently lose real fund data (exactly how a multi-day
+#    backlog once piled up in the working tree). Commit the tracked data changes so
+#    the durable log is actually durable. Scoped to the data dir (never sweeps
+#    unrelated repo edits), idempotent (a no-new-marks run commits nothing), and it
+#    NEVER pushes — pushing to the remote stays a manual, reviewed step.
+DATA_DIR="${NUMISMA_DATA_DIR:-}"
+if [[ -z "$DATA_DIR" ]]; then
```
AmetAlvirde: **Medium — the strict "red the job" guard is gated on the same env var whose absence silently leaves the log uncommitted.**

Both step 3 (line 116) and the step-4 post-check (line 142) gate on `NUMISMA_DATA_DIR`. When it's unset, both no-op and the script falls through to `exit 0` — while the in-process capture uses the *code default* (`~/Dev/accumulus/data`). So in the unset case:

- Happy path: in-process capture commits to the code-default dir; job green. Fine.
- **#132-class path: in-process capture fails → step 3 skips → post-check skips → job exits 0.** That's exactly the silent-green miss this PR's contract ("only a FAILED job reaches the operator") is built to eliminate, re-entering through the unset path.

Since #132 was itself config drift causing silent loss, a config-dependent guard against config-drift silent loss is a real hole. The doc change frames unset as merely "log left uncommitted" but never says the post-check is also disabled and the job stays green.

Suggest defaulting the shell's `DATA_DIR` to the same `~/Dev/accumulus/data` the in-process path uses (so the post-check verifies the tree actually written), or hard-failing when `NUMISMA_DATA_DIR` is unset.

ops/price-feed/run-daily-fetch.sh
```diff
@@ -104,4 +104,57 @@ fi
...
+#    (the in-process capture skipping every run, issue #132) was invisible because a
+#    launchd job's stderr goes to an UNREAD log — a "loud warning" reaches no one;
+#    only a FAILED job reaches the operator. So after the in-process capture AND the
+#    step-3 backstop have both run, verify the data-dir working tree is clean and
+#    turn the job RED if it is not. Split by ADR stance: the event log is the
+#    source-of-truth (STRICT — dirty ⇒ non-zero exit), head-digest.json is a forensic
+#    breadcrumb (LENIENT — warn only). Paths are relative to $DATA_DIR, which is the
+#    accumulus `data/` subdir holding the durable files directly.
+if [[ -n "$DATA_DIR" ]] && git -C "$DATA_DIR" rev-parse --git-dir >/dev/null 2>&1; then
+  DIGEST_DIRTY="$(git -C "$DATA_DIR" status --porcelain -- head-digest.json)"
```
AmetAlvirde: **Low — this lenient warning is dead in the exact drift scenario it documents.**

In the #132 drift shape, `head-digest.json` is *only-ignored*. Verified locally: `git status --porcelain -- <ignored-file>` returns empty even with an explicit pathspec (it needs `--ignored`). So when head-digest lags *because it's ignored* (the actual outage cause), `DIGEST_DIRTY` is empty and this warning never fires — it only fires when head-digest is tracked-and-modified, a different and less likely failure. Harmless (it's lenient by design), but the comment oversells what it catches.

ops/price-feed/run-daily-fetch.sh
```diff
@@ -104,4 +104,57 @@ fi
...
+#    the durable log is actually durable. Scoped to the data dir (never sweeps
+#    unrelated repo edits), idempotent (a no-new-marks run commits nothing), and it
+#    NEVER pushes — pushing to the remote stays a manual, reviewed step.
+DATA_DIR="${NUMISMA_DATA_DIR:-}"
+if [[ -z "$DATA_DIR" ]]; then
+  echo "[$STAMP] NUMISMA_DATA_DIR unset — skipping data-repo commit (durable log left uncommitted)."
+elif ! git -C "$DATA_DIR" rev-parse --git-dir >/dev/null 2>&1; then
+  echo "[$STAMP] WARNING: $DATA_DIR is not inside a git repo — skipping commit (durable log left uncommitted)."
+else
+  git -C "$DATA_DIR" add -u -- .
```
AmetAlvirde, line 133: **Minor (non-blocking)** — `git add -u` stages tracked modifications only, whereas the in-process capture uses `git add <file>` and also stages untracked. If a durable file were ever newly *untracked* (e.g. a first-time `genesis.json`) and the in-process capture had failed, this backstop can't stage it and the post-check would then FATAL on a file step 3 structurally cannot recover. Realistically these files are committed once at fund creation, so edge-case only — a one-line awareness comment would suffice.

PR #284, opened 2026-08-11
fix(tui): the module-graph walk reads code, not the prose around it

AmetAlvirde:
`apps/tui/src/module-graph.test.ts` walks the TUI's module graph by scraping import specifiers out of raw source text, and its matcher reads `from "…"` anywhere in the file. So an ordinary English sentence in a doc comment — `a load failure is a different fact from "this position has no plan."` — was walked as an import of a package literally named `this position has no plan.` and failed the build as an unresolvable external. It hit twice in one increment (#267), in unrelated files whose source was correct; both were worked around by rewording the sentence.

## What changed

**The input, not the matcher.** Comments are stripped before the walk; the `from "…"` matcher stays exactly as greedy as it was. The other candidate fix — anchoring to statement position (`^\s*(import|export)\b … from`) — buys the same quiet by narrowing what counts as an import, and an import this walker stops seeing is precisely the edge the prototype hid behind a computed specifier. Narrowing the input costs the guard nothing; narrowing the matcher costs it reach.

The stripper keeps string literals verbatim (the specifier is one — which is also why `//` inside a URL cannot start a comment), preserves newlines so a removed comment cannot splice two statements together, and ends a `'`/`"` string at the newline so a stray quote inside a regex literal like `/["']/` cannot swallow the rest of the file.

The `pg` half of the first assertion carried the same false positive — it matched raw source, so a comment *promising* there is no `pg` import read as one. It now reads the same stripped source the walk does.

## What holds it

Five fixture tests over a throwaway temp module tree, in both directions:

- prose saying `from "…"` in a line comment, a block comment and a JSDoc block is prose — and the real import sitting beside it is still followed;
- a comment mentioning `from "pg"` is not a `pg` import;
- an unresolvable relative import and a real external package are **both still reported** — the guard is not merely quieter;
- a multi-line import, a re-export and a dynamic `import()` inside a function body are all still followed, which is what a statement-position matcher would have dropped;
- a `//` inside a string literal does not start a comment.

Also, per the issue's third acceptance point, the engine's `PlanLookup` docstring goes back to its natural phrasing (`is a different fact from "this position has no plan."`) as the live regression fixture — verified to fail the old walker and pass the new one.

`pnpm typecheck` clean, `pnpm test` green (1692 passed).

What actually closes #273 is the stripped input plus the five fixtures that lock the guard's reach.

## Review round (a438ce9)

Two findings, both about the scanner rather than the walk, both fixed:

1. **The scanner had no regex-literal state**, so a `/` inside a regex could open a comment — which loses *code*, where the original bug only invented prose. `/^\/*$/` opens a block comment running to EOF, the file yields no imports at all, and its whole subtree drops out of the graph: a `pg` edge underneath would pass in silence. `/https:\/\//` is the line-comment half. Both shapes exist in the repo today (three test files), harmless only because no import shares those lines. The scanner now has the third state, and a wrong call is survivable in one direction only — a literal is emitted verbatim either way, so the cost is a comment left standing (loud) rather than a statement deleted (silent).
2. **The string-literal test passed with the string branch deleted** — the URL sat on a line holding no import. The import now shares that line, which is the only arrangement that locks the claim.

Two more fixture tests: the two regex shapes, and division staying division (`half / 5; // a ratio read from "prose"` must not swallow the comment and hand the walk its prose).

Checked against the TypeScript parser across all 266 TS/TSX files in the repo: every specifier `ts.preProcessFile` reports, the stripped walk still finds. `pnpm typecheck` clean, `pnpm test` green (1694 passed).

### DIFF
--- a/apps/tui/src/module-graph.test.ts
+++ b/apps/tui/src/module-graph.test.ts
@@ -15,10 +15,11 @@
  * a computed specifier that this walker cannot resolve shows up as an unresolved
  * import and fails the last assertion rather than passing silently.
  */
-import { readFileSync, readdirSync, existsSync } from "node:fs";
-import { dirname, join, relative, resolve } from "node:path";
+import { readFileSync, readdirSync, existsSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { basename, dirname, join, relative, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
-import { describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it } from "vitest";
 
 const HERE = dirname(fileURLToPath(import.meta.url));
 // HERE = apps/tui/src → the repo root is three levels up.
@@ -81,6 +82,84 @@ function resolveSpecifier(specifier: string, fromFile: string): string | undefin
   return resolve(pkg.dir, target);
 }
 
+/**
+ * The same source with every comment removed — the walker's real input.
+ *
+ * WHY THE INPUT AND NOT THE MATCHER. The matcher below reads `from "…"` ANYWHERE in
+ * the text, so an ordinary English sentence in a doc comment — `a load failure is a
+ * different fact from "this position has no plan."` — was walked as an import of a
+ * package by that name and failed the graph as an unresolvable external (#273). The
+ * other candidate fix, anchoring the matcher to statement position, buys the same
+ * quiet by narrowing what counts as an import — and an import this walker stops
+ * seeing is exactly the edge the prototype hid. A comment cannot contain an import,
+ * so the comments go and the matcher stays as greedy as it was.
+ *
+ * STRING LITERALS ARE KEPT VERBATIM, since the specifier itself is one — and that is
+ * why `//` inside a string cannot start a comment here. Newlines survive so a removed
+ * comment cannot splice two statements together. A `'`/`"` string also ends at the
+ * newline: unterminated ones do not exist in source that compiles, and the bound
+ * keeps a stray quote inside a regex literal (`/["']/`) from swallowing the file.
+ */
+function stripComments(source: string): string {
+  let out = "";
+  let index = 0;
+  while (index < source.length) {
+    const char = source[index] as string;
+    const next = source[index + 1];
+    if (char === "/" && next === "/") {
+      while (index < source.length && source[index] !== "\n") {
+        index += 1;
+      }
+      continue;
+    }
+    if (char === "/" && next === "*") {
+      index += 2;
+      while (index < source.length && !(source[index] === "*" && source[index + 1] === "/")) {
+        if (source[index] === "\n") {
+          out += "\n";
+        }
+        index += 1;
+      }
+      index += 2;
+      continue;
+    }
+    if (char === '"' || char === "'" || char === "`") {
+      out += char;
+      index += 1;
+      while (index < source.length) {
+        const inner = source[index] as string;
+        if (inner === "\\") {
+          out += source.slice(index, index + 2);
+          index += 2;
+          continue;
+        }
+        out += inner;
+        index += 1;
+        if (inner === char || (inner === "\n" && char !== "`")) {
+          break;
+        }
+      }
+      continue;
+    }
+    out += char;
+    index += 1;
+  }
+  return out;
+}
+
+const STRIPPED = new Map<string, string>();
+
+/** A file's code with its comments gone — read once, stripped once. */
+function codeOf(file: string): string {
+  const cached = STRIPPED.get(file);
+  if (cached !== undefined) {
+    return cached;
+  }
+  const code = stripComments(readFileSync(file, "utf8"));
+  STRIPPED.set(file, code);
+  return code;
+}
+
 interface Graph {
   files: Set<string>;
   externals: Set<string>;
@@ -103,7 +182,7 @@ function walkGraph(entries: readonly string[]): Graph {
       continue;
     }
     graph.files.add(file);
-    const source = readFileSync(file, "utf8");
+    const source = codeOf(file);
     const specifiers = [
       ...[...source.matchAll(/from\s*["']([^"']+)["']/g)].map((m) => m[1] as string),
       ...[...source.matchAll(/import\s*\(\s*["']([^"']+)["']\s*\)/g)].map((m) => m[1] as string),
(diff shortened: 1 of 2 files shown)


apps/tui/src/module-graph.test.ts
```diff
@@ -81,6 +82,84 @@ function resolveSpecifier(specifier: string, fromFile: string): string | undefin
...
+  while (index < source.length) {
+    const char = source[index] as string;
+    const next = source[index + 1];
+    if (char === "/" && next === "/") {
+      while (index < source.length && source[index] !== "\n") {
+        index += 1;
+      }
+      continue;
+    }
+    if (char === "/" && next === "*") {
```
AmetAlvirde, line 168: **The scanner carries no regex-literal state, so a `/` inside a regex can open a comment and delete real code — which makes the walk blind exactly where this guard is supposed to see.**

Input: a file whose first line is `const LEADING = /^\/*$/;`, with its imports below. At the escaped slash the scanner is still in code state, reads `/` followed by `*`, and opens a block comment that runs to the next `*/` — here, EOF. `stripComments` returns `"const LEADING = /^\\\n\n\n"`, the walk finds **zero** specifiers in that file, and every module below it drops out of `GRAPH.files`. A `pg` edge underneath it passes the guard silently — the "quieter by blinder" outcome the PR body argues against.

The line-comment branch just above has the same hole one character earlier: any regex whose closing `/` follows an escaped `\/` (`/\//`, `/https:\/\//`) reads as `//` and deletes the rest of its line. That exact shape is in the repo today — `apps/tui/src/ingest-commit-hardening.test.ts:176` (`/\?\? prices\//`), `apps/web/src/plans-import-guard.test.ts:76` and `apps/web/src/push/orders-stay-off-the-wire.test.ts:89` (`/\/\*[\s\S]*?\*\//g`) — harmless only because no import happens to share those lines. Note the last two are sibling import-guards that strip comments themselves, so this construct is likely to migrate into the walked set.

The docstring above does reason about regex literals, but only covers the *quote* case (`/["']/`), which the newline bound genuinely handles. The *slash* case it does not, and that one deletes rather than merely retains. A third-state check (a `/` in expression position starts a regex, consume to the unescaped closing `/`) closes both, or — cheaper — treat `/` + `*` and `/` + `/` as a comment only when the preceding non-space character is not one that can end an expression.

apps/tui/src/module-graph.test.ts
```diff
@@ -204,3 +285,115 @@ describe("the TUI's module graph", () => {
...
+      ].join("\n"),
+      "multi-line.ts": "export const a = 1;\nexport const b = 2;\n",
+      "re-exported.ts": "export const c = 3;\n",
+      "dynamic.ts": "export const d = 4;\n",
+    });
+    expect(graph.unresolved).toEqual([]);
+    expect(reachedNames(graph)).toEqual(["dynamic.ts", "entry.ts", "multi-line.ts", "re-exported.ts"]);
+  });
+
+  it("a `//` inside a string literal does not start a comment", () => {
```
AmetAlvirde, line 461: **This test passes with the entire string-literal branch of `stripComments` deleted, so it does not lock the behaviour its name asserts.**

I ran this exact fixture through a copy of `stripComments` with the `if (char === '"' || char === "'" || char === "`")` block removed. Line 1 becomes `export const docs = "https:` — the `//` in the URL does eat the rest of the line, but no import lives on that line, so nothing is lost. The only thing that actually fails a string-blind stripper here is line 2's block comment, and the block branch already handles that independently. Result: `./thing.js` is still found, `reachedNames` is still `["entry.ts", "thing.ts"]`, test green.

Moving the import onto the same line as the URL makes the fixture do what its name says. Verified:

```
export const docs = "https://example.com/a//b"; import { thing } from "./thing.js";

PR stripper  -> 'export const docs = "https://example.com/a//b"; import { thing } from "./thing.js";'  -> ['./thing.js']
string-blind -> 'export const docs = "https:'                                                        -> []
```

Worth fixing because string handling is the riskiest part of this scanner — it is what makes a quote inside a regex literal swallow to end-of-line — so leaving it unlocked means a later "simplification" that drops it goes unnoticed.

AmetAlvirde (commented):
Two findings on `stripComments`, both verified by running the scanner. Context for what I checked and did **not** find: I replayed the walk against the TypeScript parser over all 82 files it currently reaches — every specifier `tsc` sees, the stripped walk also sees, and no real specifier is dropped in any of the repo's 265 TS/TSX files. I also confirmed the restored `plans.ts` docstring is a genuine live fixture (the pre-fix walker picks up `this position has no plan.` as an external against the current tree), that fixtures 1 and 2 fail against the pre-fix walker, and that the `afterEach` cleanup leaks nothing (registered at file scope, runs on failure, dir is registered before any write can throw). So both findings below are latent, not live.

apps/tui/src/module-graph.test.ts
AmetAlvirde: Fixed in a438ce9 — confirmed the shape first: `/^\/*$/` above an import made the fixture file yield zero specifiers and `thing.ts` vanished from `reachedNames`, exactly as described.

The scanner now has the third state. A `/` in expression position opens a regex literal and is consumed to its unescaped closing slash, with `[…]` classes tracked so a `/` inside one does not close it early; `//` and `/*` are still read first, since neither can open a regex. `regexMayFollow` decides from the last significant character of *code* (comments do not feed it): `)`, `]` and identifiers/numbers divide, the fourteen keywords that can precede a literal do not.

Both halves of the hole are now locked by a fixture carrying `/^\/*$/` and `/https:\/\//` — the second sharing its line with the import, so the line-comment case cannot pass vacuously — and a companion test asserts division is still division: `half / 5; // a ratio read from "prose"` must not swallow the comment and hand the walk its prose, which would put #273 straight back.

On the direction of a wrong call, which the docstring now states: a literal is emitted verbatim either way, so misjudging division-as-regex leaves a comment standing (loud — #273 returns) while misjudging regex-as-division deletes a statement (silent). Only the second is unacceptable, and it is the one the state removes.

Verified repo-wide against the TypeScript parser: for all 266 TS/TSX files, every specifier `ts.preProcessFile` reports is still found after stripping. Separately diffed the scanner with and without the new state — 4 files strip differently, the three you named plus `apps/web/src/projection/provision.test.ts`, all test files outside the walked set, which matches your latent-not-live read.

apps/tui/src/module-graph.test.ts
AmetAlvirde: Fixed in a438ce9 — you are right that it locked nothing. The import now shares the URL's line:

```ts
export const docs = "https://example.com/a//b"; import { thing } from "./thing.js";
```

so a string-blind stripper eats the import along with the rest of the line and the test fails, instead of losing only a `docs` binding nobody asserted on. The block-comment line no longer carries the import either, so the two branches are not covering for each other. The comment above the fixture now says why the arrangement matters, since the obvious tidy-up is to put the import back on its own line.

PR #308, opened 2026-08-12
Refine the fill path to reliable: M1–M7, and the day-zero projection named

AmetAlvirde:
Implements spec #302 — the fill-path refinement increment. The six-commit chart
spike on this branch replaced the hand-rolled SVG with a `@tanstack/charts` Price
Drop Path and, along the way, built the day-zero projection. The audit bounded the
gap to reliable at seven must-fix findings (M1–M7), all confined to the web render
surface. This closes that gap and gives the day-zero projection a name.

This PR is what actually closes #303, closes #304, closes #305, closes #306 and
closes #307, and it also closes #302 — one PR for the whole spec, no per-slice
merge.

## The five slices

**Slice A (#303) — the chart's quantitative logic becomes a tested pure module.**
`cumulate`, `withRadius`, `splitAt`, `spotMarkFor` and the byte-duplicated
compact-USD formatter move out of a `.tsx` the coverage glob cannot see into
`apps/web/src/ladder/price-drop-path.ts`, now at 100% across statements, branches,
functions and lines. `FillPathRungView` gains `filled: boolean`, decided once by
`venueFilled()`; `venueAxis` is optional on that contract, so the undefined arm is
documented as `false` — `true` is only ever the venue's positive statement, never
inferred from a gap. T1–T4 landed; because their first red was module-absence red,
a 10-mutant check on the module and a 2-mutant check on `filled` were run and
recorded. No visual change.

**Slice B (#304) — a fixture surface for every started-ladder state.** No branch of
"a ladder that has started" had ever been rendered by anyone: the dev surface reads
real data through the session gate, live data has zero filled rungs, and the ladder
route had no fixture path. Four hand-authored ladders (`day-zero`,
`partly-walked`, `out-of-order`, `overfilled`) now render the real cards through the
real composition on a dev-only route. `/ladder/$planId` is untouched — branching
`getDashboard()` would put an `if` inside the session gate. Outside
`import.meta.env.DEV` the loader returns `disabled`, and the fixtures are reached by
dynamic import inside that branch, so dead-code elimination drops the module and
every invented value from a production bundle; both halves are asserted. Every
fixture value is hand-authored — nothing is derived from real trade output.

**Slice C (#305) — the `Deployed` rule gets the chart's out-of-domain contract.**
The rule plotted measured `deployed` on an axis whose domain is max cumulative
declared, and the scale did not clamp: an overfill drew at 130% of plot width,
unmarked, while twenty lines up the same chart refused exactly that failure for spot.
Now `deployedMarkFor` in the pure module decides it — clamp-and-mark, the
`spotMarkFor` family. Refusal was the other allowed arm and loses: drawing nothing
deletes the one reading an overfilled ladder most needs, leaving a picture
indistinguishable from a ladder walked exactly to plan. A mutant that refuses
instead is red, so the ruling is checked rather than preferred. Second half: the rule
moved off `--pos` onto the chart's neutral ink, with no legend swatch, so picture and
list keep exactly three state colours.

**Slice D (#306) — rung state branches on wire facts, never on engine copy.** The
engine's `label()` authored display copy inside the domain package and the web
branched on the literal at two sites; rewording the engine string broke the web
silently with `tsc` green and every test green. Copy is now authored web-side by
`rung-state-copy.ts`, and the guard is the input type: `RungStateFacts` carries no
string, so the author is never handed the engine's words and a caller who passes them
fails to compile. The engine's `label` still rides the wire untouched (C1) but is read
nowhere in `apps/web`. The seam test has two halves because with D1 deferred nothing
can render `Pills`: rewording every wire label across all four fixtures must leave the
composed view identical, and no production file may ask a question of `stateCopy`.
Rewording the engine's `"waiting"` leaves all 434 web tests green while engine tests
go red — the property the slice exists to establish.

**Slice E (#307) — the truth sweep and the ruled caption wording.** M7: the convexity
caption now says its figure is the capital *still waiting*, because the caption names
the waiting total while the x axis tops out at the whole declared total — identical on
day zero, divergent on the first fill. Prose carries the qualifier; the axis top is
untouched. M4: both uncovered `.tsx` surfaces are named in the coverage ledger, with
T7 recorded as a still-open gap rather than left silent. M5.1–M5.4: the four stale
claims are gone, including a second site of the "no chart library" claim in
`DcaCard.tsx` that the slice did not name and grep found. Both docs homes land in
`docs/ladder-fill-path.md` — the day-zero projection with its four structural
guarantees stated as load-bearing so nobody later "fixes" it, and the palette rule
generalized: only states get hue and a legend entry, annotations get neutral ink, and
a genuine fourth rung state is an ADR, not a colour pick.

## Verification

- `pnpm typecheck` — exit 0, all six packages.
- `pnpm test` — 1913 passed / 19 skipped (from 1875 before the increment).
- `pnpm coverage` — `price-drop-path.ts` and `rung-state-copy.ts` both 100/100/100/100.
- Every slice was rendered in a browser against the fixture surface, not only typechecked.
- Greps that must come back empty or single: the filled predicate has one authoring
  site; no `"waiting"` comparison survives in `apps/web`; both `.tsx` surfaces are in
  the ledger; none of M5's four stale claims remain.
- C1 held throughout: `projection/contract.ts`, `DcaWireRung` and `packages/` have no
  diff. The day-zero projection's own code is diff-free, proven per-symbol.

## Two items for a human, neither blocking

1. **The duplicated `partly filled · 40%`.** The fixture surface rendered a
   partly-filled rung for the first time and exposed two duplications — the inspect
   card's state pill beside its percent pill, and the row's substatus beside its qual
   pill. Authoring copy from facts dissolves neither. Killing either needs a ruling on
   which element wins, which is a copy decision a refactor should not make under cover.
2. **#285 §6.3b** still carries an illustrative caption matching neither the shipped
   nor the M7 wording. It is hedged as `e.g.`/synthesized, so it may be fine; a
   tracker write is a human's call.

## Deferred, with the audit's rationale intact

D1 (the component-test harness — and with it T7, the a11y invariant, held by ADR-019
meanwhile), D2 (`color-mix()` guard), D4, D6, D7. O1 — finishing `chartFor`'s
demolition — remains slice 3 of #285, which also carries this spec's two added
constraints (D3's ≥2-distinct-prices gate, Seam B's `spotIsLive` promotion).
O2 and O3 were ruled keep-as-shipped and ship no change.


AmetAlvirde (commented):
## Review — spec #302, the seven commits on `19af544`

Ran everything before writing anything:

- `pnpm typecheck` — exit 0, all six packages.
- `pnpm test` — **1913 passed / 19 skipped**, matching the PR body exactly.
- `pnpm coverage` (json-summary, per file) — `ladder/price-drop-path.ts` **100/100/100/100**, `ladder/rung-state-copy.ts` **100/100/100/100**. Claim confirmed as stated.
- `pnpm --filter @numisma/web build` (real Vercel preset build) — exit 0, and the bundle inspected byte-wise, see below.
- All four fixture routes rendered off a dev server on port 4319 (started and shut down; the pre-existing 3000/3001 listeners were left alone), and the SSR SVG read mark-by-mark rather than eyeballed.

### The seven claims, checked

**1. No visual change in slice A — holds.** The 221 lines the component lost are the five functions and the formatter, moved verbatim. `AXIS_USD`/`RANGE_USD` → `COMPACT_USD` is option-for-option identical (`en-US`, `currency`/`USD`, `compact`, `maximumFractionDigits: 1`), and `isFilled(rung)` → `rung.filled` is the same literal comparison one hop earlier (`venueFilled`). No mark, no legend entry, no `aria-hidden`/`tabIndex`/`focus`/`pointer`/`keyboard`/`tooltip` line moved in `0da95e6`. Judged by code equivalence, not by a pixel diff.

**2. The production-bundle claim — holds, and stronger than claimed.** Built and grepped. `facade00`, `fixture-rung-`, `day-zero`, `partly-walked` and the ladder's prices are absent from **both** the client assets and `__server.func/_ssr`, and no fixtures chunk is emitted at all. The loader constant-folds to literally `loader: async ({params: n}) => ({status: "disabled"})` in `static/assets/index-*.js` — the `import()` and everything behind it are gone, not merely unreachable. The only thing that still ships is the route component's own banner text (`"Synthesized data, authored in ladder/started-ladder.fixtures.ts"`), which is what the disabled page is *for*.

**3. `deployedMarkFor` covers every arm — holds.** In-domain, over-`xEnd`, under-`xStart`, absent, and exactly-on-`xEnd` are each asserted, plus the range property at `price-drop-path.test.ts:362`. The clamp is taken against `xScale.domain()[1]` *after* `.nice()`, i.e. against the domain the axis actually drew — verified in the rendered `overfilled` SVG: `rule-x-4:number:4500` at `x1=316` (the plot's right edge) under `Deployed ~$5.5K →` at `x=312,y=16`, `text-anchor="end"`, while the spot rule pins to the bottom at `y=216`. The corner-collision the module's header reasons about does not occur on any of the four fixtures.

**4. The Seam C guard's structural half — has holes.** See the inline finding on `rung-state-seam.test.ts:73`. Five natural spellings of a copy-driven branch pass both halves of the guard.

**5. C1 held — holds.** `git diff 19af544..HEAD -- apps/web/src/projection/contract.ts packages/` is empty. `grep -rn '\.label' apps/web/src` (non-test) returns zero reads of a rung's wire `label` on the render path: only `push/fixture-synthesis.ts:452` and `push/dca-block.ts:271` pass it through, plus two unrelated `SectionTable` fields and the two chart *mark* labels.

**6. The `filled` predicate — one authoring site on the render surface, holds; two negated re-spellings remain elsewhere.** See the inline nit on `started-ladder.fixtures.ts:174`. The undefined-`venueAxis` arm is `false`, and that is now pinned by name in `fill-path-view.test.ts`.

**7. The fixtures are hand-authored — holds.** Prices on a round 4,000 grid from 50,000 to 22,000, a hand-drawn 200→1,100 ramp, every id `fixture:`/`fixture-`/`facade00-…`, and `qty()`/`DECLARED_TOTAL_USD` are arithmetic over those authored numbers. Nothing has the fingerprint of real output — not scaled, not perturbed, not reshaped. Cardinality (eight rungs) is a count, and the module says so.

### Findings

Three, all low. Ranked most-severe first; each one inline at its line.

1. **Low** — the chart's legend prints a `Waiting` swatch on a ladder with nothing waiting (`PriceDropPathChart.tsx:495`). Verified against the rendered `overfilled` fixture.
2. **Low** — the Seam C structural scan misses five natural spellings of a copy-driven branch (`rung-state-seam.test.ts:73`). Verified by running the two regexes against each spelling.
3. **Nit** — the filled predicate is re-spelled inline at two sites now that `venueFilled` is exported (`started-ladder.fixtures.ts:174`, and `push/fixture-synthesis.ts:500`, which is outside this diff so it cannot be anchored).

### Clean, briefly

I went looking in these places and found nothing to report: the arithmetic in `price-drop-path.ts` (the √-on-area normalization, the `R_MIN` boundary at 12.96×, the `largest > 0` NaN guard, the array-order running sum, the `splitAt` last-index rule); the `deployedMarkFor`/`spotMarkFor` pairing of clamp-and-glyph; the `stateCopy` swap in `Pills`/`RowState`, which is behaviour-preserving because the engine emits `"waiting"` for `axis === "resting"` and for nothing else, and a wire rung cannot carry `label` without `venueAxis` (`push/dca-block.ts:258-287` sets them in one object or neither); the four fixtures' claims, each of which does exercise its named branch; and the docs — `docs/ladder-fill-path.md`'s four structural guarantees each check out against `fill-path-view.ts` (`ExpectedFigures` is a distinct type, `Expectation` owns the `~`, the average is `totalUsd / units` at line 453, and `expected` is gated on `hasNotStarted` at line 577).

The deferred items (T7/D1, the y-pad arithmetic staying in the `.tsx`, `fixture-synthesis.ts:440`, O2/O3, the duplicated `partly filled · 40%`, #285 §6.3b, D2's unguarded `color-mix()`) were read and are not reported: each is recorded, and none is a defect *now* rather than a deferral.

PR #310, opened 2026-08-13
Fix phone login, and stop a hung price-feed run from eating the nightly window

AmetAlvirde:
Two independent pieces of work that happened to be in flight together: the phone-login fixes, and the price-feed hang found while investigating a stale dashboard. They share no code — separate commits, reviewable separately.

Closes #309

## 1. Login from a phone — `feat(auth)`

Two distinct reasons sign-in failed on mobile while working on the desktop:

- **Origin.** Better Auth CSRF-checks every `/api/auth/*` request against `baseURL`, a single fixed origin — so `vite dev --host` reached at the machine's LAN IP is rejected as "Invalid origin" even though it is the same server. Adds an opt-in `DEV_TRUSTED_ORIGINS` list, dropped entirely when `NODE_ENV=production` so a LAN IP can never widen the CSRF surface on a hosted deploy.
- **Credentials.** Past that, the form still failed: phone keyboards and autofill append a trailing space and can capitalize the first letter even in a `type="email"` field, and the stored account email is lowercase. The email is now trimmed and lowercased — **the password deliberately is not**, since whitespace and case are significant there and silently trimming one would reject a legitimate credential.

## 2. The nightly price feed could hang forever — `fix(price-feed)`

Full incident write-up in #309. In short: on 2026-08-11 the 22:01 run upserted 11 of 43 anchors and the laptop slept. The socket died silently, `node-postgres` applies no client-side query deadline, and the run was still alive **20 hours later**.

**It hung rather than failed, and that distinction is the entire bug.** A crash is harmless here — red exit, heartbeat, next hourly fire heals it, `backfill` is idempotent precisely so that works. A hang defeats all three mechanisms at once: the EXIT trap never runs (so the heartbeat still describes the *previous* run and reads healthy), and launchd's per-label singleton — which this wrapper's header relies on to justify having no lock — keeps the job slot held. Since `isFreshBar` refuses yesterday's bar once the CDMX date rolls, every skipped evening is unrecoverable. One dead socket costs an unbounded run of days, with no signal.

### What changed

- **`apps/web/src/push/projection-pool.ts` (new).** Shared deadlines for both unattended writers. `query_timeout` is the hang-breaker and the only setting that covers this case — `connectionTimeoutMillis` caps a connect phase that had already succeeded, `statement_timeout` is the server's verdict on a socket it can no longer reach, and `keepAlive` depends on an OS tunable we do not set. All four are kept, each shortening a different failure; the server's cap sits below the client's so a genuinely slow query returns a diagnosable `query_canceled`.
- **Wrapper watchdog.** Bounds the whole run at 2700s → TERM the process group → SIGKILL after grace, converting the signal into `exit 124`. The ceiling beats the **next fire**, not the end of the window — the slot must be free at the top of the hour for the schedule's own retry to work — and is pinned against the plist's real interval by a test, since the two files never reference each other.
- **Two SIGTERM senders, told apart.** The watchdog is not the only thing that sends this process a bare TERM: `launchctl stop`, `launchctl unload` (the documented reinstall step), a logout and a shutdown all land in the same handler. The watchdog leaves a per-run calling card immediately before it signals, so a timeout records `exit 124` with a `"lastStep": "timeout:<step>"` label, while an **external** stop records `exit 143` (128+SIGTERM) with a plain step name — "someone stopped this" rather than "this hung". The heartbeat is written on both paths, which is new: before the watchdog existed an external TERM killed the shell by signal and the EXIT trap never ran. Deliberate, because a run stopped *after* `spine` really did land the day's marks and only a heartbeat can say so.

### Two defects found by testing the watchdog, not by reading it

Both would have defeated it in production, and both are the reason this diff is larger than "add a timeout":

- **Teardown lost the heartbeat on 3 of 5 runs.** The group-TERM also killed the `tee` behind the script's stdout, so the shell died of SIGPIPE partway through `exit`, before the EXIT trap body ran — the watchdog would have killed the run and left no breadcrumb, which is exactly the failure it exists to prevent. The logging `tee` is now deaf to TERM so the pipe outlives teardown.
- **The watchdog leaked and re-created the original bug.** Killing the subshell does not kill its `sleep` child; the orphan stayed in the process group *and* held the inherited write end of the log pipe, so `tee` never saw EOF either. Observed on a **healthy** run: it reported "complete", then sat there with a stray `sleep`, a live `tee`, and a held job slot. Children are now enumerated before the parent dies and reaped with it.

### Plus a latent heartbeat bug this makes reachable

The `schemaVersion`-1 fallback fired for any file not saying `"markWindow": false` — also true of an ordinary v2 file from an in-window run that died before `spine`. It promoted that failure's own `finishedAt` into a marked-day stamp, **manufacturing evidence that the day was covered**, silencing the staleness warning on exactly the morning it was needed. Latent before; a timed-out run is precisely that shape, so the watchdog would have started tripping it. Now gated positively on v1, so an unrecognised future version fails closed.

## Review round — two more defects, again found by running it

Reviewed in a fresh context so the session that authored the change did not review its own work. Six findings; the two that mattered were both in the watchdog, and both were reproduced before being fixed. Commits `69b909d`, `e05baa0`, `c3a45a9`.

- **The timeout label broke the "did this run land the marks?" predicate.** The TERM handler rewrote `LAST_STEP` to `timeout:<step>` in place, and `MARKS_LANDED_STEPS` is an **exact-string** match — so a run the watchdog killed at `backfill`, the literal 08-11 shape, had already appended, committed and verified the day's marks yet recorded itself as not having landed them. `lastMarkWindowFinishedAt` fell back to the previous evening and the next morning read *"nothing recorded for today"* beside a true exit-124 line: the cry-wolf that the comment above `MARKS_LANDED_STEPS` exists to prevent. Fixed by carrying the timeout as a separate `TIMED_OUT` flag and building the `timeout:` label once, at print time, **after** the predicate has run — so the predicate sees a bare step name and the reader still gets the diagnosis in one field.
- **Cancelling the watchdog made its own SIGKILL escalation dead code.** The reap in `write_heartbeat` ran within milliseconds of the group TERM, killing the watchdog long before its grace expired — so the escalation could never fire in the one case it exists for, and any group member that *ignored* the TERM survived. Reproduced exactly that way: a TERM-deaf grandchild lived on in the run's pgid and, holding the inherited write end of the log pipe, kept `tee` alive with it — both still holding launchd's slot. The reap is now skipped on the timeout path so the watchdog finishes its own grace and group-SIGKILL. Gated on `TIMED_OUT` rather than on `exitCode == 124`, since a step is free to exit 124 on its own terms and treating that as a timeout would leave a live watchdog sleeping out the full ceiling over an already-dead run.
- **The watchdog now waits in 5s hops instead of one `sleep 2700`.** The `pgrep -P` / `kill` cancel is inherently racy and the arming window — before the first timer is forked — is not covered by the original "it is always asleep for the full ceiling" argument. Short hops do not close the race, they make losing it cheap: a lost child orphans a 5-second timer rather than a 45-minute held job slot. `SECONDS` is wall-clock, so the ceiling stays honest however the hops round.
- **Two test-only guards.** The exact-string predicate is now pinned against the `timeout:` decoration, and the pool test's "gives up well inside the one-hour gap" claim — which sums one connect + one query against a backfill that issues 43-44 sequential upserts — now rests on a stated, pinned invariant (`runBackfill` aborts on first error) rather than an unstated one. The watchdog-ceiling guard also gained a floor: `MAX_RUN_SECONDS=600` used to pass green while TERMing every healthy ~15-minute run.

## Verification

- **Timeout path:** 6/6 consecutive runs with a reactive child and 6/6 with a `trap "" TERM` grandchild — exit 124, `lastStep=timeout:<step>`, the day stamped to *this* run's `finishedAt`, **zero leaked processes** in the pgid.
- **External stop:** 6/6 `killpg` mid-`prices-fetch` — exit 143, undecorated step name, zero leaked.
- **Early failure:** 12/12 `exit 127` under a 300s ceiling — zero leaked.
- **Healthy path:** exit 0, heartbeat `"complete"`, zero leaked.
- **Negative control against the pre-review commit:** 3/3 timeouts carried yesterday's stamp forward, left **4 processes** in the pgid (deaf grandchild, its `sleep`, the `tee` subshell, `tee`), and 3/3 external TERMs mislabelled themselves `124 / timeout:prices-fetch`. Both new vitest guards confirmed red against their regressions before being restored.
- **Heartbeat carry:** v2-unmarked no longer invents a stamp; genuine v1 still carries its `finishedAt`.
- **Real run via `launchctl start`:** watchdog armed, completed in 76s, exit 0, 44 anchors through `2026-08-12` — the day the wedged process had already eaten was recovered before the date rolled.
- `/bin/bash -n` clean under 3.2.57 (what launchd runs). `shellcheck` unavailable on this machine.
- `pnpm typecheck` clean; `pnpm test` **1925 passed / 19 skipped**.

## Reviewer notes / known gaps

Three residuals are filed as follow-ups rather than fixed here — none of them blocks this merge, and all three are tracked:

- **No committed test harness for the wrapper** — issue #313. Two rounds, four signal-handling defects, every one of them found by *running* the script and none by reading it; both harnesses were thrown away. This remains the weakest part of the change and the most likely place for a regression to hide. It is a real increment, not a tidy-up.
- **The `pgrep` cancel race is bounded, not closed** — issue #311. Cost reduced from a 45-minute held job slot to a 5-second stray timer. Keeping the mitigation and closing that issue is a legitimate outcome.
- **The external-stop path has only been exercised synthetically** — issue #312. `killpg` is not `launchctl unload`, a logout, or a shutdown, and the exit-143 path writes a heartbeat where the pre-watchdog code wrote nothing.

Two more, unchanged from the original write-up:

- `apps/web/src/projection/snapshot-reader.ts` builds its own pool without these deadlines. Different lifecycle (request-scoped, behind a platform function timeout) so it is out of scope here, but it is the same hazard class.
- The local dev projection DB is only ever written by hand — the nightly job writes to the hosted DB — so a local dashboard drifts stale by design. Not addressed here; worth deciding separately.

### DIFF
--- a/apps/price-feed/src/schedule-window.test.ts
+++ b/apps/price-feed/src/schedule-window.test.ts
@@ -209,6 +209,38 @@ describe("the launchd fetch window", () => {
     expect(declared).toEqual(afterSpine);
   });
 
+  it("never rewrites LAST_STEP into a decorated value, which empties the landed-steps match", () => {
+    // The sibling test above pins WHICH names count as "the marks landed". This one
+    // pins that the variable those names are compared against still holds one of
+    // them. The comparison is an exact string match, so any assignment that turns
+    // `LAST_STEP` into a LABEL rather than a step name silently empties the whole
+    // predicate — every branch of the loop falls through and the run records itself
+    // as not having marked the day.
+    //
+    // That regression shipped once. The watchdog's TERM handler set
+    // `LAST_STEP="timeout:$LAST_STEP"`, so a run killed at `backfill` — after it had
+    // already appended, committed and VERIFIED the day's marks — stamped nothing,
+    // and the next morning claimed "nothing recorded for today" standing beside a
+    // true exit-124 line and a gap report that said the opposite. The `timeout:`
+    // label is still emitted; it is built at print time from a separate flag, which
+    // is what lets both facts be true at once.
+    //
+    // THE ANCHORED SCAN THE SIBLING USES CANNOT SEE THIS. That rewrite lived inside a
+    // single-quoted `trap` body on a `trap` line, so the `/^LAST_STEP="…"$/gm` list
+    // was still exactly the post-`spine` steps and the guard passed. This scan finds
+    // every assignment wherever it sits, and requires the two to agree — an
+    // assignment the anchored scan misses is itself the bug, because that scan is the
+    // oracle for `MARKS_LANDED_STEPS`.
+    const wrapper = readFileSync(WRAPPER_PATH, "utf8");
+    const everywhere = [...wrapper.matchAll(/\bLAST_STEP="([^"]*)"/g)].map((m) => m[1]!);
+    const anchored = [...wrapper.matchAll(/^LAST_STEP="([^"]+)"$/gm)].map((m) => m[1]!);
+    expect(everywhere).toEqual(anchored);
+    // A bare step name: no `$` (an expansion of itself), no `:` (a label prefix).
+    for (const value of everywhere) {
+      expect(value).toMatch(/^[a-z][a-z-]*$/);
+    }
+  });
+
   it("bounds a wedged run inside the gap between fires, so the next one is not skipped", () => {
     // THE ORACLE IS THE PLIST, WHICH IS WHY THIS LIVES HERE. The wrapper's watchdog
     // ceiling is only meaningful relative to how often launchd actually fires: the
@@ -244,5 +276,20 @@ describe("the launchd fetch window", () => {
       Math.min(...minutesOfDay.slice(1).map((m, i) => m - minutesOfDay[i]!)) * 60;
 
     expect(ceiling + grace).toBeLessThan(smallestGapSeconds);
+
+    // AND THE FLOOR, because the bound is two-sided and only the ceiling had a test.
+    // "Tighten it to be safe" is the single most plausible edit anyone will make to
+    // this number, and it is the one edit with the worst consequence: a ceiling at or
+    // below a healthy run's duration TERMs EVERY fire, not only the wedged ones. All
+    // six fires of an evening then die identically, the CDMX date rolls, and the day
+    // is unrecoverable — total loss, from an edit that looks like prudence and that
+    // the assertion above waves through (600 passes it cleanly).
+    //
+    // THE ORACLE IS THE WRAPPER'S OWN MEASURED RUN LENGTH, stated at the watchdog:
+    // "a healthy run takes ~15 minutes". Weaker than DEFAULT_CONFIG or the plist, and
+    // named as such — but it is a real observation of the thing being bounded, and
+    // 20 minutes is that figure with a third of itself as margin. This asserts the
+    // ceiling leaves room for a healthy run; it does not pin 2700.
+    expect(ceiling).toBeGreaterThan(20 * 60);
   });
 });
--- a/apps/price-feed/src/schedule-window.test.ts
+++ b/apps/price-feed/src/schedule-window.test.ts
@@ -208,4 +208,41 @@ describe("the launchd fetch window", () => {
     const declared = /^MARKS_LANDED_STEPS="([^"]+)"$/m.exec(wrapper)?.[1]?.split(" ");
     expect(declared).toEqual(afterSpine);
   });
+
+  it("bounds a wedged run inside the gap between fires, so the next one is not skipped", () => {
+    // THE ORACLE IS THE PLIST, WHICH IS WHY THIS LIVES HERE. The wrapper's watchdog
+    // ceiling is only meaningful relative to how often launchd actually fires: the
+    // whole point is that a hung run is dead and the job slot released BEFORE the
+    // next fire, so the hourly schedule's own retry still works.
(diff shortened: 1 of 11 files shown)


AmetAlvirde (commented):
Review of the full diff (10 files, 439/-7): the wrapper's signal machinery, `projection-pool.ts` + its two call sites + its test, the `schedule-window` structural guard, and the auth/login pair. Read against `packages/event-store/src/heartbeat.ts` (the consumer), the plist, `pg@8.22.0`/`pg-pool@3.14.0` sources, and `better-auth@1.6.23`'s origin check.

Method note: the wrapper's teardown was checked by running a faithful reproduction under its own session (`setsid`-equivalent) with a 3s ceiling, not only by reading it.

6 findings: 1 medium-high, 1 medium, 4 low. Most of what the handoff flagged checked out — the `tee` TERM-deafness, the TERM->`exit 124` heartbeat, the steady-state `pgrep -P` reap, the v1 gate, the pool's four settings and their ordering, the `trustedOrigins` production drop, and the email normalization were all verified sound and carry no comment here.

PR #322, opened 2026-08-13
feat(observability): a venue that goes dark silently stops reading as clean

AmetAlvirde:
A venue can go dark **silently** — serving stale or empty bars rather than erroring — and every instrument self-skips as INFO, the run exits 0, the heartbeat reads healthy, and the gap report reads clean because the crypto marks landed. Three health surfaces all report fine while a whole venue produced nothing. This is what actually closes #266.

It ships as one vertical slice, in two commits, because the pieces are not independently shippable: the constant move is a behaviour-preserving precondition for the derivation, and without the delivery change the verdict would be relocated rather than delivered.

## The derivation (D1–D5, D7–D8)

`computeGapReport` gains a **second question**, and its header now states both. `GapReport` gains `venueDark: VenueDarkDay[]` (`{ date, source, expected }`) and an inert `unattributedMarks` bucket. `LostDayReason` stays two members on purpose: three surfaces render `lost.length` as "N lost day(s)", and a venue-dark day is **not** a lost day — the feed ran, the day is anchored, NAV computes.

It is strictly additive: it can only fire on a day that already has an anchor **and** at least one mark, so `no-anchor` and `no-marks` keep their exact current meanings. Attribution is built from `instrumentsForSource`, never `resolveInstrument` — that one throws loud on an unknown id by design, which is right for the fetch orchestrator and catastrophic in a pure retrospective derivation where a since-retired instrument would crash every historical report. Unattributable marks are bucketed separately and can never suppress a verdict.

Warn only. No suppression, no exit-code change, `SUPPRESSION_KEYS` untouched, `exitCode` still the literal type `0`. With no holiday calendar (D7, deliberately deferred), roughly 9–10 weekday holidays a year fire — and a non-zero exit would therefore fail the daily automation ten times a year. The failure direction is the safe one: a false *no* is the one failure a triage surface cannot have; a holiday false positive is a false *yes*, and the message names venue and weekday so a holiday reads as one.

## The move (D4)

`VENUE_CADENCE`, `isWeekend` and `lastExpectedMarkDate` move down to the dependency-free `@numisma/engine` — `isWeekend` and a new `weekdayName` into `calendar.ts` (still zero imports), the venue-keyed half into a new `price-feed/venue-calendar.ts` beside the registry that defines `PriceSource`. `glance.ts` and `gap-report.ts` now import the one definition. The `satisfies Record<PriceSource, VenueCadence>` latch only ever guaranteed exhaustiveness, not agreement: two copies compile happily while disagreeing about whether a venue marks on weekends, and the visible result is two surfaces contradicting each other about the same day. ADR-001 is not in the way — it keeps *file IO* out of engine, and this is pure data and pure date math. Engine still has zero workspace dependencies.

The hand-copied `arrivedUnderOldRule` pair in `gap-report.test.ts` is deliberately **left** hand-copied, with a DO-NOT-COLLAPSE note: importing the shared definition would make both sides of the disagreement the same code.

## The delivery (D6)

The gap report's only channels were TUI startup — which speaks when the operator opens it — and an on-demand command whose output lands in a log nobody reads. The formatter change gives both the new lines for free; the payload is what reaches the phone unbidden.

`GlanceBlock` gains an optional `venueDark?: GlanceVenueDarkDay[]`, most-recent-dark-day-per-venue over a 7-day window, and it renders as a fifth verdict trigger placed after `feedGap` and before `reserveFloor` — the first three triggers answer *is the data trustworthy*, the last two *what is the fund doing*. Three states, all load-bearing: **absent** means this build could not answer (pre-v6 row, or the derivation degraded) and renders nothing — never an all-clear; `[]` means checked and clean; non-empty names the venues.

**It can never withhold the push.** Both the log read and the derivation sit inside one `try` that degrades the field and still writes the anchor, with a pure never-throws wrapper tested directly. An observability finding must not be able to cost the phone its NAV.

**Backfill resolves per-anchor, not stamped.** `summarizeVenueDark(events, asOf)` is a pure function of `asOf` and never of the wall clock, reproducing D2's yesterday-pinned ceiling for each historical anchor the way the plans sidecar already resolves as-of each anchor's own date.

### ADR-007 — verified, not assumed

D6 attached this as a condition and worded it as *"a verification, not an assumption."* It was run before any source was touched.

- **No ISO-date-shaped value reaches the wire.** The verdict names **venue + weekday**, which is D7's ruled message shape anyway. The two date-scan tests (`projection-payload.test.ts`'s D14 block and `anchor-fixture.test.ts`'s fixture scan) pass **unmodified** — that they still pass, unweakened, *is* the verification.
- **It lands inside the existing `glance` branch, not as a fifth top-level branch.** `contract.ts` rules a new top-level branch outside the C5 carve-outs by construction; both prior top-level branches (v3 `glance`, v4 `dca`) shipped an ADR-007 amendment in the same PR. The precedent followed here is instead **v5 (`b32a9ad`, the fill path)**: optional fields inside an existing branch, version bump, both allow-lists grown, ADR-007 untouched. **No amendment is owed and none is made.**
- Substantively it discloses strictly **less** than `glance.feedGap.missing[]` already ships — that carries the row id and label of every instrument whose mark did not arrive; this carries a venue name and a weekday name. Zero new classes of data.
- Both allow-lists grew, enumerated one path per field with no wildcard: `ProjectionKeyAllowList` gains `venueDark` on the `glance` exact-keys assert plus a `glanceVenueDark` row assert; the runtime key-path list gains three paths.

`COMPOSITION_SNAPSHOT_SCHEMA_VERSION` bumps 5 → 6 for the **writer's**-contract reason v5 records, not reader safety: the version is what tells a reader whether an absent `venueDark` means "this build could not have written one" or "no venue was dark". `MIN_SUPPORTED_SNAPSHOT_SCHEMA_VERSION` stays **4**, so v4 and v5 rows stay renderable and this bump costs no cutover window.

## That the delivery is real

Replayed over the regenerated 44-anchor fixture: **five days the old surface said nothing while a whole venue was dark** (`2026-07-06`…`07-10`) — precisely the case `feedGap` cannot catch, since the equity feed returned partway through and `feedGap` went quiet that morning while the earlier silence stayed invisible. Two days (`07-04`, `07-05`) show same-day trigger collisions, where `feedGap` correctly takes the line; `verdict-replay.test.ts` now names every day's full firing set rather than asserting no day fires twice, which is strictly stronger.

## Review round

Four findings were filed by an independent reviewer and all four are fixed in this PR.

The one that mattered: `formatGapReport` concatenated `[...lost, ...venueDark]` and the TUI's 7-line bound kept the tail, so permanent lost days were the **first** lines withheld — and since the TUI's window floor is fixed while D7 accepts ~9–10 holiday false positives a year, `venueDark` alone would cross 7 within about a year and the only daily-read channel could then never name a lost day again. Not a severity trade-off; starvation. The bound is now **reserved per-kind capacity** (`RESERVED_LOST_LINES = 4` against `MAX_GAP_LINES = 7`), a floor rather than a quota — a quiet venue-dark side still yields all seven lines to lost days — with each kind keeping its own most-recent tail and the withheld-count line naming what was withheld from each. `formatGapReport`'s own output is byte-identical, so the CLI and `gap-report.json` consumers are untouched.

Also fixed: the verdict precedence docstring, which still listed four triggers and still claimed no same-day collision had ever been observed while the same commit measured some; and the window floor in `summarizeVenueDark`, which was not clamped to `LAUNCHD_ERA_START`, so backfilled anchors derived verdicts from the pre-scheduler hand-run era. That clamp is what moved the numbers above — `06-28` returns to a *no* day and two of the four apparent collisions turn out to have been artifacts of the unclamped window.

## Gates

`pnpm typecheck` exit 0 across all six projects. `pnpm test` exit 0 — 1954 passed, 19 skipped, 0 failed.

Every fixture and test value here is authored or run through the sanctioned synthesis path; no test is seeded from real output.


AmetAlvirde (commented):
Review of #322 (fresh non-authoring context; skill `pr-review` → `review`).

Gates re-run locally and they hold: `pnpm typecheck` exit 0, `pnpm test` exit 0 — 1951 passed, 19 skipped, 0 failed.

**The claims I was asked to check independently, and what I found**

- **ADR-007 date invariant — holds.** Both date scans are genuinely unweakened: `projection-payload.test.ts`'s D14 block and `anchor-fixture.test.ts`'s fixture scan are byte-identical to `main` (the only edit to those files is the `5`→`6` version pin). They are also *exercised* rather than merely still-passing — `TEST_GLANCE` now carries a non-empty `venueDark`, so `scanDateShapedValues` walks into `$.glance.venueDark[].source` and `[].weekday`. Both values are closed by construction (`PRICE_SOURCES`; the hardcoded `WEEKDAY_NAMES` table, deliberately not `Intl`), so no ISO-shaped value can reach the wire through the new field. Both allow-lists grew by enumeration with no wildcard, and the fixture's 44 anchors carry only weekday names.
- **Push can never be withheld — holds.** `loadVenueDarkAsOf` wraps the path resolve, the log read *and* the derivation; `venueDarkOrOmit` wraps the pure derivation again. No throw escapes to `buildGlanceForAnchor`. The docstring's premise that `loadFoldedReview` has already run `assertLogFullyLoaded` checks out (`event-store.ts:196`).
- **v5→v6 with the floor at 4 — holds.** `isSupportedSchemaVersion` is a range test; the only reader of `glance.venueDark` is `computeVerdict`, which does `glance.venueDark ?? []` and requires `length >= 1`, so absent renders *nothing*. There is no all-clear string anywhere on this path.
- **`summarizeVenueDark` is a pure function of `asOf` — holds.** `now` is synthesized as `${asOf}T18:00:00Z`; CDMX is UTC-6 year-round, so `dueThrough` lands on `asOf - 1` and matches the explicit `until`. No wall clock is read.
- **D4 move is behaviour-preserving — holds.** `VENUE_CADENCE`, `lastExpectedMarkDate` and `allRegisteredInstruments`' key list are identical; `calendar.ts` still imports nothing; `packages/engine/package.json` still declares no dependencies. The hand-copied `arrivedUnderOldRule` pair is correctly left alone (not flagged).
- **Test authorship — holds.** Registry instrument ids are code-owned reference data, and every price/amount in the new tests is invented. The fixture goes through `fixture-synthesis.ts` (`glance` copied verbatim, which is what `glance.test.ts`'s live re-derivation then pins), and `expect(recorded.glance.venueDark).toBeDefined()` correctly closes the hole where a degraded regeneration could pass as agreement.
- **No ADR-007 amendment owed — the precedent claim checks out.** `b32a9ad` did not touch ADR-007 and it added a fourth root field (`tornActs`) to `DcaBlock` inside an existing branch, exactly as this adds a fourth to `GlanceBlock`. Worth noting only that both amendments now under-describe their branch ("carries three things"); that is a pre-existing pattern, not something this PR introduces, so I am not filing it.

**Four findings below**, one of which I'd want resolved before merge.

PR #339, opened 2026-08-13
fix(tui): a unique temp path per write, so concurrent log writers stop silently losing whole images

AmetAlvirde:
`writeLogImage` built its temp path by appending a fixed suffix to the log path, so every writer against a given log shared `${logPath}.tmp`. Two concurrent writers interleaved on that one file, and because each builds a **complete next image** and renames it over the log, the loser's entire image vanished silently — no torn file to notice afterwards, just a staged image that never landed.

Closes #296.

## What changed

`apps/tui/src/event-store.ts` — the temp name is now `${logPath}.${process.pid}.${tick}.tmp`, where `tick` is a module-level counter incremented per call.

Pid alone is not sufficient. The observed corruption is two in-flight writes from **one** process (`Promise.all` over callers), so the counter is the part that actually separates them; the pid covers the cross-process case. The temp file stays a **sibling of the log**, so the rename remains same-filesystem and therefore still atomic — that directory constraint is the whole reason this is not simply "stage into `os.tmpdir()`".

Cleanup is a `try`/`catch` around write+rename that `rm`s its own temp and rethrows. No shared state, no directory scanning, nothing to reconcile.

The three real callers — `appendEvents`, `migrateLegacyLog`, `restoreLogImage` — plus the `record-fill` injection point are untouched. Note that `restoreLogImage`, the rollback half of the two-file fill act, routes through the same function, so the fix covers rollback as well as append.

## The test, and why it is large on purpose

Extended the house file `apps/tui/src/event-store.test.ts` with `writeLogImage — concurrent writers cannot corrupt each other`. Four writers, each a distinct ~4 MB image built from an authored marker line repeated 60,000 times. It asserts that every write resolves, that the surviving image contains exactly one distinct line, its full length, that it is byte-identical to one of the four inputs, and that no `.tmp` files are left in the directory.

**Verified red before the fix.** Against the pre-fix code the concurrency assertion fails with three `ENOENT … rename '…/events.jsonl.tmp' -> '…/events.jsonl'` errors: three of four writers lost their staged image entirely because a sibling renamed the shared temp out from under them. That is the defect itself, not an assertion on an implementation detail.

**The payload size is load-bearing.** With a small payload the whole write completes inside one turn, the writers never overlap, and the test passes against broken code. The multi-megabyte payload is what makes the race deterministic. This is stated in a comment in the test so nobody "optimizes" it back into green-on-broken.

Every byte of test input is authored — generated filler, nothing read from any log or data directory.

## Note for a future increment: should concurrent writers be *prevented*?

Issue #296 asks whether concurrent writers to one log should be prevented outright rather than merely made safe. **Deliberately not answered here, and deliberately not built.** The TUI is single-operator, so today the collision needs two processes; a lock would say that explicitly instead of leaving it to the temp-name scheme. This PR makes the existing concurrency safe and records the question rather than growing into a locking increment.

`event-store.test.ts:551` ("migrateLegacyLog writes through writeLogImage") already anticipated this fix — its comment names "a distinct temp name" as hardening that must land in `writeLogImage` because it is the one writer of the whole log. That structural assertion still holds.

## Verification

- `pnpm typecheck` — clean across all six packages.
- Test suite scoped to this tree — 137 files, 1953 passed, 21 skipped, 0 failed.

One caveat for reviewers: a bare `pnpm test` from a worktree root is contaminated while sibling worktrees exist, because `.claude/` is a symlink into the main checkout's worktree directory and vitest follows it (822 files collected, failures belonging to other lanes). This is an artifact of concurrent worktrees, not of this change, and the vitest config was **not** modified to paper over it. CI is unaffected — `.claude/` is gitignored.

## Merge order

**Base branch: `main`.** This is the first PR of the lane C stack:

1. **this PR** — `feature/unique-log-temp-path` → `main`
2. #294 — `feature/cli-shell-tests` → `feature/unique-log-temp-path`
3. #298 — `feature/r6-r7-single-owner` → `feature/cli-shell-tests`

Merge in that order. GitHub retargets each stacked PR to `main` automatically when its base is deleted on merge. Nothing on this board merges before the end-of-workflow batch-merge session.

### DIFF
--- a/apps/tui/src/event-store-write-failure.test.ts
+++ b/apps/tui/src/event-store-write-failure.test.ts
@@ -1,0 +1,130 @@
+/**
+ * THE FAILING LOG-IMAGE WRITE — the one path on which `writeLogImage` can leave litter
+ * behind, and the reason it needs a test file of its own.
+ *
+ * `writeLogImage` writes a unique temp sibling (`<log>.<pid>.<tick>.tmp`) and `rename`s
+ * it over the log. On the happy path the rename CONSUMES the temp, so every passing
+ * success case — including the four-writer concurrency case in `event-store.test.ts`,
+ * where every writer succeeds — proves exactly nothing about the `catch { rm }`. Delete
+ * the whole try/catch and those tests stay green.
+ *
+ * The window is between `writeFile` and `rename`, and no arrangement of real files opens
+ * it: every candidate (a directory at the target, a read-only parent) fails BEFORE a temp
+ * file exists, so the test would pass vacuously. So `rename` is mocked, and only
+ * `rename`: everything else is the real filesystem in a real temp directory. This file is
+ * separate because that mock is module-wide and has no business near the ingest tests.
+ *
+ * What it locks: a failed image write leaves NOTHING beside the log, and the caller sees
+ * the TRUE cause. Without the first, `events.jsonl.<pid>.<n>.tmp` accumulates one
+ * multi-megabyte orphan per failed attempt, forever and invisibly — accumulus ignores
+ * `/data/*.tmp`, so nothing on the durability path would ever surface it. Without the
+ * second, a cleanup that throws in turn (the dir went read-only, which is often WHY the
+ * write failed) replaces "the log image failed to publish" with an `unlink` error naming
+ * a temp file the operator has never heard of.
+ *
+ * Every fixture here is authored filler — no real event, price or balance.
+ */
+import { mkdtemp, readdir, readFile, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+/** Flipped by a test to make the next `rename` fail the way a full volume would. */
+let renameFails = false;
+/**
+ * Flipped to make the CLEANUP fail too. Real-file arrangements cannot stage this
+ * reliably — making the dir read-only mid-flight races the `writeFile` — and the
+ * masking bug it targets is exactly a two-failure sequence, so both failures are
+ * injected rather than provoked.
+ */
+let cleanupFails = false;
+
+vi.mock("node:fs/promises", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:fs/promises")>();
+  return {
+    ...actual,
+    rename: async (from: string, to: string) => {
+      if (renameFails) {
+        throw Object.assign(new Error("ENOSPC: no space left on device, rename"), {
+          code: "ENOSPC",
+        });
+      }
+      return actual.rename(from, to);
+    },
+    rm: async (path: string, options?: Parameters<typeof actual.rm>[1]) => {
+      if (cleanupFails) {
+        throw Object.assign(new Error(`EACCES: permission denied, unlink '${path}'`), {
+          code: "EACCES",
+        });
+      }
+      return actual.rm(path, options);
+    },
+  };
+});
+
+const { writeLogImage } = await import("./event-store.js");
+
+const PRIOR_IMAGE = '{"kind":"synthetic","n":1}\n';
+const NEXT_IMAGE = '{"kind":"synthetic","n":1}\n{"kind":"synthetic","n":2}\n';
+
+const createdDirs: string[] = [];
+
+afterEach(async () => {
+  renameFails = false;
+  cleanupFails = false;
+  for (const dir of createdDirs) {
+    await rm(dir, { recursive: true, force: true });
+  }
+  createdDirs.length = 0;
+});
+
+async function tempDir(): Promise<string> {
+  const dir = await mkdtemp(resolve(tmpdir(), "numisma-log-image-"));
+  createdDirs.push(dir);
+  return dir;
+}
+
+describe("writeLogImage cleans up after ITSELF when the write cannot land", () => {
+  it("leaves no temp sibling behind, and does not create the log", async () => {
+    const dir = await tempDir();
+    const logPath = join(dir, "events.jsonl");
+    renameFails = true;
+
+    await expect(writeLogImage(logPath, NEXT_IMAGE)).rejects.toThrow(/ENOSPC/);
+
+    // The whole directory, not a guess at the temp's name: anything at all left here is
+    // litter, including a name a future revision of the scheme might choose instead.
+    expect(await readdir(dir)).toEqual([]);
+  });
+
+  it("leaves the PRIOR image untouched — a failed write loses no committed line", async () => {
+    const dir = await tempDir();
+    const logPath = join(dir, "events.jsonl");
+    await writeLogImage(logPath, PRIOR_IMAGE);
+
+    renameFails = true;
+    await expect(writeLogImage(logPath, NEXT_IMAGE)).rejects.toThrow(/ENOSPC/);
+
+    // Crash-atomicity is the entire warrant for writing through a temp, so the cleanup
+    // must not have cost it: the log still holds exactly what it held before.
+    expect(await readdir(dir)).toEqual(["events.jsonl"]);
+    expect(await readFile(logPath, "utf8")).toBe(PRIOR_IMAGE);
+  });
+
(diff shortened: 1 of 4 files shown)


AmetAlvirde (commented):
## Review — fresh context, no assumption that the author's reasoning holds

**Verdict: sound to merge.** The `writeLogImage` change itself is correct — I tried to construct an interleaving that still collides and could not. One finding I'd want addressed (test-only, a few lines), three smaller ones.

---

### 1. This PR silently empties four pre-existing litter guards, and nothing it adds covers the cleanup it introduces — MEDIUM

Four assertions across two files assert the absence of the *old* fixed temp name:

- `apps/tui/src/event-store.test.ts:374` — after `ingestInbox` (`appendEvents` path)
- `apps/tui/src/event-store.test.ts:599` — after `migrateLegacyLog`
- `apps/tui/src/record-fill-reliable.test.ts:223` — `${eventsPath}.tmp` after a clean fill
- `apps/tui/src/record-fill-reliable.test.ts:252` — `${eventsPath}.tmp` after a **rollback** via `restoreLogImage`

All four are `expect(await exists(`${…}.tmp`)).toBe(false)`. The new scheme never produces that name, so as of this commit **none of them can fail**. They are green for the wrong reason.

The new concurrency test does not replace them. Every writer in it succeeds, so every temp is consumed by its own `rename` and the `catch { rm }` never executes. I verified this rather than assuming it: I ran a faithful copy of the new implementation with the entire `try/catch` deleted, 10 runs — `leftover: 0` every time. **The cleanup half this PR introduces has zero coverage**, and the same commit emptied the four assertions that would have caught a failed-write leftover.

`:252` is the worst of the four: it is the litter guard on the rollback half of the two-file fill act.

Concrete regression this now permits: drop the `try/catch` from `writeLogImage`, let a `rename` fail (full volume), and a multi-MB orphan is left beside the durable log. Full suite stays green.

The repo already solved this and named the failure mode: `packages/preferences/src/orders-reliable.test.ts:117-122` does the `readdir`-over-the-directory check with a comment saying a single `${path}.tmp` assertion "would pass vacuously while a whole generation of temps piled up beside the sidecar." And `packages/preferences/src/sidecar-io-append-failure.test.ts` exists precisely to cover the failure-path cleanup for the sidecar's identical scheme. Neither lesson was carried across.

Fix: convert the four to the `readdir` form, and add one failure-path case (inject a `rename` failure, assert no `.tmp` survives).

### 2. The cleanup can throw and replace the real error — LOW

`apps/tui/src/event-store.ts:368`. `force: true` swallows `ENOENT` only; any other error from `rm` propagates *instead of* the write/rename error.

Verified concretely — data dir becomes non-writable between the temp's creation and the rename (read-only remount, perms change on a mounted data dir):

```
caller sees   : EACCES unlink -> EACCES: permission denied, unlink '…/events.jsonl.74189.1.tmp'
TRUE cause was: EACCES rename  (the log image failed to publish)
```

The operator is pointed at a temp file instead of at "the durable log image failed to publish" — on a spine whose contract is fail-loud-with-the-true-cause. One line:

```ts
await rm(tempPath, { force: true }).catch(() => {});
```

Note `packages/preferences/src/sidecar-io.ts:205` has the same hole in its `finally`, so this is existing house practice rather than a new sin — but this is the line the PR is adding.

### 3. The test's stated reason for its size is empirically false — LOW

`apps/tui/src/event-store.test.ts:383-387` says a tiny payload "completes inside one turn", the writers never overlap, and the test "passes against broken code." Measured against a faithful copy of the pre-fix implementation, same four-writer `Promise.all`:

| payload | old code: any writer rejected | old code: blended/truncated image |
|---|---|---|
| 1 line (67 B) | **10/10** | 0/10 |
| 1 000 lines (~67 KB) | **10/10** | 0/10 |
| 60 000 lines (~4 MB) | 10/10 | **4/10** |

`fs/promises.writeFile` always yields (open → write → close), so all four writers open the shared temp before any rename fires; three renames hit `ENOENT` **at any size**. The `rejected` assertion is red at 67 bytes.

What the multi-MB payload actually buys is the *blend* mode — only at 60 000 lines did the surviving image fail byte-identity. So the honest statement is: the rejection assertion is load-bearing at any size, and only `distinctLines` / byte-identity need the size. As written the comment is the thing that would stop a future maintainer from re-checking, and it's wrong.

**Cost is a non-issue and I would not shrink it** — the whole test runs in **38 ms**. No flakiness risk: 10/10 clean under the fix at every size.

Also trivial: `:389` says "68 bytes/line"; it's 67 (66 + `\n`), so ~4.02 MB not ~4.1 MB.

### 4. Duplicated scheme, with a drift hazard the new docstring drops — optional cleanup

`apps/tui/src/event-store.ts:346-363` reimplements `packages/preferences/src/sidecar-io.ts:69-73` (`tempPathFor`) exactly — same `${path}.${pid}.${counter}.tmp`, same module counter, the same reasoning re-argued from scratch. `apps/tui` already depends on `@numisma/preferences` (`workspace:*`), so sharing would not draw a new package edge.

The hazard is not cosmetic. **Both copies must keep the `.tmp` suffix** or accumulus's `/data/*.tmp` ignore rule (ADR-006) stops covering the temp and a multi-MB staged image becomes committable into the private data repo. `sidecar-io.ts:67` says this explicitly; the new docstring does not mention it at all. At minimum carry that sentence across. Note also that the repo's own duplication policy (`sidecar-io.ts:78-88`) licenses a copy only when the helper carries **zero policy** — this one carries the gitignore-suffix policy, so it doesn't cleanly qualify.

### 5. Orphan litter is now unbounded where it used to be bounded at one — informational

With the fixed name, a hard crash (SIGKILL, power loss) mid-write left exactly one `events.jsonl.tmp`, and the next successful write truncated it. With unique names each crash leaves a *new* multi-MB orphan; `*.tmp` is gitignored in accumulus and the daily-fetch post-check runs `git status --porcelain` **without** `--ignored`, so nothing on the durability path ever surfaces them. This is the right trade — correctness beats tidiness, and sidecar-io accepted the same — but nothing sweeps them. Worth a line in the PR body or a follow-up, not a change here.

---

### Verified sound — checked, not findings

- **Rename is still atomic on every path.** The temp is built by *appending* to `logPath`, so `dirname(temp) === dirname(log)` for every input — dots, spaces, no extension, deep paths. Same filesystem, `rename(2)` atomic. No path through the function escapes to `os.tmpdir()`.
- **The catch cannot delete another writer's temp.** `tempPath` is a `const` local; the counter is incremented before the `try` and captured per invocation.
- **The catch cannot swallow the original error.** It rethrows unconditionally (only #2's masking applies).
- **Rename-succeeded-then-throws.** `rename` is the last statement in the `try`; nothing follows it. Had anything followed, `rm` with `force` is a harmless no-op on the already-consumed name.
- **pid reuse.** A recycled pid restarting the counter at 0 can only collide with litter from a *dead* process, which it then truncates and renames. Pids are unique among live processes, so `(pid, tick)` is unique among live writers.
- **Module reload / vitest worker threads.** These share a pid and each restart the counter at 0 — but every test gets its own `mkdtemp` dir (`event-store.test.ts:157`), so no two share a `logPath`. Not reachable.
- **Counter wrap** (`Number.MAX_SAFE_INTEGER` writes) and **counter shared across log paths** (only widens uniqueness, never narrows it): both fine.
- **Callers.** `appendEvents` (`:324`), `migrateLegacyLog` (`:306`), `restoreLogImage` (`:386`) and the `record-fill` injection (`record-fill-cli.ts:33-34`) all pass `logPath` and never name the temp. No production code, ops script, or config globs `events.jsonl.tmp`. The only `.tmp` couplings in the repo are the four assertions in #1 and `ops/price-feed/run-daily-fetch.sh:25`, which is a comment about `atomic-write.ts`'s separate fixed temp — untouched, and out of scope here.
- **Test-data authorship.** Clean. Inputs are `String.fromCharCode`-generated filler; `makeStore` uses `mkdtemp` plus an authored `genesisSeed()`. Nothing reads a real data directory.
- `src/event-store.test.ts` passes locally: 36/36, 7.3 s.

Out of scope as instructed: the absence of a lock, and the `.claude/` symlink test contamination.

PR #341, opened 2026-08-13
feat(engine): a close is the position's last dated event — ADR-017

AmetAlvirde:
Implements **ADR-017** (`context/adr/ADR-017-a-close-is-the-positions-last-dated-event.md`), which resolves remediation-ledger item 21.

A `PositionTrimmed` or `PositionAddedTo` dated **strictly before** its target position's close date is now **admitted**. The guard stops asking the date-blind `closedPositionIds.has(id)` and asks `asOf >= closedAsOf`.

Closes #300.

## What it was costing

On the pin's own `caseC` — `[Opened 06-05 @ 1 × 100 in c1, Closed 06-10 @ 100, Trimmed 06-08 0.5 @ 50]`:

| | closed rows | realized |
| --- | --- | --- |
| refused (before) | one, at full size | **0** |
| admitted (after) | two, `closedAsOf` `["2026-06-08","2026-06-10"]`, final `{costBasisUsd: 50, proceedsUsd: 100, realizedPnlUsd: 50}` | **50** |

Zero realized profit reported on a position that made fifty, and one whole position closed where half had been sold two days earlier. The operator's only remedy was editing the durable log by hand.

## Exactly two of the four call sites move

All four verified present in `packages/engine/src/events/crossref.ts`. The other two are **not** over-strict and are deliberately untouched:

| line | verb | moves? | why |
| --- | --- | --- | --- |
| `:625` | `InvalidationMarked` | no | breach is derived per OPEN position only, so a mark on a retired position can never be watched whatever its date |
| `:939` | `PositionClosed` | no | whichever close sorts first retires the id; the other hits `if (closing)` and drops |
| `:1025` | `PositionTrimmed` | **yes** | slice 2 |
| `:1117` | `PositionAddedTo` | **yes** | slice 1 |

## Slice 1 — `PositionAddedTo` (`f0abba0`)

`crossReferenceAddedTo` never reads `positionLots` — it checks the funding Reserve's per-tier debit sufficiency and nothing about the position's holdings. **This half needs only the comparison.**

It introduces the carrier both halves share: `EventReference.closedPositionAsOf: Map<string, string>` — retired position id → the `closedAsOf` of the full-close row that retired it (partial rows skipped, earliest close wins). `closedPositionIds` is now derived as `new Set(closedPositionAsOf.keys())`, so **membership and date cannot diverge**, and the two flat sites plus every existing test that spreads or `.has()`es the set are untouched.

## Slice 2 — `PositionTrimmed` (`4704495` red checkpoint, `57031d6` green)

**The work is not the comparison.** `EventReference.positionLots` is built from `folded.positions` (`crossref.ts:459`) — the fold's **survivors** — so a closed position has no entry at all. Two things follow:

1. Relax only the membership test and a backdated trim falls through to the lot-sufficiency gate (`crossref.ts:1054-1063`) with `available = 0`, refused anyway with a *worse* message ("holds only 0") that misdescribes the fund's own history.
2. **The settlement-magnitude gate silently stops firing on exactly this class of event.** `crossref.ts:1083-1084` resolves `last` through `held ? reference.lastClose.get(held.instrumentId) : undefined` and guards on `if (held && last !== undefined)`. An absent `held` does not weaken that check — it **deletes** it, with no error, no warning and nothing going red. A backdated trim would be the one class of trim admitted with **no proceeds sanity check at all**.

The fix is to judge the trim against the world **at its own date**, ADR-015's doctrine one level deeper:

    foldEvents(genesis, accepted.filter((e) => e.asOf <= trim.asOf))

wrapped as `EventReference.worldAsOf(asOf)` — a lazy closure over `genesis` and `priorEvents`, recursing into `buildEventReference` and returning a **full** `EventReference`, not a fragment. `crossReferenceTrim` binds `const world = retiredAsOf === undefined ? reference : reference.worldAsOf(event.asOf)` and reads `positionLots` / `lastClose` off `world`.

This restores `held` **as a side effect**, which is why the magnitude gate keeps firing structurally rather than by a patched conditional. **Ordinary path cost is zero folds** — one closure allocation per reference build; only a trim aimed at an already-retired position ever folds.

## The non-negotiable test, and the proof it is load-bearing

**A backdated trim with absurd proceeds is still rejected.** Without it, consequence (2) ships unnoticed.

Proving it required care, and the first attempt would have fooled us. Stubbing the whole trim path back to naive relaxation turns 9 of 13 red — but the absurd-proceeds case fails via the **sufficiency** gate's "holds only 0", which *masks* the real hole. The sharper isolation — sufficiency reading the as-of world, magnitude alone reading `reference.positionLots` — fails exactly 3 tests with **`rejection` null for 5000 proceeds against an expectation of 50**: the silent admission, with nothing else in the repo complaining. Both stubs were reverted; `git status` is clean.

## The pin inverts, and its two halves part company

`packages/engine/src/position-seal.test.ts:753` — the first test asserted the behavior this reverses and is rewritten to assert acceptance (`rejection` null, all three ids committed). The describe comment is repointed from ledger item 21 to **ADR-017**. The second test at `:772` has a **byte-untouched body** — only its title and comment change, from evidence-for-a-complaint to the positive expectation.

One fixture flipped, inside this lane: `position-backdated-add-to.test.ts:277`, slice 1's own "sites this slice leaves alone" block, which pinned the backdated trim as still refused. That arm is inverted and the block renamed; the second-close and `InvalidationMarked` arms are untouched. Nothing outside `packages/engine` moved.

## Held to the ADR's exclusions

- **No bounded-retroactivity window.** Rejected by name in the ADR — it puts a clock inside a pure fold and makes the same batch admissible on Tuesday and refused on Friday.
- **`EVENT_SCHEMA_VERSION` stays `2`.** No verb added, no record reshaped, no line of `events.jsonl` migrated.
- The refusal message keeps **`already closed`** as a literal substring — `event-ingest.test.ts`, `position-born-by.test.ts`, `position-seal.test.ts` and two TUI suites key off it.

## Carried forward, deliberately not fixed here

**A closed-book row can now move with nothing announcing it.** Today's flat refusal incidentally prevented this; after this lands it is live. Accepted deliberately in the ADR — a book that becomes *right* silently beats one that stays *wrong* silently — and it is exactly what makes **#293** this change's natural companion, the next item in this lane. The first thing a fold diagnostics channel should report is an append that landed behind a close.

## Where a reviewer should look hardest

1. **The `<=` prefix in `worldAsOf` — corrected by review, and the reason matters.** This PR originally shipped a **strict `<`**, argued here as deliberate. That argument was wrong, and the review caught it. Same-dated siblings must be **included**: `foldEvents` orders by `(asOf, then log index)`, and every event in the reference was logged *before* the one under check — `ingest-walk.ts:191` rebuilds from `[...priorEvents, ...accepted]` and the candidate joins `accepted` only once admitted. So a same-dated prior really is applied **first**, and a strict prefix judged against a world the fold never builds. It cost correctness in both directions: three `Trimmed 06-08 0.4` behind a `Closed 06-10` on a 1-unit position were **all committed** (1.2 units sold out of 1, the full-retirement refusal bypassed by splitting the sale); a backdated `AddedTo 06-08` followed by `Trimmed 06-08 1.5` was **falsely refused**; and a trim dated on its position's own open date was refused with exactly the `holds only 0` message this ADR exists to eliminate. `<=` and the guard's `>=` are one boundary read from its two ends — reaching the world means `event.asOf < retiredAsOf` **strictly**, so no prefix can pull the retiring close in, and the event is never in its own prefix. Four regression tests pin the four scenarios; the oversell one goes red on the strict `<`, verified by reverting.
2. **The trim's already-closed refusal now mirrors the add-to's** — same rule, one operator experience, naming both dates and the remedy. `already closed` is kept as a literal substring, so the five suites keying off it are unaffected.

## Merge order — this PR is stacked

**Base is `feature/tighten-four-test-setups` (A1 / #299), not `main`.**

    A1 #299  →  A2 #300 (this PR)  →  A3 #293/#323  →  A4 #297

A1 must merge first; GitHub retargets this PR to `main` automatically when its base branch is deleted on merge.

**`closingIssuesReferences` is expected to come back empty on this PR.** GitHub only registers closing references against the default branch, so the link stays inert until the retarget above happens — at which point the `Closes #300` prose takes effect. This is not a defect and does not need chasing.

## Verification

`pnpm verify` fully green — typecheck clean, **1983 passed / 21 skipped** (the standing Postgres-integration skips), openTUI startup smoke completed.

### DIFF
--- a/packages/engine/src/events/crossref.ts
+++ b/packages/engine/src/events/crossref.ts
@@ -513,6 +513,32 @@ export function buildEventReference(
     });
   }
 
+  // ADR-017's as-of world, LAZY AND MEMOISED BY DATE. Nothing folds until a guard asks,
+  // and today the only caller is a trim aimed at an already-retired position — so every
+  // trim, mark and close on an OPEN position pays exactly one closure allocation and no
+  // fold. The memo is keyed by date, not by event, because that is the whole of the
+  // question: two backdated verbs sharing a date share a world.
+  //
+  // The prefix is STRICTLY BEFORE `asOf`, mirroring the fold's own placement of a
+  // backdated verb: everything dated earlier is already in the world it lands in, and
+  // its same-dated siblings are not (the fold's (`asOf`, then log index) order settles
+  // those, and a gate that pre-applied them would judge against a world the fold has not
+  // built yet). Recursion is not a hazard: the returned reference is built by this same
+  // constructor and its own `worldAsOf` is equally lazy.
+  const worldsAsOf = new Map<string, EventReference>();
+  const worldAsOf = (asOf: string): EventReference => {
+    const memoised = worldsAsOf.get(asOf);
+    if (memoised !== undefined) {
+      return memoised;
+    }
+    const world = buildEventReference(
+      genesis,
+      priorEvents.filter((event) => event.asOf < asOf),
+    );
+    worldsAsOf.set(asOf, world);
+    return world;
+  };
+
   return {
     positionIds,
     reserveIds: new Set(folded.reserves.map((reserve) => reserve.id)),
@@ -532,6 +558,7 @@ export function buildEventReference(
         { instrumentId: position.instrumentId, lots: position.lots },
       ]),
     ),
+    worldAsOf,
   };
 }
 
@@ -1096,12 +1123,34 @@ function crossReferenceTrim(
   if (bornBy) {
     return { ...bornBy, message: `PositionTrimmed ${bornBy.message}` };
   }
-  if (reference.closedPositionIds.has(event.positionId)) {
+  // ADR-017, THE TRIM HALF: "a close is the position's last dated event" is a rule about
+  // DATES, so this refusal must be too. A trim dated STRICTLY BEFORE the close that
+  // retired the position is a correctly-dated trim reported late, and the fold places it
+  // where its date says. `>=` matches the seal rule's strictness in the other direction:
+  // the two rules meet on the close date itself and must not disagree there.
+  const retiredAsOf = reference.closedPositionAsOf.get(event.positionId);
+  if (retiredAsOf !== undefined && event.asOf >= retiredAsOf) {
     return eventError(
       "positionId",
       `PositionTrimmed targets position id '${event.positionId}', which is already closed.`,
     );
   }
+  // AND THE COMPARISON IS NOT THE WORK. Reaching here with `retiredAsOf` set means the
+  // trim is BACKDATED and its target is retired in `reference` — so `positionLots` has no
+  // entry for it. Judged against that world the sufficiency gate below would see
+  // `available = 0` and refuse every backdated trim with a message that misdescribes the
+  // fund's own history; worse, the settlement-magnitude gate is guarded on
+  // `if (held && last !== undefined)` and would SILENTLY STOP FIRING — no error, no
+  // warning, nothing red — leaving the backdated trim the one class of trim admitted with
+  // no proceeds sanity check at all.
+  //
+  // So the remaining gates judge against the world AS OF THE TRIM'S OWN DATE (ADR-015's
+  // doctrine one level deeper: the gate judges against the fold, and for a backdated verb
+  // the relevant fold is the one at the verb's date). `held` comes back as a SIDE EFFECT
+  // of that, which is why the magnitude hole is closed structurally here rather than by
+  // patching its conditional. For an ordinary trim this is `reference` itself — no fold,
+  // no allocation, no change in behavior.
+  const world = retiredAsOf === undefined ? reference : reference.worldAsOf(event.asOf);
   const trimSettlesInto = requireReserveBornBy(
     reference,
     event.settlement.reserveId,
@@ -1111,7 +1160,7 @@ function crossReferenceTrim(
   if (trimSettlesInto.kind === "event-error") {
     return { ...trimSettlesInto.error, message: `PositionTrimmed ${trimSettlesInto.error.message}` };
   }
-  const held = reference.positionLots.get(event.positionId);
+  const held = world.positionLots.get(event.positionId);
   const availableByTier = new Map<CapitalTier, number>();
   for (const lot of held?.lots ?? []) {
     availableByTier.set(lot.tier, (availableByTier.get(lot.tier) ?? 0) + lot.quantity);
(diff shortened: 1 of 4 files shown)


AmetAlvirde (commented):
## Fresh-context review — ADR-017 (PR #341)

Base is `feature/tighten-four-test-setups`; I read only `git diff feature/tighten-four-test-setups...feature/adr-017-backdated-trim` and judged the branch's final state (`57031d6`), not the red checkpoint.

**One material finding**, on the one choice the body itself flags as arguable — and it is arguable in the direction that costs correctness, not style. Details inline; two lower-severity notes follow it.

### What I verified independently, and what held

- **The magnitude hole is genuinely closed.** `world.positionLots` restores `held`, so `if (held && last !== undefined)` fires. I confirmed the load-bearing test fails for the *magnitude* reason and not incidentally: `closeThenTrim(06-08, 0.5, 5000)` passes the sufficiency gate (0.5 ≤ 1 held as of 06-08) and is rejected at `settlement.proceeds`, and the sibling test pins `expected 50.00` / `0.5 × last close 100` / `btc-usd` — figures only computable from a restored `held`. The body's reasoning about the naive stub masking the hole through "holds only 0" checks out.
- **The two non-moving sites really are flat.** `crossref.ts:726` (`InvalidationMarked`) and `:1040` (`PositionClosed`) still ask `reference.closedPositionIds.has(...)`; the diff touches neither, and neither reads `closedPositionAsOf`.
- **The inverted pins are load-bearing.** I regressed the trim guard back to the date-blind `closedPositionIds.has(...)` and `position-seal.test.ts` went red on exactly the rewritten `:753` (1 failed / 25 passed), then reverted. The `:772` test's body is byte-untouched — the file has a single hunk and every line of that body appears as diff context.
- **`closedPositionIds` cannot drift from `closedPositionAsOf`** — it is the map's key set, so the two flat sites and every existing `.has()`/spread caller keep the same answer.
- **Scope is clean.** `EVENT_SCHEMA_VERSION` is still `2` (`events/parse.ts:61`); no wall-clock, no `Date`, no bounded-retroactivity window anywhere in the diff; `already closed` survives as a literal substring in all three messages, so `event-ingest.test.ts`, `position-born-by.test.ts`, `position-seal.test.ts` and the two TUI suites keep matching.
- **`worldAsOf` does not throw on a strict prefix.** `worldAsOf("2020-01-01")` on a live reference returns a well-formed genesis-only world; the as-of-before-genesis guard lives in `requirePositionBornBy`, which `crossReferenceTrim` runs *before* the world is ever built. Recursion is one level deep in practice (nothing calls `world.worldAsOf`), and the extra cost is one fold per backdated trim on top of the walk's existing per-accept fold — no quadratic term.

I did **not** re-litigate the ADR, and I do not count "a closed-book row can move with nothing announcing it" as a defect — deliberately accepted, #293's job.

All probes reverted; `git status` clean.

PR #342, opened 2026-08-13
feat(preferences): a discarded policy line gets a voice, and the push still goes out

AmetAlvirde:
Three members of ADR-004's sidecar class reject a bad line, and they did it in three different ways. Two of them told somebody. `preferences.jsonl` did not — and it is the silent one **on the live push path**.

`loadPreferences` returned a bare `ProfitPolicyEntry[]`, so a caller could not distinguish *"the file held two entries"* from *"the file held five and three were rejected."* If the entry that should have won as-of selection was dropped, the push published a Reserve floor derived from a superseded policy — or suppressed the slot entirely — and nothing anywhere recorded that a line was discarded.

After this increment: an operator who appends a malformed policy line learns about it from the run itself, with the line number and the reason, **and the push still goes out.**

## This is not a regression, and PR #275 did not make it worse

Worth stating plainly so the history reads correctly. #275 made this strictly *better* — a calendar-overflow `effectiveAt` such as `2026-02-30` now falls back to the previous valid entry instead of putting a new policy in force a month early. The silence predates #275 and survived it unchanged. This increment is what gives it a voice.

## What landed, in three slices

**Slice 1 — the contract** (#330). `loadPreferences` returns a `LoadedPreferences` envelope (`{load, entries, skipped}`) mirroring `LoadedPlans`; the pure contracts live in the engine beside `ProfitPolicyEntry` per ADR-001, reusing the existing `LoadOutcome`. `validateProfitPolicyEntry` changes **shape, not verdict** — it returns a discriminated result carrying a closed-vocabulary reason, and a 28-row parity test pins the accept/reject set byte-for-byte unchanged. The loader is now total: a non-`ENOENT` read error becomes `load-failed` instead of a rethrow. Every caller, the barrel and all 13 test sites migrated. Observable behavior at the push was deliberately unchanged — the diff is a contract change.

**Slice 2 — the policy** (#331). `unattendedPreferencesVerdict` turns the envelope into `{exitCode, messages}`, mirroring `unattendedPlansVerdict`. Both shells **upsert the snapshot first and emit the diagnostic after** — ordering asserted by a shared sequence log, not inferred from line order.

**The two shells fold that exit code differently, deliberately.** The push exits non-zero: seam C's warrant is that a launchd job's stderr goes to an unread log, so an exit code is a checked value where a warning is a thing someone must happen to read. The **backfill reports and exits zero**, because `ops/price-feed/run-daily-fetch.sh` runs it under `set -euo pipefail` — a non-zero exit there aborts the nightly script before it stamps completion, and the heartbeat then renders "the daily price job FAILED" on every TUI startup until someone edits the sidecar. A malformed policy line is a **standing fact, not an errand**: it does not extinguish on its own, so a non-zero exit would permanently redden the errand channel and retire it for the next real failure. That trade-off is already resolved the same way in-repo at `gap-report-core.ts` — *a permanently red job is one nobody reads* — and it is the same reasoning lane A's spec applies to its fold verdict. `unattendedPreferencesVerdict` still returns the code either way, so the policy stays a value a test asserts; only the shell's use of it differs, and a test pins each side. Refusal is proven impossible: on a sidecar where *every* line is rejected the snapshot still lands, the Reserve slot suppresses, and no floor is invented. The backfill collapses a per-anchor re-read into one report per run. No diagnostic reaches the projection payload or the glance block — the operator's channel and the phone's channel stay separate.

**Slice 3 — the durable record** (#332). **ADR-020, The Discard Channel**, stating the idiom generally enough that a fold, a parser or an importer can apply it without opening `preferences.ts`. Plus the docs that asserted the silence as an invariant — `packages/preferences/README.md`, ADR-004, ADR-012's citation, and `docs/coverage-rationale.md`, whose entry was re-derived rather than re-worded because slice 1 deleted the lines it cited.

## The idiom

**The Discard Channel — report, never refuse.** Any component that may discard an input must return that discard as part of its own result: not a side effect, not a log line, and never nothing at all. The component reports; the caller decides; and the decision may never be to withhold the fund's daily output.

The counter-rule was already written down at `unattendedPlansVerdict` — *"a plans failure must never kill the NAV fold or withhold a push"* — and it governs here. Availability of the fund's daily view outranks the completeness of any one sidecar. The non-zero exit is what that same docstring's warrant asks for: a launchd job's stderr goes to an unread log, so an exit code is a checked value where a warning is a thing someone must happen to read.

Framed against **ADR-004** (the sidecar class and its per-member on-load validation contract) and the new **ADR-020**.

## Scope guards held

- `validateProfitPolicyEntry`'s **strictness is untouched** — not tightened, not loosened. The issue was the asymmetry between the loaders' *reporting*.
- **`orders.jsonl` does not enter this PR.** #320 §7 audited it: it already satisfies every clause and exceeds clause 2 with an injectable `warn`. The class's asymmetry was one member wide, not two. Nothing filed, nothing changed.
- The three sidecar loaders are **not** refactored into a shared mechanism. Three compliant implementations of a stated contract is the goal.
- Every fixture, example and expected log line is **authored**. None is seeded from real sidecar or push output.

## Cross-lane note — merge order

**This PR merges first.** #320 and lane A's #326 both edit `apps/web/src/push/push-core.ts` and both add prose to the same unattended surface. Lane A merges `feature/preferences-discard-channel` into `feature/fold-discard-channel` — not the other way round.

The surface was left **composable rather than exclusive**, so #326 does not have to rewrite it:

- `AnchorGlance` is a record, not a pair — lane A adds its folded-review key as a sibling.
- `RunReport` (`apps/web/src/push/unattended-report.ts`) files lines under a **kind**, bounded **per kind** with a truncation notice. PR #322's `formatGapReport` starvation lesson is pinned by a test that files 500 preferences lines and proves a single co-tenant line survives.
- **Exit codes never touch the channel** — `RunReport` exposes none, asserted by test. Lane A's deliberately divergent prose-only/exit-zero fold verdict composes by filing prose and folding in no code.

ADR-020 is the artifact #293/#323 were waiting on; they can now cite it by name instead of deriving a second answer to the same design question.

## Review

A fresh-context `pr-review` raised six findings and **all six were real and are fixed in this PR**:

1. **The nightly job would have read as FAILED** — the backfill's non-zero exit under `set -euo pipefail`. Fixed as described above; the wrapper script gains no diff.
2. **`onPreferencesLoad` fired before that anchor's upsert**, so a throwing callback (EPIPE on closed stderr) abandoned the whole replay before a single row was written — and it violated this increment's own report-after-publish rule. Moved after the upsert, pinned by an observed sequence.
3. **`seedDefaultPreferences` would seed over a file it could not read.** Making the loader total silently turned "propagate the read error" into appending a default 10% policy that `pickPolicyAsOf` then selects — the invented read-gap floor its own docstring forbids. It now refuses on `load-failed`.
4. **A path leak in the `load-failed` diagnostic** — Node's raw error carries an absolute path on `EACCES`, which `SIDECAR_NAME` deliberately withholds. Sanitized at the source, so no consumer of the envelope can reintroduce it. The previously passing test used a directory, whose `EISDIR` message happens to carry no path; the leaking branch was the unasserted one.
5. **O(n²) dedup** in the run channel — `Array.includes` over unbounded retention, ~1.6×10⁹ comparisons in the pathological case the class docstring itself names. Now a `Set`.
6. **The backfill's ordering had no test**, unlike the push's, and the test re-implemented the shell rather than exercising it. It now drives the real function.

The reviewer separately verified the parity claim by hand — all eight guards byte-identical against the pre-image — and confirmed channel separation, the dedup key's safety under a mid-run file change, and that fixtures are authored throughout.

## Verification

`pnpm typecheck` green. `pnpm test`: 135 files passed, 4 skipped; 2012 tests passed, 0 failed. The 4 skipped files are the pre-existing `NUMISMA_TEST_DATABASE_URL`-gated integration suites.

Green tests are not the gate. This PR has had its fresh-context review and awaits an explicit go; **it is not to be merged by the authoring session.**

## What this retires

This single PR closes the whole spec, not one slice each.

Closes #276. Closes #320. Closes #330. Closes #331. Closes #332.


packages/preferences/src/preferences.ts
```diff
@@ -198,8 +338,8 @@ export async function seedDefaultPreferences(
   routingReserveId: string,
 ): Promise<ProfitPolicyEntry[]> {
   const existing = await loadPreferences(path);
-  if (existing.length > 0) {
-    return existing;
+  if (existing.entries.length > 0) {
```
AmetAlvirde: **3. Making the loader total silently turned this writer from "propagate the read error" into "seed over a file it could not read".**

Before slice 1, a non-ENOENT read error rethrew out of `loadPreferences` and straight out of `seedDefaultPreferences` — no write happened. Now `load-failed` returns empty buckets, `existing.entries.length` is `0`, and control falls through to the `appendFile` at :346.

**Failure scenario.** `preferences.jsonl` holds the fund's real policy, and the read fails for any non-ENOENT reason — EMFILE under descriptor pressure, EACCES after a permissions change, a transient EIO. `existing.entries` is `[]`, so the seeder appends `defaultProfitPolicyEntry` with today's `effectiveAt`. Because the file is append-only the real entries survive, but `pickPolicyAsOf` now selects the seeded default from that date forward — a **10% floor the fund never set**, published to the phone. That is exactly what this function's own docstring forbids twenty lines above, at :330-333: *"On a read gap the correct behavior is the opposite of this function."* A read gap now routes into it.

The docstring added by this PR says the seeder "reads `entries` and nothing else" as a deliberate choice — which is right about `skipped` and wrong about `load`. Suggest gating on the outcome: `if (existing.load.status === "load-failed") throw …` (or return without writing), so the pre-PR refusal is preserved. Note this is currently reachable only from tests — no production caller — so it is latent, not live.

AmetAlvirde (commented):
Fresh-context review. The contract change is clean and the parity claim holds — I read all eight guards in `validateProfitPolicyEntry` line by line against the pre-image and every predicate is byte-identical; only the `return` grew a reason, and the accepted-entry construction is unchanged. Ordering at the push is genuinely asserted rather than inferred (the shared sequence log in `discard-channel.test.ts`), the missing-file and empty-file cases both exit zero, and no diagnostic reaches the payload or the glance block — `SKIP_DETAIL` is a fixed table and nothing from a rejected line is ever interpolated.

Six findings inline, ranked. One of them (the first) is the only one I'd call a merge blocker: the new non-zero exit is correct at the command, but the one live consumer of that exit code cannot tell it apart from a broken job, and the result is a permanently red daily run.

Deliberately **not** flagged, because the spec settled them: the envelope shape, report-never-refuse, the non-zero exit itself (finding 1 is about how the wrapper reads it, not about downgrading it), `validateProfitPolicyEntry`'s strictness, `orders.jsonl`, the absence of a shared abstraction over the three loaders, and the engine-types-not-re-exported house rule. Also not flagged: the truncation notice sitting outside the per-kind bound (correct — a bound that hid its own notice would read as an all-clear), and the PR body saying "26-row parity test" where the table has 28 rows.

Fixtures are authored throughout — `QUARANTINE_TOKEN`, `ACCEPTED_POLICY`, `priceMarkedLine` marks and the whole `VERDICTS` table are invented values, and I found nothing shaped like real trade data.

PR #344, opened 2026-08-13
recordFill reconciles against the plan that claims the position it opens

AmetAlvirde:
Closes the loop left open by #274: `plans.jsonl` lets the operator declare a plan for a `positionId` **before that position exists**, and until now nothing ever checked the declaration against the fill that made it real. `recordFill` now reconciles the fill against the plan that claims its position, and writes what the operator was told to a seventh durable artifact, `data/reconciliations.jsonl`.

Implements spec #321 (roadmap item **E2**) in five slices on one branch.

---

# ⛔ DO NOT MERGE until one line lands in the private data repo

This PR is **not mergeable on its own**, and the blocker is deliberately outside it.

`reconciliations.jsonl` is now the **seventh** member of `TRACKED_FILES`. ADR-006's `.gitignore` polarity is a **deny-by-default allowlist**, so a file staged but not allowlisted is *"written and silently discarded"* (`apps/tui/src/ingest-commit.ts:40`). The registration is a **three-place change and all three must land together**. Two are in this diff. The third is not, and cannot be — it lives in a different repository:

**Repo:** `accumulus` (the private durable-data sibling, `~/Dev/accumulus`)
**File:** `.gitignore`
**Edit:** add exactly this line to the durable allowlist block, immediately after `!/data/plans.jsonl`:

```
!/data/reconciliations.jsonl
```

That is the entire change — no reordering, no other edit.

**Until it lands, one test fails on purpose.** `apps/tui/src/durable-log-guards.test.ts:169` (end 1 of the durable-file floor) reports `ignored: ["reconciliations.jsonl"]`. **That red is the contract working, not a defect.** It goes green the moment the line above exists. Do not "fix" it here.

## Why a green CI run is not evidence for place 3

`durable-log-guards.test.ts:157-175` asserts both ends of the contract, but **end 1 `skipIf`s when the private data repo is absent** — and it is always absent in CI. So the allowlist end is **verified locally or not at all**.

**It was verified locally, and it failed — which is the finding.** On the authoring machine `accumulusDataDir()` resolved a real git repo, so `skipIf` did **not** fire and **end 1 actually executed** rather than skipping. `git check-ignore -v` attributes `data/reconciliations.jsonl` to `.gitignore:17:/data/*`, with no negation downstream; the allowlist at lines 21–26 names only the existing six. The polarity control was confirmed non-vacuous separately.

Neither the authoring session nor its orchestrator wrote to the `accumulus` repo — that repository was outside this board's scope and no one here was authorized to change it.

---

## What actually closes

This PR closes #274 and its spec #321, and it closes all five slice issues, which were never intended to close individually.

Closes #274
Closes #321
Closes #334
Closes #335
Closes #336
Closes #337
Closes #338

> Note on `Closes #337`: slice 4's acceptance clauses 2 and 5 depend on the `accumulus` line above. It is listed anyway because a `Closes` keyword fires **on merge**, not on push — and merging is already gated on that line. At the only moment the keyword has any effect, its precondition is satisfied. Omitting it would instead leave #337 orphaned and open after the work completing it has landed.

## Merge order

**This item is independent — it can merge in any position.** It shares no file with any other lane on the board and sits in no stacked chain; its base is `main` and it depends on no other open PR. Its only precondition is the `accumulus` line, which is not a PR.

---

## The five slices

| | Slice | Commit |
| --- | --- | --- |
| 1 | engine — the record contract and the pure verdict function | `d0b34a2` |
| 2 | preferences — the IO shell, the total loader, the warning append, the verbatim header | `4efbb7a` |
| 4 | tui — `reconciliations.jsonl` joins the durable allowlist | `b831676` |
| 3 | tui — the advisory reconcile read and the best-effort trail append | `e4850a6` |
| 5 | tui — `pickReconciliationAsOf` and the plans-report annotation | `97edd5e` |

Built in the order 1 → 2 → 4 → 3 → 5, so the file was registered before anything wrote to it.

## The design, as built

**A plan is a declaration of intent, not an authorization.** The reconcile is an advisory read that **can never refuse a fill and never fail one** (D1). The fold stays unable to fail on a plans problem — ADR-004's never-folded discipline.

**Position level only, joining on `positionId`** (D2). There is no rung-level join, and there was never going to be one: order ids come from `synthesizeOrderId({venue, symbol, side, price, observedAt})`, a content hash **including an observation timestamp**, so nothing hand-authored in `plans.jsonl` could ever equal one. `DcaRung.id` describes a join that was **unbuildable from the start** — not deferred, and not blocked on the fills export as #274 originally claimed. It is untouched here.

**A mismatch is exactly two things** (D3): the lot's tier is not in the plan's `tierOrder`, or no plan is in force (absent, or a `noPlan` terminator). Symbol and direction are **unbuildable** — a plan record carries neither `instrumentId` nor `direction`. Size is out by construction, since a ladder fills in pieces. The tier check is **membership**, never drawdown order.

**Nothing flips.** `plans.ts:490` derives `pending`/`active` fresh on every read, so there is no stored state and no flip event to hook. The fill that creates the position is the only available moment, which is where the check went.

**The write is best-effort, after the act, outside the rollback** (D6). The two-file fill act completes and is durable first; the trail append is attempted last; every failure degrades to a loud stderr warn and returns. `record-fill.ts`'s two-file atomicity argument and its one named residual crash window are **entirely intact** — the trail is not part of the act.

**One line per fill, clean or warned** (D7). "Warned then, clean now" is visible either way; **"clean then, warns now"** — a fill consistent with its plan when recorded that now violates it because the plan was narrowed retroactively — is visible only this way. It also disambiguates absence, which D6 makes necessary: a missing line would otherwise mean clean, or write-failed, or never-ran.

**The plans report marks an `active` row whose most recent fill warned** (D8), scoped to the latest fill so a later clean fill clears the mark.

### Why the file is durable rather than printed

The trail passes ADR-006's membership test — *"is this durable, non-re-derivable truth?"* — **only because each line carries its own copy of the declared values as shown**. Plan supersession is by **append**, and `pickPlanAsOf` selects the latest `effectiveAt <= asOf`, so a line appended later but **dated earlier** retroactively changes what re-derivation returns **for a fill already recorded**. A past verdict is therefore not reproducible.

That makes a disagreement between a trail line and a fresh re-derivation **a finding, not a corruption**: a past plan was rewritten after the fact, which is recorded nowhere else.

This is the clause most likely to be "simplified" away later, on the reasonable-sounding grounds that the declared values are already in `plans.jsonl`. They are not — **not as they stood on the day the operator was told**. Anything that drops `declared` also drops this file's right to be tracked. The argument is written at the registration site and in the module header, and per #321 §6 **no ADR was written**: the decision clears only one of `adr`'s three tests.

### A gap must never render as clean

D6 makes gaps possible, and a gap displayed as a clean `active` row is false assurance — worse than not reading the trail at all. So:

- A line is **clean** iff `mismatches` is empty **and** `declared.status` is `active` or `pending`. An empty `mismatches` is necessary but **not sufficient** — a fill recorded while `plans.jsonl` was unreadable would otherwise render as an unqualified clean row.
- The reader is **total** and carries **three** load arms. `absent` is distinct from `load-failed` because this file is **machine-written**: its absence means the writer has never succeeded, which is D6's own failure mode. `absent` exits 0 (before the first fill, absence is normal); `load-failed` and any skip exit 1.
- The row verdict carries **four** reasons to be unknown — `no-trail`, `no-line`, `trail-unreadable`, `plans-were-unreadable` — each with its own operator instruction.
- Clean, warned and each unknown reason **render as visibly different text**. This is asserted over rendered rows, not over verdict arms: six scenarios differing only in trail state produce six distinct row strings, and the clean row is byte-identical to the un-annotated plan row.
- An `unsupported` line renders unknown, never clean — the member this checkout cannot name might be a warning.
- An **unattributable** skip is not a blackout. It is counted file-globally and rendered on its own line, matching the rule `plans-report.ts` already states: attaching it to one position "would turn a single broken line into a fund-wide blackout."

### Ordering

D8 scopes to the latest fill per position, compared on the pair **`(asOf, line)`, lexicographic, `asOf` primary**. `toldAt` is deliberately **not** the ordering key: a fill for the 4th recorded on the 6th must not outrank a fill for the 5th recorded on the 5th. Making `toldAt` the key would silently convert D8 into "the most recent thing the operator was told," which is a different rule. The comparison reuses the existing supersession comparator rather than restating it — `isLater` is now a naming adapter over a single shared body, so there is one comparator, not two with matching semantics.

---

## A spec correction found while building

**#321 §7 and #274 both describe `record-fill-funding.ts:140` wrongly.** They say the operator is asked to type the tier "when the reserve holds a mix." A mixed reserve actually returns `ambiguous` from `deriveFundingTier` and is **rejected** with `ambiguous-tier` — the tier ordering was decided at Transfer time. The prompt at `:140` is reached when the reserve is **untiered** (no lots).

The underlying claim survives and is what mattered: a tier mismatch is **genuinely reachable rather than true by construction**, and both reachable routes are covered by tests — a mistyped operator answer, and funding drawn from a reserve the plan never declared. Only the stated cause of the prompt was wrong.

## Test data

Every fixture and every test input in this PR is **authored**. Nothing is seeded from real output — the line is authorship, not shape. Real trade data lives in the gitignored sibling repo and no part of this work read it. The trail also carries **no figures at all**: no price, size, amount or balance, in the record or in any diagnostic. Diagnostics name a line number, never a line's content.

## Review round

A fresh-context `pr-review` pass found **one blocker and four other findings**, all fixed in `80ea0d8` (correctness) and `116be5c` (cleanups), with every thread resolved.

**The blocker is worth recording, because the increment would have shipped inert and green.** The round-trip echo held `eventId` to the same 64-character cap as `positionId`, but a real `fill:{orderId}@{observedAt}` id runs about 68 characters — so `appendReconciliation` would have warned and written nothing **for every real fill**, `reconciliations.jsonl` would never have been created, and the plans report would have shown no trail forever. The suite passed only because every fixture id is 13 characters and the end-to-end test stubbed the appender.

The cap now applies to `positionId` alone, which is what §4.1 actually binds it to and the only one of the two that renders onto the report page. The regression test drives a **real** `fillEventId(synthesizeOrderId(...))` through the real appender and loader, and a sibling test pins `length > 64` so the fixture cannot regress into the short shape that hid it.

Also fixed: a dead `err` sink could let a rejection escape and fail an **already-durable** fill (D1's absolute); a duplicated `CAPITAL_TIERS` whose drift would have turned every `c4` fill into `trail-unreadable`; a duplicated `byLine`; and a stale `docs/scripts.md:33`.

One finding came from verification rather than the review: `reconciliations-reliable.test.ts` carried raw non-text bytes (a literal U+FEFF and raw `U+0000 U+0001 U+0002`) that made the whole 672-line file classify as `data`, so **grep-based tooling silently skipped it**. Both are escapes now; every changed file on the branch was swept.

## Verification

- `pnpm typecheck` — green, all six packages.
- `pnpm test` — **143 files, 2054 passed, 21 skipped, 1 failed**, and the single failure is the durable-file floor described at the top of this body, red by design until the `accumulus` line lands.
- ~90 tests added across five new test files.


AmetAlvirde (commented):
## Fresh-context review — #344

Read the PR body, spec #321 and idea #274 first, then the whole diff plus the surrounding fill path, the plans selector and `sidecar-io.ts`. Suite reproduced locally: **143 files, 2043 passed, 21 skipped, 1 failed** — exactly the documented deliberate red at `durable-log-guards.test.ts:169`, which I am not reporting.

**5 findings: 1 high, 1 low, 3 cleanups.** Inline where a line is at fault.

| | severity | headline |
| --- | --- | --- |
| C1 | **high** | the `eventId` length cap rejects **every real fill** — the trail is never written |
| C2 | low | a throwing `err` sink escapes `reconcileRecordedFill` and fails an already-durable fill |
| K1 | cleanup | `CAPITAL_TIERS` / `PULL_INSTRUCTION` / `isRecordObject` / `BOM` duplicated from the adjacent `plans.ts` |
| K2 | cleanup | `byLine` duplicated inside `@numisma/engine` |
| K3 | cleanup | `docs/scripts.md:33` no longer describes what `pnpm plans` does |

### K3 — `docs/scripts.md:33` (not in the diff, so no inline)

`pnpm plans` now reads a **third** file and composes **two** exit-code policies. The doc still says *"Exits 0 only when the file loaded and every line was readable"*. After this PR an unreadable `reconciliations.jsonl` exits 1 with a perfectly good `plans.jsonl`, and the page gains a `Trail — <path>` header line and a second unattributable count. `update-docs` territory, not a blocker.

---

## Verified clean — explicit negatives, section by section

These were checked and are genuinely clean. Saying so rather than inventing findings.

**D1 and the ordering in `record-fill.ts` — clean.** The trail call sits at `record-fill.ts:930`, after `appendOrders` resolves (`:889`) and after the `Recorded:` line, outside both `try` blocks, with the rollback's scope already closed. No input can make a plans or trail problem refuse, fail, or roll back a fill: `reconcileRecordedFill`'s whole body is inside one `try`, `lookUpPlanForFill` catches a throwing `loadPlans`, and `appendReconciliation` never throws. `recordFill` returns `{status:"recorded"}` on every reconcile outcome, and `ReconcileTrailIo`'s `Pick` genuinely denies the reconcile any writer of the act. The two-file atomicity argument and the one named residual crash window are untouched — the trail is not in the act. C2 below is the single exception, and it is a throwing *sink*, not a plans problem.

**`isIsoCalendarDate` before `pickPlanAsOf` — clean.** The predicate is cleared at `record-fill.ts:399`, ahead of the only `pickPlanAsOf` call reachable from `recordFill` (`:409`), and it is a guard rather than a `catch`. No other non-total call is on that path: `reconcileAgainstPlan`, `classifyReconciliation` and `isRenderableRecordId` are total; `serializeReconciliationRecord`'s throw is both caught inside `appendReconciliation` **and** pre-empted by the `renderable` check at `:483`. The spy-on-`pickPlanAsOf` test is the right way to pin this — it asserts the ordering rather than the absence of a crash.

**Totality of `loadReconciliations` — clean.** Malformed, torn/truncated, empty, BOM-prefixed, CRLF, absent, a directory (`EISDIR`), permission-denied (`EACCES`), over-large (`ERR_STRING_TOO_LONG`) and arbitrary garbage all land on one of the three arms. `JSON.parse` is wrapped; every `readReconciliationLine` branch is guarded before it indexes; there is no unguarded property or array access anywhere in the reader. I could not construct an input that throws.

**`appendReconciliation` cannot destroy the existing trail — clean.** `appendSidecarLines` is the pre-existing lock + temp + rename path shared with `appendPlan`/`appendOrders`, unchanged here. Traced each failure: a non-ENOENT read failure inside `readOptional` **throws before `writeFile`**, so the truncating rename never happens; a crash between temp write and rename leaves the prior file intact; a rename failure leaves it intact and `rm`s the temp in `finally`; a stale lock throws after 10s having written nothing; two concurrent appends serialize on `open(…, "wx")`. And *"the echo writes nothing"* is literally true — the echo failure returns at `reconciliations.ts:543`, before `appendSidecarLines` is ever reached. Confirmed empirically: after a rejected echo the file is still `absent`, 0 records.

**Clean vs unknown at the pixel — clean.** No state renders a gap as a clean `active` row. `describeVerdict` returns `""` only for `clean`, and `classifyReconciliation` requires **both** an empty `mismatches` **and** `declared.status ∈ {active, pending}`. `absent` → `no-trail`; `load-failed` → `trail-unreadable`; no in-window record → `no-line`, or `trail-unreadable` when an in-window skip names the position; a superseding skip → `trail-unreadable`; `declared.status: "unreadable"` → `plans-were-unreadable`. An **empty** file lands on `no-line`, and an `unsupported` line lands on unknown, never clean. The four qualifiers are four distinct strings, all distinct from `""` and from the warned marker — six distinguishable renderings, and asserting that over rendered rows rather than over verdict arms is the right level.

**The `(asOf, line)` ordering — clean.** `toldAt` appears nowhere as a sort key: it is written, validated and read, and never compared (grepped every non-test occurrence). Selection goes through `isLaterByDateThenLine({date: asOf, line})`, and collapsing `isLater` into a naming adapter over one shared body is the right call — it is the piece where two copies would be a silently wrong answer rather than a slow one.

**Leakage — clean.** No price, size, amount or balance reaches any record, warning, qualifier or diagnostic. Every skip `detail` is a string literal in `reconciliations.ts`; no branch interpolates file content, **including** the unrecognized `mismatches` token, which is deliberately not echoed. Diagnostics name `skip.line`, never a line's content. `declared` excludes `rungs` by construction in both the serializer and the reader, so the rung figures cannot travel. The one thing that does travel is the resolved absolute sidecar path plus the OS error message on `EACCES` — but that is this surface's existing, deliberate convention (`plans-report.ts` already prints `Plans — ${sourcePath}`; `unattendedPlansVerdict` already prints `load.message`), the path is env/homedir-derived rather than operator-supplied, and the stated discipline is about figures and line content. Following the ruling, not deviating from it.

**`positionId` — clean, held at all three gates.** The serializer refuses (`reconciliations.ts:491`); the reader refuses **and** deliberately leaves an unrenderable id off `scraped`, so an id it will not render is also an id it will not attribute by (`:292`); and `record-fill.ts:483` checks before appending and withholds the id from its own stderr label. A newline cannot forge a row on the report page.

**D2 / D3 / D4 / D5 / D7 / D8 fidelity — no deviations found.** Join is `positionId` only and `DcaRung.id` is untouched; the vocabulary is two members derived from one `as const` rather than restated; the tier check is `tierOrder.includes(lotTier)`, membership and not drawdown order; `declared` is required in the type, the serializer and the reader, and its admission argument is written at the `TRACKED_FILES` site; one call site per fill, clean or warned; the report marks `active` rows only, scoped to the latest fill. No ADR, per §6.

---

## Verdict

**Not safe to merge as-is — and the blocker is C1, not the `accumulus` precondition.** The design is faithful to D1–D8 and the hard parts (advisory placement, totality, the clean/unknown split, the single comparator) are right. But C1 means the writer refuses every real fill, so the increment ships inert behind 2,043 green tests. Once C1 is fixed with a test that composes `fillEventId(synthesizeOrderId(...))` through the **real** `appendReconciliation`, and once the `accumulus` line lands, this is good to go. C2 and the cleanups are not merge blockers.

PR #347, opened 2026-08-13
test(web): the pg-free walker follows workspace edges instead of stopping at them

AmetAlvirde:
The module-graph guard that proves `contract.ts` stays pg-free followed relative import edges to any depth but treated **every** bare specifier as a leaf. So `contract.ts` importing `@numisma/engine` ended the walk there, and a pg driver reached **through a workspace package** was invisible to it.

## Why this matters, and what it is not

**ADR-016 makes authentication the disclosure ceiling.** The invariant these guards defend — that the browser bundle carries no pg driver, no connection string, no auth secret — is now what stands between private fund data and the network. That is the reason to close the hole.

**It is not the reason the issue gave, and that correction is half of what this PR delivers.** #295 argues the module-graph walker is "the only guard standing" because the built-asset guard is `describe.skipIf(!hasBuild)`. **That is true locally and false in CI.** Re-verified on this branch: `.github/workflows/ci.yml` runs a `Build web app (ADR-007 — produces the client bundle to inspect)` step invoking `pnpm --filter @numisma/web build` immediately before the test step, and `CLIENT_DIR` resolves to `apps/web/.vercel/output/static`, exactly what `vite build` produces. The authoritative guard runs on every push to every branch and on every PR.

So the issue's suggestion to make the built-asset guard "non-skippable in CI" is **not implemented, because there is nothing to implement** — its skip condition and forbidden-token list are untouched, and `.github/workflows/ci.yml` gains no diff. The real gap was narrower than filed: a local `pnpm test` gave false confidence, and neither guard said which one was authoritative.

## What changed

**1. Workspace specifiers are now followed; true third-party ones are still leaves.**

`reachablePackages` became `reachableGraph`, returning both the files opened and the package specifiers reached — the pg-free assertion reads that specifier set exactly as before, so the existing guarantee is unchanged. What is new is that `@numisma/*` edges are additionally *walked*.

Resolution follows the repo's real layout rather than a naming convention: the workspace groups are **parsed out of `pnpm-workspace.yaml`** rather than hardcoded, each `package.json` in them supplies a `name` and `exports`, and a specifier resolves **through that `exports` map** — `@numisma/engine/calendar` via its `"./calendar"` key, not by guessing at a source path. Conditional targets (`{"types": …, "default": …}`) reduce through `import` → `default` → `node`; wildcard keys match around the star. Anything outside the workspace — `pg`, `react`, `better-auth`, `node:fs` — stays a leaf. **`node_modules` is never opened**, which was the whole point of preferring this fix over the alternative.

**Nothing in the workspace scope can silently become a leaf, and unresolvable edges throw.** A walk that quietly drops an edge is precisely the defect being fixed, so it must not be able to reappear as a silent no-op. Concretely: a specifier naming a scope the workspace publishes but no package that exists throws; a glob shape in the yaml the walker cannot handle throws, because a shape it cannot handle is itself a blind spot; and an `exports` entry it cannot reduce throws with a message distinguishing *absent key* from *target shape I cannot reduce*. The scope set is derived from the names actually found rather than hardcoded to `@numisma/`, so it stays true if the workspace ever publishes a second scope.

**2. Both files now say which guard is authoritative.** Previously neither did, and that ambiguity was the other half of the issue. Each header now states the relationship, so a reader landing in either file learns it without having to find the other:

- `client-bundle.integration.test.ts` scans the **built** assets — it reads what actually shipped, after bundling, tree-shaking and minification, through whatever path the bytes took. It is the **authoritative** boundary.
- `contract.test.ts` walks the module graph in **source**. It is a **fast local pre-check**: it runs with no build, which is exactly why it exists and exactly why it is not what the invariant rests on. It sees static imports only — a dynamic `import()`, a re-export through an installed package, or anything a bundler injects is outside it.

Stated in both: a failure in the pre-check is a real defect, but **a pass is not the invariant proven**, and if the two ever disagree the built-asset scan wins.

## Tests

Three new tests under `the module-graph walker's boundary`. Both halves are pinned, deliberately: a walker that resolves **nothing** fails the first, and a walker that resolves **everything** fails the second (`files` would contain `node_modules` paths, since `pg` drags in pg-pool, pg-protocol and pg-types).

Assertions are **containment, not absence** — the walk must be shown to land on `packages/engine/src/calendar.ts` and to still name `pg`. That matters here more than usual: a walk that silently found nothing would pass an absence-only assertion, and silently finding nothing is the exact failure this PR exists to remove.

The red was staged honestly — the walker was first restructured to expose `files` while keeping the old leaf-stopping behavior verbatim, and the two tests failed for the right reason (the walk stopped at `as-of.ts` and never reached `calendar.ts`). The third test, covering leaf behavior and the absence of `node_modules` paths, passed already, which is correct: that half was never broken and must stay unbroken.

## Verification

The widened walk found **no** forbidden token. The new edge from `contract.ts` is `as-of.ts → @numisma/engine/calendar → packages/engine/src/calendar.ts`, which imports nothing. A broader probe across engine, event-store and preferences reached 53 files across three workspace packages leaving only `node:` builtins as leaves — that probe was temporary and is not in the diff.

`pnpm typecheck` green. `pnpm test`: 133 files passed, 4 skipped; 1964 tests passed, 0 failed. The 4 skipped files are the pre-existing `NUMISMA_TEST_DATABASE_URL`-gated integration suites and the build-gated client-bundle suite.

No production source changed — the diff is two test files.

## Review

A fresh-context `pr-review` raised five findings and **all five were real and are fixed in this PR**:

1. **A silent-skip hole that defeated the widening** — and it contradicted this PR's own "throw, never skip" rule, which is why it mattered most. An unknown `@numisma/*` name returned `null` and was dropped as a leaf, reachable because the workspace groups were hardcoded to `packages`/`apps` while the comment claimed they followed the yaml. The reviewer's scenario: add `services/*` to `pnpm-workspace.yaml`, put a pg-importing `@numisma/ledger` there, and the guard goes green on **exactly the case this PR exists to catch**. Closed at both ends — the groups now come from the yaml, and a workspace-scope specifier that resolves to nothing throws. A drift test re-reads the yaml and asserts the derived globs match it; its teeth were verified by mutation.
2. **A false-red risk in extension resolution.** `moduleFileFor` only rewrote a `.js` suffix and never appended `.ts`, so an extensionless relative import — legal under the `Bundler` resolution this app sets — threw and reddened the guard on a non-defect. A false red on a security guard is expensive: it trains people to distrust it. Resolution is now shape-aware, tries `.ts`/`.tsx` **before** a literal `.js` so a stray emit never beats its source, and strips `?`/`#` query suffixes (live today at `routes/__root.tsx`).
3. **Misnamed failure causes** — conditional export objects and wildcard keys both reported "not in exports map". Failed loud rather than silently, so a diagnosis-quality fix; both are now handled and the two failure messages are distinct.
4. **A substring assertion posing as set membership** — `[...files].sort().join("\n")` with `toContain`. Now `expect(files).toContain(calendar)`.
5. **Duplicated `REPO_ROOT` / `CONTRACT` / `READER` declarations**, hoisted.

The reviewer separately verified that the refactor **does not weaken the original guard** — the pg-free assertion and the reader teeth-check are byte-identical and the specifier set is strictly larger — and confirmed the walk is cycle-safe, bounded, derives its root from `import.meta.url` with no cwd or env dependence, and carries no empty-string env fallback.

Green tests are not the gate. This PR has had its fresh-context review and awaits an explicit go; **it is not to be merged by the authoring session.**

## Merge order

**Independent.** This branch is cut from `main` and is file-disjoint from lane D's other PR (#342, the Discard Channel), so it can merge in any order relative to it and to lane A's stack. It targets `main` directly.

## What this retires

Closes #295.

### DIFF
--- a/apps/web/src/client-bundle.integration.test.ts
+++ b/apps/web/src/client-bundle.integration.test.ts
@@ -15,6 +15,21 @@
  * survive minification: the projection table name and the DB-URL / secret env
  * names. Any hit means server code leaked into the client.
  *
+ * WHICH GUARD IS AUTHORITATIVE. Two tests defend this invariant, and they are not
+ * peers:
+ *
+ *   - THIS one scans the BUILT client assets. It reads what actually shipped —
+ *     after bundling, tree-shaking and minification, through whatever path the
+ *     bytes took. It is the AUTHORITATIVE boundary.
+ *   - `apps/web/src/projection/contract.test.ts` walks the module graph from
+ *     `contract.ts` in source. That is a FAST LOCAL PRE-CHECK, not the real
+ *     boundary: it runs without a build (which is why it exists), it sees static
+ *     imports only, and it stops at true third-party packages. It fails EARLIER
+ *     and more cheaply than this test; it does not fail more authoritatively.
+ *
+ * So: a green pre-check is not this invariant proven. If the two ever disagree,
+ * this one wins.
+ *
  * SUBSTRATE-GATED, like the Postgres integration tests: it needs a build to
  * inspect, so it SKIPS with a loud warning when `.vercel/output/static` is
  * absent (a plain `pnpm test` on an unbuilt tree still passes). CI builds the
--- a/apps/web/src/projection/contract.test.ts
+++ b/apps/web/src/projection/contract.test.ts
@@ -5,10 +5,25 @@
  *
  * The first suite below is the seam's own guard — it asserts, from the module
  * graph rather than from a comment, that `contract.ts` stays pg-free.
+ *
+ * WHICH GUARD IS AUTHORITATIVE. Two tests defend the ADR-007 invariant that no pg
+ * driver, connection string or auth secret reaches the browser, and they are not
+ * peers:
+ *
+ *   - `apps/web/src/client-bundle.integration.test.ts` scans the BUILT client
+ *     assets. It reads what actually shipped, so it is the AUTHORITATIVE boundary.
+ *   - the module-graph walk below is a FAST LOCAL PRE-CHECK, not the real boundary.
+ *     It reads source, so it runs on a plain `pnpm test` with no build — which is
+ *     exactly why it exists, and exactly why it is not what the invariant rests on.
+ *     It reasons about static imports only; a dynamic `import()`, a re-export
+ *     through an installed package, or anything a bundler injects is outside it.
+ *
+ * So: a failure here is a real defect, but a PASS here is not the invariant being
+ * proven. If the two ever disagree, the built-asset scan wins.
  */
-import { readFileSync } from "node:fs";
+import { existsSync, readFileSync, readdirSync } from "node:fs";
 import { fileURLToPath } from "node:url";
-import { dirname, resolve } from "node:path";
+import { dirname, join, resolve } from "node:path";
 import type { CompositionReport } from "@numisma/engine";
 import { describe, expect, it } from "vitest";
 import { fundIdOf } from "./contract.ts";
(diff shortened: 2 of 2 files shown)


AmetAlvirde (commented):
Fresh-context review. The core of the PR holds: `reachablePackages` still returns `reachableGraph(entry).packages`, and the pg-free assertion (`p === "pg" || p.startsWith("pg/")`, plus the `toContain("pg")` teeth check on the reader) is byte-identical to `main` while now reading a strictly larger specifier set — the refactor strengthens the guard and does not weaken it. Cycle safety is fine (dedup by resolved absolute path on dequeue), the walk is cheap (5ms locally, 11 tests green), `REPO_ROOT_DIR` derives from `import.meta.url` with no cwd or env dependence, and no `?? ""`-style empty-env hazard exists here.

Five findings below, most severe first. One is a genuine silent-skip path that the PR's own stated rule ("unresolvable edges throw rather than being skipped") is meant to forbid; one is a false-red the guard can hit on legal `apps/web` source; the rest are robustness and precision.

I did not flag the built-asset guard's `skipIf`, and I did not re-litigate the resolve-vs-`node_modules` choice or the authority ordering.

PR #351, opened 2026-08-14
fix(engine): an invalidation dated after its position's close stops vanishing silently

AmetAlvirde:
`damagePrevented` (`packages/engine/src/position-seal.test.ts`) folds the gated and ungated outcomes and diffs **cash and closed rows**. A vanished `InvalidationLevel` moves neither — so case D proved only that the bogus closed row was prevented, never that the level itself was accounted for. **The one fact the seal rule exists to protect was the one its own test was blind to.**

Closes #297.

## This item was scoped XS. Its premise was wrong.

The plan of record said A3 had already landed the Discard Channel and detects `InvalidationMarked` drops explicitly, so #297 was a one-line assertion on a reported warning. **The channel does not report case D's drop, and could not.**

Case D's batch is `[open 06-05, invalidate 06-20, close 06-10]`. The fold sorts by `asOf`, so the **close applies before the invalidation**, putting the id in the closed book. A3's kind-5 pass filters `!positions.has(id) && !closedBookIds.has(id)` — the id is in the closed book, so the pass correctly stayed silent. Probed directly: `skipped` came back `[]`.

This is not fixture-specific. Case D's shape is definitionally `close(id, early) + invalidate(id, late)` on a position that must exist for the seal rule to fire, so the close always succeeds, the id is always in the closed book, and the pass could never fire. **No case-D fixture reaches a detected case.**

## The contradiction, which is the actual finding

A3 named this residual and accepted it, on the stated ground that **"No observable state moves in that case."**

**#297 exists *because* no observable state moves.** That is its entire complaint. A3's justification for staying silent is, verbatim, #297's reason to speak. **Unobservable is the reason to report, not the reason not to.** The two documents were written against each other and nothing caught it until the code existed.

## What changed — the exactly-decidable sub-case, and only that

The kind-5 post-loop pass now **also** reports an `InvalidationMarked` whose target is in the closed book **and** whose `asOf` is **strictly later** than that id's `closedAsOf`.

**This is a decision, not an inference** — which is why it is admissible where the heuristic the residual comment declines would not be. A mark dated **before** its position's close applied while the position was live and is **no drop**. A mark dated **after** it was discarded. The comparison separates them exactly, from state the fold already holds:

- `latestInvalidation`'s value type gained `asOf` (it already carried `eventId` and `index` from A3's slice A).
- A local `retiredAsOf` map is built beside the existing `closedBookIds`, from **non-partial** `closedPositions` rows only, earliest wins — mirroring `crossref.ts`'s ADR-017 derivation. Partial rows are trims and the fold never deletes on a trim, so the full-close row is the retirement.
- `positions.has(id)` still short-circuits first, and an id absent from **both** books still reports exactly as before.

**A3's cry-wolf guard is intact.** The pass was built around not false-alarming on every legitimately closed position that ever carried a mark; a mark dated *before* its close is still silent, and that is pinned by its own test rather than left to inspection.

## The residual comment was corrected in the same commit

The old paragraph asserted a boundary the code no longer has. Shipping that would have been the exact doc-drift class this board has filed issues about this run — a comment asserting a source of truth the code does not read.

**Both claims in that rewrite were wrong, and review caught them.** It said the comment named "the one residual that genuinely survives," and it described the same-date tie as constrained by state the fold does not carry. Neither holds:

- **Two residuals survive, not one.** `latestInvalidation` is latest-wins per position, so when two marks both land after the close, only the later one is even visible to this pass — the earlier one is dropped and leaves **no record at all**. That is a second silent drop on a closed id, and "exactly one case" was wrong on its own terms.
- **The tie is a siting choice, not a missing-state constraint.** "Only from the close's index, which the closed book does not carry" is true of `ClosedPositionRecord` and false of this function: `order` is in scope in the `PositionClosed` arm, right where the row is pushed. Carrying it in a fold-local map would settle the tie outright. The silence is a consequence of **where** the derivation is sited, not of information the fold lacks.

The comment now names **both** residuals and describes the tie honestly. **Neither detection is widened** — the ruling for this item was the exactly-decidable sub-case only; superseded marks and the same-date tie stay undetected, they are just named accurately instead of misdescribed. A comment asserting a boundary the code does not hold, or a constraint that does not bind, is the same doc-drift class this rewrite existed to fix.

## Review finding: `retiredAsOf` was the wrong operator — fixed

`retiredAsOf` was built **earliest-wins**, mirroring `crossref`'s `closedPositionAsOf`. Review showed that mirror was drawn from the wrong question, and it cried wolf on exactly the property this pass was built around.

When one id carries **two full-close rows** (closed, reopened, closed again), earliest-wins dates a mark against the **first** close — so a mark placed during the **second live period** is reported as a drop, flatly contradicting the pass's own rule that a mark applied while the position was live is no drop.

The two maps legitimately differ and must not be harmonised:

- **A2's `closedPositionAsOf` is a gate looking forward** — "from what date on is this id retired, so which incoming verbs do I refuse?" Earliest is the conservative bound; it errs toward **refusing**.
- **This is a report looking backward** — "on the mark's date, was the id already gone for good?" An id absent from `positions` was retired by its **last** close, so latest-wins is the right operand. It errs toward **silence** (a mark landing in the dead gap between two live periods goes unreported), and this channel prefers a miss to a false alarm.

The comment now says so, so nobody aligns them later.

**Reachability:** a gated log cannot produce two full-close rows for one id (`crossReferenceOpen` refuses a duplicate id). It is reachable through **ungated `foldEvents` callers** — the legacy migration and pre-rule history, which is precisely the population this channel exists for.

**Pinned by test** (`fold-discard-channel.test.ts`): a reused id with two full closes — a mark inside the second live period asserts **no record**, a mark after the last close asserts **exactly one**. Reverting to earliest-wins turns the first arm **red** (1 failed / 7 passed); restored, and `pnpm verify` is green at 2092 passed.

## Tests — two, and the assertion is mutation-verified

**Fold-level, in the channel's own suite** (`fold-discard-channel.test.ts`): one fold, three retired ids differing *only* in mark date — after the close (**reported**, `position-absent`, locator `evt-invalidation-late`), before the close (**not** reported — the cry-wolf guard), and *on* the close date (**not** reported — pins the boundary direction). Asserted with `toEqual` over the whole `skipped` array, so a widening reddens it.

**Seal-level, case D**: asserts the ungated fold's `skipped` contains `{eventId: "evt-invalidate", verb: "InvalidationMarked", reason: "position-absent"}`. Added **alongside** the existing `damagePrevented` assertions; nothing was weakened or removed.

**Mutation check:** inserting `&& false` into the new date comparison turns case D **red** (1 failed / 25 passed). The assertion is load-bearing, not vacuous — verified rather than assumed, because a sibling lane this week shipped a matcher that went green on the very regression it existed to catch.

## Why the projection fork was never viable — recorded so nobody revives it

The issue offered a fallback: project the folded invalidation level for `damagePrevented` to diff against. The roadmap killed that branch, but **for a different reason than the one that actually holds**.

`ingestBatch` returns `committed: []` on rejection, so `damagePrevented`'s gated side folds **nothing**. Any level-valued dimension added to it is `0 − 0 = 0` regardless of implementation — vacuous in precisely the way that file's own docstring warns about. The projection branch could only have escaped that by capturing the level inside the fold *before the close consumes the position*, which is a change to `foldEvents` — not to the harness. It was never the cheap alternative it looked like.

## Merge order — this PR is stacked, and it is the top of lane A

**Base is `feature/fold-discard-channel` (A3 / #350), not `main`.** #350 must merge first — this PR narrows a residual that #350 introduces, so it is meaningless without it.

    lane D #342 → main
    A1 #299 (#333)  →  A2 #300 (#341)  →  A3 #293 (#350)  →  A4 #297 (this PR)

**`closingIssuesReferences` is expected to come back empty.** GitHub only registers closing references against the default branch, so `Closes #297` stays inert until GitHub retargets this PR to `main` when its base is deleted on merge. Not a defect; does not need chasing.

## Verification

`pnpm verify` green — **2091 passed / 21 skipped, 141 files**, openTUI startup smoke completed. A3's digest/verdict equality suite still passes (re-run directly, 8/8): the new detection adds records only on fixtures that suite does not use, and both surfaces still count through `dedupeFoldSkips`.


AmetAlvirde (commented):
**Fresh-context review — diff only (`feature/fold-discard-channel...feature/seal-test-invalidation-level`), base #350 excluded.**

The core of the change holds. The comparison is right and it is pinned in both directions — I mutated it myself rather than trusting the body:

- `&& false` in the new comparison → **2 failed / 31 passed** (case D in `position-seal.test.ts` *and* the new fold-level test). The case-D assertion is load-bearing, not vacuous.
- widening `>` to `>=` → **1 failed** (the fold-level test's on-the-close-date arm). The boundary direction is pinned, not incidental.

A3's cry-wolf guard survives the normal lifecycle: a mark dated before its close stays silent, and a trim-only position never acquires a retirement date (`retiredAsOf` takes non-partial rows only, and a trimmed position is still in `positions`, so the first short-circuit catches it anyway). Full engine + event-store + TUI suites re-run green (882 + 93 + 49), so nothing on the base regressed and both `dedupeFoldSkips` surfaces still agree.

Two findings inline, both low severity — one behavioural, one about the accuracy of the new residual comment. Neither blocks.

---

**On "does it fire on real-shaped data" — worth stating plainly, not a defect.** Nothing today's ingest gate admits reaches the new arm. Both orderings are refused upstream: a mark logged *after* the close is rejected by `crossReferenceInvalidation` (`crossref.ts:781`, via `closedPositionIds`), and case D's shape — a close backdated behind an already-logged mark — is rejected by `requirePositionUntouchedAfter`, which is the very rule case D exists to test. So the live paths are the legacy-migration fold, any direct `foldEvents` caller over pre-seal-rule history, and the ungated harness. That is exactly the split #323 declares ("this rule PREVENTS… the channel REPORTS whatever was already lost"), so the arm is defence-in-depth rather than dead code — but the body's "this is not fixture-specific" is a claim about the *shape*, and it should not be read as a claim that a gated log can produce one.

Fixtures: authored throughout — round numbers, `"authored fixture"` decision strings, no capture from the durable log, no `?? ""` data-dir path anywhere in the diff.

*(Verified in a scratch worktree; every probe reverted, `git status` clean.)*

PR #362, opened 2026-08-18
Fix the class: one gitignore-aware source walker for every repo-wide scan

### LINKED ISSUE #353: two toThrow matchers still pass on the wrong error — the remainder of #299's sweep
Found by lane A while closing #299, and left alone under that issue's scope discipline ("fix the assertions, not the code under test; anything else becomes its own issue"). #299 named three `toThrow(/--as-of/)` sites; these are the same defect at sites it did not name, and #299 closes with its PR, so they would otherwise be orphaned.

## The defect, same as #299 item 3

A regex matcher loose enough to pass on the wrong error. The test asserts that *something* threw mentioning a substring, not that the refusal it names actually fired — so a different failure with overlapping wording satisfies it, and the test keeps passing while the refusal it guards is gone.

## Sites

- `apps/tui/src/startup.test.ts:250` — `toThrow(/cross-reference/)`
- `apps/tui/src/spine-args.test.ts` — the `--magnitude-threshold` matcher set

## Fix

Tighten each to the exact message, as #299's PR did for the `--as-of` family. Mechanical; the pattern to copy is in that PR.

## Why bother, given the tests pass today

This is the thesis #299 was filed on: a test whose title claims more than its assertion checks is worse than no test, because it spends the credibility of a green suite. The `--as-of` sites were not special — they were the ones somebody happened to notice. These two are the remainder of the same sweep.

### LINKED ISSUE #355: the repo-wide contract scan walks gitignored dirs, so any live worktree turns a guard red
Found running the final full-suite verification for the open-issue board, with five git worktrees live under `.claude/worktrees/`.

## Symptom

Two tests in `apps/web/src/calendar-contract.test.ts` fail in the main checkout:

```
× keeps ZERO private copies repo-wide — one declaring file, and it is the engine's
  → addDays must be declared once, in packages/engine/src/calendar.ts:
    expected [ …(6) ] to deeply equal [ 'packages/engine/src/calendar.ts' ]
× keeps ZERO private copies of the cadence table repo-wide
  → VENUE_CADENCE must be declared once, in .../venue-calendar.ts: expected [ …(6) ] ...
```

The six "declarers" are the one real file plus **one copy per live worktree.**

## The proof it is the scan and not the code

**The same test file passes in all five worktrees and fails only in the main checkout.** From a worktree the walk is rooted where no worktrees are nested; from the main checkout it sees every file six times. Identical code, identical assertions, different root.

## Cause

`typescriptSources` (`calendar-contract.test.ts:51-60`) is a raw `readdirSync` walk filtered by a hardcoded denylist at `:41-49`:

```ts
const SKIPPED_DIRS = new Set([
  "node_modules", ".git", ".vercel", ".output", ".nitro", "dist", "coverage",
]);
```

`.claude` is not in it, and nothing consults `.gitignore` — where `.claude/` has been listed all along. So a directory the repo already declares untracked is walked as if it were source.

## Why this is worth fixing rather than working around

This repo now runs **one worktree per concurrent lane** as its normal execution model, and the harness places them at `.claude/worktrees/`. So this is not an exotic configuration — it fires on every parallel run, and it fires on a **guard**, producing a red that is not a defect. A guard that cries wolf whenever the team works in parallel gets muted, and this particular guard exists to stop a shared table being privately re-declared, which is exactly the kind of drift nobody notices by hand.

Note also that vitest's own `--exclude` does not help: this walk is independent of vitest's file collection, so the suite can collect exactly the right 137 files and still fail on a scan of 800.

## Suggested fix

**Drive the walk from `git ls-files` rather than extending the denylist.** It is gitignore-aware by construction, worktree-safe by construction, and cannot silently rot the way a denylist does — every new tool that drops a directory into the repo root is a future false red otherwise. The same argument applies to any other repo-wide scan built on `readdirSync`; worth checking whether this is the only one.

If a raw walk is preferred for some reason, adding `.claude` closes today's instance but not the class.

AmetAlvirde:
Twelve guard tests each walked the repo with their own recursive `readdirSync` filtered by a hand-maintained denylist. A denylist only skips what someone remembered to add, and `.claude/` — line 2 of `.gitignore` — was in none of them. The moment an agent worktree existed under `.claude/worktrees/`, those walks saw the whole tree a second time.

Closes #355 by replacing every one of those walks with a single shared walker driven by `git ls-files`, which is gitignore-aware and worktree-safe by construction rather than by vigilance. Closes #353 by tightening four `toThrow` matchers that were loose enough to pass on the wrong error.

## The observed failure

Reproduced in the main checkout with a worktree deliberately live, before any fix:

```
FAIL src/calendar-contract.test.ts > keeps ZERO private copies repo-wide
AssertionError: addDays must be declared once, in packages/engine/src/calendar.ts
+   ".claude/worktrees/probe/packages/engine/src/calendar.ts",
    "packages/engine/src/calendar.ts",
```

Two files, one declaration, counted twice.

## The headline: the merge gate was measuring the wrong repository

Vitest's own test discovery was not gitignore-aware either. In the main checkout with one worktree live it collected **312 test files where this branch has 156** — every live worktree's copy of the suite ran too.

That is worse than a doubled runtime. With one worktree per lane, each holding a different branch, a repo-root `vitest run` was executing other lanes' unmerged, mid-flight work and reporting the verdict as this branch's. **It is a false green as well as a false red, and the false green is the worse half:** a lane could merge on a suite that never ran its own code, or be turned red by a neighbour's broken WIP. That is the argument for this being Wave 0 — a reader who sees only a config diff will not see what it bought.

Re-measured on the final branch, four ways — the branch now has 157 test files rather than the 156 of the original measurement:

| | contributed excludes on | off |
| --- | --- | --- |
| **no worktree** | 157 | 157 |
| **one probe worktree live** | **157** | **314** |

The bottom-right cell is the bug: 157 of those 314 are the worktree's own copies of the suite. The top row is the other half of the claim — with no worktree live the excludes remove **nothing** that was ever legitimately collectable, so this buys the false-green fix at zero cost to real coverage.

## Discovery asks git what it ignores — it does not parse `.gitignore`

The first version of this change parsed the repo-root `.gitignore` as text and translated its trailing-slash directory form into globs. Review found that wrong in two ways, both of which failed *toward green* — the direction that deletes tests while the shortened suite still reports success. Both are fixed by not parsing at all:

- Its metacharacter filter was a hand-maintained denylist that omitted `{`, `}` and `\`. Git has no brace expansion, so `{tmp,scratch}/` names one literal directory to git; picomatch, which vitest's `exclude` really runs through, expands it and eats every test under any `tmp/` or `scratch/` in the tree. **The guard on the parser shared the bug** — its own character class omitted the same three characters — so the check agreed with the thing it checked, wrongness included.
- It read the root file only, so nested `.gitignore` files never reached discovery. `apps/web/.gitignore` declares `.tanstack/` and `.vercel/`; both exist on disk and matched nothing in vitest's defaults.

Now `git ls-files --others --ignored --exclude-standard --directory --no-empty-directory -z` supplies the answer. Nested `.gitignore` files, negations, anchored paths, `.git/info/exclude` and the global excludes file are handled **by git**, not re-implemented. The globs come out anchored to real paths (`apps/web/.vercel/**`, not `**/.vercel/**`), and metacharacters in real filenames are **escaped rather than detected** — escaping closes the class, where a denylist's correctness depends on the completeness of a character list that was already incomplete once.

The exclude form was chosen over deriving vitest's `include` from `git ls-files`: the include form would need vitest's default include pattern re-implemented as a hand-maintained filter — a new list of exactly the kind this branch retires, failing toward green when a `.spec.ts` silently stops being collected. The exclude form leaves vitest's own `include` authoritative.

Its guard's oracle is **picomatch**, vitest's own matching engine, and its file list is `git ls-files` filtered by vitest's own `defaultInclude` — so `.spec.ts`, `.test.mts` and `.test.js` are visible to it, which the retired two-extension list was not.

## Which scans were actually vulnerable

Only a scan rooted at the repo root can see `.claude/`. Every root was verified:

| scan | root | vulnerable |
| --- | --- | --- |
| `calendar-contract.test.ts` | repo root | **yes — the reproducer** |
| `module-graph.test.ts:32`, `projection/contract.test.ts` | repo root, one level into `apps/`+`packages/` | no — `.claude/` is under neither group |
| six `apps/web/src` guards and seams | `apps/web/src` | no |
| `module-graph.test.ts:294`, `de-prototype.test.ts` | package `src` dirs | no |
| `client-bundle.integration.test.ts` | `apps/web/.vercel/output/static` | **not an instance — see below** |

Two of the "no" rows were safe by range rather than by design: the `apps/web` walks self-defended with a leading-dot skip, while the `de-prototype` and suppression-seam walks had **no** exclusions at all and were safe only because nothing gitignored lives under those roots. That strengthens the case for converting them.

All are converted regardless: the shared walker is the point, and a surviving `readdirSync` is a future instance. **The roadmap enumerated ten sites; a grep found twelve** — `projection/contract.test.ts` and `de-prototype.test.ts` are the two it missed, and both are the same class. Treat that enumeration as a starting point, not a canonical list.

`client-bundle.integration.test.ts` keeps its `readdirSync` deliberately. It walks generated build output, which is gitignored and therefore invisible to `git ls-files` — converting it would silently empty the guard instead of fixing it. The three `apps/price-feed` walks over temp log dirs outside the repo are the same category and are also untouched.

## Every conversion was proved by membership, not by count

Each converted scan had its file list captured before and after and diffed as a list, because a count can match while membership differs. All diffs empty, and independently recomputed during review (`de-prototype` 91/83/30, the import guards 92/29 — empty both ways). For `module-graph.test.ts` the reached-file set (89 files) and the unresolved-import list (empty) are byte-identical across the change, so no module resolution was lost.

Two guard-narrowing risks were caught and closed rather than logged:

- The three import guards declared six extensions (`.ts .tsx .js .jsx .mts .cts`); the walker defaults to two. Membership is identical today, but the class had narrowed. They now pass their original six explicitly.
- `workspacePackageDirs()` hardcoded `apps|packages` — the mirror image of the denylist this issue retires. It now derives the groups from `pnpm-workspace.yaml`.

## Why `ops/testkit/` is the walker's home

Two of these guards take a census of the workspace's own packages, so **the census tool must not itself be a workspace package or it changes the shape it reports on.** A `packages/testkit` would enter `workspacePackageDirs()`'s output and `module-graph.test.ts`'s package map — the walker would start reporting itself. That is a correctness argument, not a stylistic preference, and it was ruled independently by the reviewer as well as by the coordinator.

The cost is a relative-path climb out of each consumer's package, which is ugly but **statically** ugly: every importer is typechecked, so a wrong `../` count is a compile error, never a silent empty scan.

That placement had one consequence nobody had costed: `ops/**` was in no tsconfig `include`, and the root `tsconfig.json` has no `include` and no script running it, so 259 of this branch's lines were never seen by `pnpm typecheck`. `repo-sources.testkit.ts` survived only by being pulled in transitively from included files. `tsconfig.ops.json` now covers `ops/**/*.ts` and `vitest.config.ts` as a seventh `tsc` invocation, verified by mutation.

## The oracle was checking a copy of itself

Review found three parsers of `pnpm-workspace.yaml` in this repo: the shared `workspaceGroupsIn`, a near-copy local to `contract.test.ts`, and an ad-hoc inline parse in that file's test body. The oracle relationship ran between the local copy and the inline parse, so **the shared parser — the one every guard's census runs through — was reachable from no assertion**. Replacing its body with `return ["apps", "packages"]` left the whole suite green.

`contract.test.ts` now imports the shared parser as the **subject**, keeps its hand-written inline parse as the **oracle**, and the middle copy is deleted. That removes a copy and restores the guarantee at once. The comparison is also now order-sensitive: the manifest lists `packages/*` before `apps/*`, and the old assertion sorted both sides, so it would have stayed green against exactly the hardcoded list it exists to catch.

The rule is written down where a DRY pass will read it — next to the inline parse, not only at the top of the file: **an oracle may not share an implementation with its subject, so that inline parse must never become an import.** Everywhere else in that file, importing the shared walker is the right call.

## Silent-empty paths closed

- `repoRelativeDir` validated that a scan root was inside the repo but never that it existed, so a renamed directory returned `[]` where the replaced `readdirSync` threw `ENOENT`. It throws again, which fixes every caller at once. Two guards that had no false-pass floor to notice (`de-prototype.test.ts`, `route-move.test.ts`) now assert a non-empty scanned set, matching the three converted guards that already did.
- `workspaceGroupsIn` returned `[]` on flow-style YAML and on a mis-cased key, despite a docstring promising it throws. An empty group list is an empty workspace census in every guard downstream, and a formatter reformatting the manifest is enough to take that path. It throws, and the docstring is true.
- One of #353's four matchers pinned a string thrown by the test's own stub rather than by production, so the claim that all four were verified against altered production text was **not** true of that one. It now asserts the production behaviour worth asserting — `prepareStartup` propagates an ingest rejection verbatim, without wrapping — using `toThrow(new Error(msg))`, which compares the message for equality rather than substring. The other three matchers and the `--as-of` family's substring convention are untouched.

## Acceptance

- Red reproduced first, with the probe worktree live, on unmodified code (quoted above).
- Green with the probe still live; probe removed; green without it.
- **Discovery re-measured on the final branch**, with and without a live worktree and with the contributed excludes on and off — the table above. The exclusion was also proved end-to-end by planting an authored test file inside `.claude/` and confirming it is collected only when the excludes are disabled.
- Every guard this branch adds or touches was **mutation-tested**: the mutation that should turn it red was named, applied, observed red, and restored. That includes the discovery guard (three separate mutations), the typecheck gate, the workspace-group oracle (two mutations, one of which reds only on ordering — stated as such), the missing-scan-root throw, the empty-census throw, and the startup matcher, whose mutation left the old substring form fully green.
- Three of #353's four matchers were observed failing against a deliberately altered production message before being kept, then the production text restored unchanged. The fourth is described above.

## Recorded, not acted on

- The restored six-extension list omits `.mjs`/`.cjs`. That gap predates this change; the list was restored verbatim rather than improved on.
- `git ls-files` reports a submodule as a single gitlink path, so submodule contents would be invisible to every converted guard. There is no `.gitmodules` here; the walker's header notes it for whoever adds one.
- `--exclude-standard` reads `.git/info/exclude` and the user's global excludes file as well as the committed `.gitignore`, which makes the guard input machine-dependent in principle. Ruled accepted, in the walker's header, with the reasoning.


ops/testkit/repo-sources.testkit.ts
```diff
@@ -0,0 +1,306 @@
...
+ * uses. Anything else THROWS rather than being skipped: an unrecognised glob means
+ * this walker's idea of the workspace is no longer the yaml's, and a census quietly
+ * missing a whole group is the failure this function was rewritten to prevent. A
+ * missing or unreadable manifest throws for the same reason — there is no useful
+ * partial answer, only a confident wrong one.
+ *
+ * Exported as a pure text-to-groups function so a caller with the yaml already in
+ * hand can reuse it.
+ *
+ * ⛔ `apps/web/src/projection/contract.test.ts` CARRIES ITS OWN IDENTICAL COPY, AND
```
AmetAlvirde: **The independence argument does not hold as wired — nothing holds this walker to the manifest.**

This block says `apps/web/src/projection/contract.test.ts` "exists to assert that this walker derives its workspace groups FROM `pnpm-workspace.yaml`". I traced the import graph and it does not.

- `contract.test.ts` never imports `workspaceGroupsIn` / `workspaceGroups` / `workspacePackageDirs`. It imports only `REPO_ROOT` and `repoFiles`.
- Its `WORKSPACE_GROUPS` comes from *its own local copy* of the parser.
- Its assertion `derives its workspace groups from pnpm-workspace.yaml rather than a hardcoded list` (contract.test.ts:488) compares that local copy against a *third* ad-hoc parse written inline in the test body.

So there are three parsers of `pnpm-workspace.yaml` in this repo, and the oracle relationship is `local copy` ⟷ `inline parse`. The shared one here is on neither side of it.

Concretely: replace the body of `workspaceGroupsIn` in this file with `return ["apps", "packages"]` and the entire suite stays green. The only consumer is `workspacePackageDirs()` → `module-graph.test.ts`, and that output is identical today. The regression this comment says the duplication exists to catch is exactly the regression nothing catches.

The fix is one line and it makes the duplication *actually* load-bearing — flip which copy is the subject:

```ts
// contract.test.ts
import { workspaceGroupsIn } from "../../../../ops/testkit/repo-sources.testkit.js";
```

…keeping the inline ad-hoc parse in the test body as the oracle. That reads the manifest independently (it always did), it holds the *shared* parser to the manifest, and it removes a hand-kept copy rather than adding one. As shipped you have the cost of the duplication without the property it was ruled in for.

If you'd rather keep the local copy, then this comment needs to stop claiming a guarantee that doesn't exist, and a real assertion on the shared parser needs to go somewhere.

ops/testkit/repo-sources.testkit.ts
```diff
@@ -0,0 +1,306 @@
...
+ * `.gitignore`, absent from every denylist — so the moment an agent worktree was
+ * checked out under `.claude/worktrees/`, `apps/web/src/calendar-contract.test.ts`
+ * saw the whole repo a second time and failed on "duplicate" declarations that
+ * were the same file viewed twice. Adding `.claude` to the denylists would have
+ * closed that instance and none of the class; the next tool to drop a directory at
+ * the repo root would be the next false red.
+ *
+ * The fix is to stop guessing. `git ls-files --cached --others --exclude-standard`
+ * IS the definition of "the files git considers part of this repo": tracked files,
+ * plus untracked files that are not ignored, with `.gitignore` honoured by
```
AmetAlvirde: Small precision point on the file-set definition, and it matters because this header's rhetorical claim is that `git ls-files --cached --others --exclude-standard` *is* "the definition of the files git considers part of this repo".

`--exclude-standard` is three sources, not one: `.gitignore` files (root **and nested**), `.git/info/exclude`, and `core.excludesFile` (the global ignore). The last two are **per-machine and uncommitted**. That makes the guard input machine-dependent in principle: a developer whose global ignore happens to match something under a scanned root gets a strictly smaller guard input than CI, and the guard passes for them while guarding less.

This is not hypothetical infrastructure — this checkout's own `.git/info/exclude` is non-empty (it carries `**/.claude/worktrees/`, `**/.claude/checkpoints/`, and friends). It happens to be redundant with root `.gitignore` here, so today there is no divergence and I'm not calling this a defect. But the header is the place a future reader forms their model of what the input is, and "with `.gitignore` honoured by `--exclude-standard`" understates it.

Two options, either fine:
- Add a sentence naming the two per-machine sources and ruling the risk accepted.
- Or make it deterministic: `--exclude-per-directory=.gitignore` instead of `--exclude-standard` honours root + nested `.gitignore` and *only* those, dropping `info/exclude` and the global file.

Related, and lower-risk than it sounds: `git ls-files` reports a submodule as a single gitlink path, so submodule contents would be invisible to every converted guard. No `.gitmodules` here, so it's a note for whoever adds one, not a finding.

apps/web/src/projection/contract.test.ts
```diff
@@ -96,6 +101,16 @@ const READER = resolve(HERE, "snapshot-reader.ts");
  * Only the `<dir>/*` glob shape is understood, which is the only shape this repo
  * uses. Anything else THROWS: an unrecognised glob means the walker's idea of
  * the workspace is no longer the yaml's, and that must be loud.
+ *
+ * ⛔ YES, THIS IS A CHARACTER-FOR-CHARACTER COPY OF `workspaceGroupsIn` IN
```
AmetAlvirde: "CHARACTER-FOR-CHARACTER COPY" is not true, and the PR body repeats the claim. `diff` of the two function bodies:

```
-    const glob = /^\s*-\s*["']?(?<glob>[^"'\s]+)["']?\s*$/.exec(line)?.groups?.glob;
+    const glob = /^\s*-\s*["']?(?<glob>[^"'\s]+)["']?\s*$/
+      .exec(line)
+      ?.groups?.glob;
-    if (dir === undefined) {
+    if (!dir) {
-          `the walker cannot see would be missing from every workspace census.`,
+          `the walker cannot see would have its packages treated as third-party.`,
```

Behaviourally equivalent (`[^*]+` can't produce `""`, so `!dir` and `dir === undefined` agree), so nothing is broken. But this is a comment whose entire job is to establish "ruled on, don't touch", and it opens with a verifiable falsehood. The first careful reader who checks it loses confidence in the rest of the paragraph — which is the paragraph you most need them to believe. Either make them actually identical, or drop the "character-for-character" framing and say "the same parser, kept separate deliberately".

apps/web/src/projection/contract.test.ts
```diff
@@ -140,25 +155,33 @@ const workspacePackages = new Map<
...
-      .flatMap((entry) => {
-        const dir = join(groupDir, entry.name);
-        const manifestPath = join(dir, "package.json");
-        if (!existsSync(manifestPath)) return [];
-        const manifest = JSON.parse(readFileSync(manifestPath, "utf-8")) as {
+  WORKSPACE_GROUPS.flatMap((group) =>
+    // The GROUPS stay yaml-derived (a test below locks that); only the per-group
+    // DIRECTORY CENSUS moves onto the shared gitignore-aware listing. The
+    // walker's own `workspacePackageDirs()` would have supplied both, but it
+    // hardcodes `apps|packages` — taking it would silently re-hardcode the group
```
AmetAlvirde: This comment is stale on arrival — it describes the pre-PR state of a function **this same PR fixed**.

`workspacePackageDirs()` no longer "hardcodes `apps|packages`". As of `ops/testkit/repo-sources.testkit.ts` in this branch it calls `workspaceGroups()` → `workspaceGroupsIn(readFileSync("pnpm-workspace.yaml"))`. Taking it would *not* "silently re-hardcode the group list"; it would read the same manifest this file reads.

The conclusion (don't use it here) may still be right — but the reason given is now false, and it's the load-bearing half of the justification. Restate it as the independence argument it actually is, e.g. "using the shared census would make this file's oracle share an implementation with its subject", rather than a claim about hardcoding that the diff two directories over just invalidated.

ops/testkit/gitignore-dir-globs.ts
```diff
@@ -0,0 +1,59 @@
...
+ * only. Negations, anchored or nested paths, wildcard entries, and bare entries
+ * (which may name a file, e.g. the repo's own `.vercel`) are AMBIGUOUS in this
+ * translation, so they are SKIPPED rather than guessed at.
+ *
+ * The asymmetry is the whole design. Under-excluding costs nothing — vitest's own
+ * `defaultExclude` still does its job and at worst a few extra files get
+ * collected. Over-excluding silently removes real tests from the suite and the
+ * suite still passes. When in doubt, emit nothing.
+ */
+export function gitignoredDirGlobs(gitignoreText: string): string[] {
```
AmetAlvirde: **The class is narrower than the PR claims: only the *root* `.gitignore` reaches discovery.**

`vitest.config.ts` reads exactly one file — the repo-root `.gitignore` — so the promise in its header ("the next agent tool to land gets a line in `.gitignore`, not here") holds only for tools that land at the repo root. `git ls-files --exclude-standard`, which the walker next door uses, honours **nested** `.gitignore` files too. The two halves of this PR therefore disagree about what "not this repo's source" means.

This is live, not theoretical. `apps/web/.gitignore` declares five entries in exactly the trailing-slash directory form this parser understands:

```
.output/  .nitro/  .tanstack/  .vercel/  .vinxi/
```

None of them become exclude globs. Cross-checking against `defaultExclude` (`**/node_modules/**`, `**/dist/**`, `**/cypress/**`, `**/.{idea,git,cache,output,temp}/**`, config files): `.output/` is covered by the `{...,output,...}` alternation; `.nitro/`, `.tanstack/`, `.vercel/`, `.vinxi/` are **not covered by anything**. On my checkout `apps/web/.vercel/` (3.7 MB, 91 files) and `apps/web/.tanstack/` both exist right now and vitest walks into them on every run.

I checked: no `*.test.*` / `*.spec.*` files in there today, so collection is still 157 and nothing is currently wrong. But that is a property of what Vite happens to emit, not of the design — and "a build output directory silently doubling the collected suite" is precisely the failure this PR was written to end. The `.claude/` instance is closed; the class is not.

Cheapest honest fix: have the config concatenate every `.gitignore` git knows about, which it can get from the walker it already ships beside —

```ts
// vitest.config.ts
sourceFiles({ extensions: [".gitignore"] })  // or a plain git ls-files
  .flatMap((f) => gitignoredDirGlobs(readFileSync(join(root, f), "utf8")))
```

Note this raises a second question the current parser sidesteps: a nested entry is anchored to *its own directory*, and `**/<name>/**` is broader than that. `**/.vercel/**` is fine (you want it everywhere), but the general form would want `<dirname-of-gitignore>/**/<name>/**`. Worth deciding explicitly rather than by omission.

If you'd rather scope this to a follow-up issue, that's defensible — but then the header comment and the PR body should stop claiming the class is closed, because it's closed for one directory level.

ops/testkit/gitignore-dir-globs.test.ts
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * Raise this ONLY when the branch's real test count has grown well clear of it,
+   * and never to make a red test green — a red here means discovery SHRANK, and
+   * the question to answer is which files went missing and why.
+   *
+   * Set below the 156 measured on this branch with the probe worktree live, with
+   * enough slack that deleting or merging a handful of test files in normal work
+   * does not red the suite, and little enough that losing a directory's worth does.
+   */
+  const FLOOR = 140;
```
AmetAlvirde: **The floor is the weakest of the three assertions here, and the PR body credits it with a catch it cannot make.**

`testFiles` comes from `sourceFiles({ extensions: [".test.ts", ".test.tsx"] })` → `repoFiles()` → `git ls-files`. That set is **completely independent of the exclude globs**. Nothing this file computes about `contributedGlobs` feeds into `testFiles.length`, so `>= FLOOR` cannot detect over-exclusion — the one failure mode the header says the floor exists for ("a floor on how many tests discovery may find", "backed by a discovery floor").

The PR body's verification claim inherits the error:

> The floor was verified by appending `src/` to `.gitignore`: it caught a change that would have deleted all 156 tests while the suite reported success.

Adding `src/` to `.gitignore` does not change `git ls-files --cached` output at all — gitignore has no effect on already-tracked files — so `testFiles.length` stays 157 and this assertion stays green. What actually caught that mutation is the assertion above it, `excludes no test file that git considers part of this branch`. Give the credit to the right test; that one *is* genuinely load-bearing and it's a good design (globs from `.gitignore` text, file set from git's index — two independent inputs, real cross-check).

What `>= 140` actually guards is "the repo still contains at least 140 test files that git knows about". That reds only on a ~17-file deletion event, which is not a silent failure mode. I'd either drop it, or relabel it honestly — its real (small) value is as a smoke test that `repoFiles()` hasn't broken and started returning a short list, which is worth having but is not a discovery floor.

Separately, and I think you're right to have declined it: none of the three assertions measures what vitest *actually collected*. If `include` changed, or a workspace file appeared, or `defaultExclude` grew, discovery could shrink with all three green. Your "WHY NOT SHELL OUT TO `vitest list`" reasoning is sound and I confirmed the adjacent claim — neither `tinyglobby` nor `picomatch` resolves from the repo root under pnpm, so the hand-rolled matcher is justified. Just don't let the floor stand in for a guarantee it doesn't provide.

apps/tui/src/startup.test.ts
```diff
@@ -247,7 +247,7 @@ describe("prepareStartup — ingests, surfaces the report, and wires the fold lo
           throw new Error("cross-reference: unknown instrumentId");
         },
       }),
-    ).rejects.toThrow(/cross-reference/);
+    ).rejects.toThrow("cross-reference: unknown instrumentId");
```
AmetAlvirde: Of the four matchers in #353, this is the one I'd push back on: the string being pinned is thrown by the test's own stub eight lines up, not by production. There is no production message here to have "observed failing against a deliberately altered production message" — the commit body and PR body claim that for all four, and it can't be true of this one.

The assertion that *would* be worth making is "`prepareStartup` propagates an ingest rejection verbatim rather than wrapping it" — a real production behaviour. But `toThrow(<string>)` is substring, which this PR correctly records elsewhere, so this matcher stays green if production starts wrapping as `new Error("startup failed: " + cause.message)`. It cannot detect the only thing it could be testing, and `rejects` alone already proves rejection. Net over the old `/cross-reference/`: nothing.

Worth noting because it also softens the "known cap" in the PR body — vitest *does* have an exact-message form, and it's the same length:

```ts
).rejects.toThrow(new Error("cross-reference: unknown instrumentId"));
```

Passing an Error instance compares `message` for **equality**, not substring. So the cap on #353 is a repo-convention choice (matching the `--as-of` family), not a vitest limitation, and framing it as the latter is what makes a follow-up issue sound bigger than it is. The other three matchers I verified byte-for-byte against `apps/tui/src/spine-args.ts:67` and `:84` — those are correct and genuinely tighter.

AmetAlvirde:
## Independent review

Fresh eyes, no stake in the authoring. I re-derived the load-bearing claims rather than reading the PR body for them. Seven line-specific findings are inline; below are the structural ones, the ruling you asked for on `ops/testkit/`, and what I independently confirmed.

---

### 1. `ops/testkit/` as the home — my read: keep it, but the new code is outside the typecheck gate

**On the location itself: `ops/` is right, and the header's reason is the good one.** The self-reference argument — two guards take a census of the workspace's own packages, so the census tool must not itself be a workspace package or it changes the shape it reports on — is not a stylistic preference, it's a correctness argument, and it holds. Making this `packages/testkit` would put it in `workspacePackageDirs()` output and in `module-graph.test.ts`'s `workspacePackages()` map, i.e. the walker would start reporting itself. The relative-path climb is ugly but it is *statically* ugly: every one of the eleven importers is typechecked, so a wrong `../` count is a compile error, never a silent empty scan. I'd rule this closed in favour of `ops/`.

**But the location has a consequence nobody costed.** `ops/**` is in no tsconfig `include`:

| tsconfig | include |
| --- | --- |
| `packages/*`, `apps/{tui,price-feed}` | `["src/**/*.ts", "tests/**/*.ts"]` |
| `apps/web` | `["src", "vite.config.ts"]` |

`repo-sources.testkit.ts` survives this by accident — it's imported from files that *are* included, so `tsc` pulls it into each program transitively. `gitignore-dir-globs.ts`, `gitignore-dir-globs.test.ts` and `vitest.config.ts` are not: their only importers are each other and the config. **259 of this PR's 785 added lines are never seen by `pnpm typecheck`.** The root `tsconfig.json` has no `include` and no script runs it, so it isn't covering them either.

That is not academic for this particular file. `gitignore-dir-globs.test.ts` does `import rootConfig from "../../vitest.config.ts"` and then indexes `rootConfig.test?.exclude` — exactly the kind of structural access that a vitest major would break, and the break would surface as a runtime `undefined` inside a guard, not as a red typecheck. Cheapest fix: add `"ops/**/*.ts"` and `"vitest.config.ts"` to one package's tsconfig include (or add a seventh `tsc -p tsconfig.json` at the root to the `typecheck` script). Non-blocking, but it should not ship unnoticed.

Note also that `ops/testkit/gitignore-dir-globs.ts`'s own header argues at length that it is *not* testkit — while sitting in a directory named `testkit`. The reasoning is sound and the missing suffix is the honest part, as it says; I'd just observe that the directory name now carries the same untruth the suffix was avoiding.

### 2. The vulnerability table's "no" rows are right, but for a reason the PR doesn't state

The table justifies "no" by root — `.claude/` is under neither `apps/` nor `packages/`. True, but incomplete: the pre-conversion `apps/web` walks *also* self-defended with `if (entry.name.startsWith(".")) return []`, and the `de-prototype` / `suppression-seam` walks had **no** exclusions at all and were safe only because nothing gitignored lives under those roots. Worth knowing that two of the twelve sites were unguarded rather than out-of-range, because it strengthens the case for converting them rather than weakening it.

### 3. `client-bundle.integration.test.ts` — reasoning verified, and it was left alone

Confirmed: not in the changed-file list, still on `readdirSync` at `:112`, still walking `apps/web/.vercel/output/static`. The reasoning holds exactly — `apps/web/.gitignore` has `.vercel/`, so `git ls-files` cannot see a byte of it and conversion would have emptied the guard into a permanent silent pass. Correct call, correctly documented. The three `apps/price-feed` walks (temp log dirs outside the repo) are the same class and are also correctly untouched.

---

### What I independently verified

**Guard membership, both requested sites, computed rather than trusted.** I reproduced each pre-conversion walk verbatim and set-differenced it against the new one on this checkout:

```
de-prototype  packages/engine/src      old 91  new 91   onlyOld []  onlyNew []
de-prototype  apps/tui/src             old 83  new 83   onlyOld []  onlyNew []
de-prototype  apps/price-feed/src      old 30  new 30   onlyOld []  onlyNew []
import-guard  apps/web/src             old 92  new 92   onlyOld []  onlyNew []
import-guard  apps/web/src/push        old 29  new 29   onlyOld []  onlyNew []
```

Empty both ways. The `SELF` exclusion survives correctly (`entry.name !== SELF` → `basename(file) !== SELF`, same semantics), the three-root shape is preserved, and `GUARD_SOURCE_EXTENSIONS` restores the six extensions on all three import guards. I also checked the regex-on-basename → regex-on-absolute-path moves in `module-graph.test.ts:289` and `rung-state-seam.test.ts` — both patterns are `$`-anchored, so widening the subject to the full path is a no-op. No silent narrowing found at any site I checked.

**Discovery numbers are self-consistent.** `vitest list --filesOnly` collects **157**; `git ls-files … | grep -cE '\.test\.tsx?$'` is **157**. The PR body's 156 is the pre-`228fef5` figure and the delta is exactly the test file that commit adds. `312 → 156` is consistent with one live worktree.

**Hygiene, all clean.** `pnpm typecheck` green across all six packages. The eight converted-guard files plus the new discovery test run green (64 tests). No self-attribution in any commit body or in the PR body. No release artifacts — no `chore(release)`, no manifest version bump, no CHANGELOG, no tag; the diff is 17 files, all tests/config. Every fixture in `gitignore-dir-globs.test.ts` is a hand-authored gitignore string; no real trade data anywhere. Commit subjects are Conventional Commits and match the repo's own descriptive-present-tense house style (`fix(engine): a reused position id stops crying wolf`) rather than strict imperative — consistent, so I'd leave it.

**Two things I checked and am *not* raising**, so you don't have to re-check them: `--cached` listing files deleted from the working tree is handled (the `existsSync` filter in `repoFiles()`), and so is the duplicate-path case during an unmerged index (the `Set`). Case sensitivity is a non-issue — both old (`/\.tsx?$/`) and new (`endsWith`) are case-sensitive, so the conversion changes nothing there.

---

### Bottom line

**Mergeable with fixes.** The engineering is real and the central insight is right — replacing twelve hand-maintained denylists with git's own answer is the correct shape, the membership diffs hold up under independent recomputation, and the gitignore-aware discovery change fixes a genuinely dangerous false-green in the merge gate. I did not find a silently-weakened guard.

What I would fix before merge, by severity:

1. **The duplication ruling is wrong as wired** (inline on `repo-sources.testkit.ts:216`). The shared `workspaceGroupsIn` is reachable from no assertion — hardcode it to `["apps","packages"]` and the suite stays green. You are paying for the duplication and not receiving the property it was ruled in for. One-line fix: import the *shared* parser into `contract.test.ts` as the subject and keep the inline ad-hoc parse already in the test body as the oracle. This is the finding I'd hold the merge for, because the whole PR is an argument against facts kept in two places by memory.
2. **Two comments state things that are false** (inline on `contract.test.ts:105` and `:162`) — "character-for-character copy" (they differ in three places) and "`workspacePackageDirs()` hardcodes `apps|packages`" (this PR removed that). Both sit inside the paragraphs meant to stop a future DRY pass. **To answer your question 4 directly: no, these comments do not reliably prevent the collapse** — they are loud enough, but a reader who verifies them finds two falsehoods and a justification that doesn't survive checking, and loud-plus-wrong is weaker than quiet-plus-true.
3. **The class is closed one directory level deep** (inline on `gitignore-dir-globs.ts:45`). Nested `.gitignore` files are honoured by the walker and ignored by discovery. `apps/web/.gitignore`'s `.nitro/ .tanstack/ .vercel/ .vinxi/` are in the exact form the parser understands, are matched by nothing in `defaultExclude`, and two of those directories exist on my checkout right now. Nothing collectable in them today, so latent — but "closed the class" overstates it, and either the code or the claim should move.
4. **`ops/**` is outside every tsconfig** (§1 above) — 259 new lines, including the config the whole suite loads, unchecked.
5. **The floor is credited with a catch it can't make** (inline on `gitignore-dir-globs.test.ts:148`). To answer question 3: the `excludes no test file that git considers part of this branch` assertion is genuinely load-bearing and well-designed; `>= 140` is close to theatre and the `src/` verification in the PR body actually exercised the former.
6. **One of #353's four matchers pins its own stub** (inline on `startup.test.ts:250`) — and `toThrow(new Error(msg))` gives exact-message equality in vitest, so the recorded cap is a convention choice rather than a tool limitation. The other three are correct and verified against production text.

Items 1–3 are substantive; 4–6 are cheap. None of them is a reason to rethink the approach — the walker should ship. I'd want 1 and 2 in this PR, since they are the parts of the change that are themselves about not letting facts drift, and 3 either fixed or explicitly descoped in the body rather than claimed as done.

Not approving and not requesting changes as a formal state — this is a comment-only review.

ops/testkit/gitignore-dir-globs.ts
```diff
@@ -0,0 +1,59 @@
...
+    gitignoreText
+      .split("\n")
+      .map((line) => line.trim())
+      .filter((line) => line !== "" && !line.startsWith("#"))
+      // Trailing-slash directory form only; anything else is skipped (see above).
+      .filter((line) => line.endsWith("/"))
+      // The name itself must be a plain directory name: no negation, no wildcard,
+      // no character class, and no interior separator (which would make it an
+      // anchored or nested path rather than a name matchable anywhere).
+      .filter((line) => !/[!*?[\]/]/.test(line.slice(0, -1)))
```
AmetAlvirde: **This character class has a hole, and it is the exact over-exclusion this module says it exists to prevent.**

`[!*?[\]/]` catches negation, star, question mark, brackets and separator — but not `{`, `}` or `\`. Git has **no** brace expansion, so `{tmp,scratch}/` in a `.gitignore` is one literal directory name. Picomatch, which is what vitest's `exclude` ultimately runs through (tinyglobby → picomatch 4.0.4, present in this lockfile), *does* brace-expand. Verified on this checkout:

```
gitignoredDirGlobs("{tmp,scratch}/\n")  →  [ '**/{tmp,scratch}/**' ]
gitignoredDirGlobs("foo\\bar/\n")       →  [ '**/foo\\bar/**' ]
picomatch("**/{tmp,scratch}/**")("apps/web/src/tmp/a.test.ts")  →  true
```

So one plausible `.gitignore` line silently deletes every test file under any `tmp/` **or** any `scratch/` directory anywhere in the tree, and the shortened suite reports success. That is fail-toward-green, in the one module whose header argues at length that it never guesses in the dangerous direction. The asymmetry the design rests on — "under-excluding costs nothing, over-excluding is silent" — is exactly right; the filter just doesn't implement it.

Fix is one character class: add `{`, `}` and `\`. Or invert to an allowlist, which is the shape the doc actually describes ("a plain directory name") and which can't rot the same way:

```ts
.filter((line) => /^[\w.@+-]+\/$/.test(line))
```

An allowlist also closes whatever the next glob metacharacter turns out to be, which is the same argument this PR makes against denylists everywhere else. Worth noting that the module's own fixtures test `*`, `?`, `[bc]`, `!` and `**` — every metacharacter that *is* in the class, and none that isn't, which is how the gap survived.

ops/testkit/gitignore-dir-globs.test.ts
```diff
@@ -0,0 +1,200 @@
...
+/**
+ * The directory name inside a glob of the one shape this parser emits, or
+ * `undefined` for any other shape.
+ *
+ * A hand-rolled matcher rather than a glob library because there is no glob
+ * matcher resolvable at the repo root, and because narrowness is a feature here:
+ * it refuses anything it does not recognise instead of quietly matching nothing.
+ */
+function dirNameOfGlob(glob: string): string | undefined {
+  return /^\*\*\/(?<dir>[^*?[\]/]+)\/\*\*$/.exec(glob)?.groups?.dir;
```
AmetAlvirde: **The safety net is blind to the gap above it, for the same reason.**

`dirNameOfGlob`'s class `[^*?[\]/]+` omits `{`/`}`/`\` too, so it happily accepts the malformed glob:

```
dirNameOfGlob("**/{tmp,scratch}/**")  →  "{tmp,scratch}"
```

Consequences, in order:

1. `contributes only gitignore-derived directory globs on top of vitest's defaults` (:161) passes — `dirNameOfGlob` returned a string, so the "unrecognised exclude glob" check is satisfied. That assertion's stated job is to red "if the parser ever emits something else"; it does not.
2. `matchesAny` then does a **literal** `segments.indexOf("{tmp,scratch}")`, which never matches a real path.
3. So `excludes no test file that git considers part of this branch` (:167) stays green while picomatch is eating those tests.

This is the structural risk of hand-rolling the matcher: it can only catch mistranslations it happens to share a mental model with, and here it shares the *bug*. I still think declining to shell out to `vitest list` is right, and I confirmed your adjacent claim that no glob matcher resolves from the repo root under pnpm — but `picomatch` **is** resolvable from `ops/testkit/` if you add it as a root devDependency (it's already in the store at 4.0.4 as a transitive vitest dep). Matching with the same engine vitest uses would have caught this and every future variant of it.

If you keep the hand-rolled matcher, at minimum make `dirNameOfGlob`'s class and the parser's class the **same constant**, so a hole can only ever exist in both or neither — right now they are two hand-kept copies of one character set, which is the drift pattern this whole PR is about.

Separately, a smaller version of the same alignment problem at :153: the oracle is `sourceFiles({ extensions: [".test.ts", ".test.tsx"] })`, but vitest's default `include` is `**/*.{test,spec}.?(c|m)[jt]s?(x)`. A `.spec.ts`, `.test.mts` or `.test.js` file is invisible to both assertions, so an over-broad glob eating only those would go unnoticed. None exist today (I checked), so this is a note, not a finding — but the guard should be keyed to the same pattern as the thing it guards.

ops/testkit/repo-sources.testkit.ts
```diff
@@ -0,0 +1,306 @@
...
+ * simply not in `repoFiles()`.
+ */
+export function sourceFiles(options: SourceFilesOptions = {}): string[] {
+  const {
+    dir = REPO_ROOT,
+    recursive = true,
+    as = "repo",
+    extensions = DEFAULT_SOURCE_EXTENSIONS,
+  } = options;
+  const prefix = repoRelativeDir(dir);
```
AmetAlvirde: **A scan root that no longer exists now returns `[]` instead of throwing — every converted guard lost its ENOENT.**

`repoRelativeDir` validates that `dir` is *inside* the repo but never that it *exists*; `sourceFiles` then filters `repoFiles()` by a prefix that matches nothing. Verified:

```
sourceFiles({ dir: "apps/price-feed/srcTYPO", as: "absolute" })  →  []
```

The `readdirSync` walks this replaced all threw `ENOENT` on a missing root. That was a real property: it made a directory rename loud. Now the guard scans zero files and passes.

Two converted sites have no false-pass floor to catch it:

- **`packages/engine/src/de-prototype.test.ts:49`** — `const files = SCAN_DIRS.flatMap(tsFiles)` over three hardcoded absolute roots, then three `expect(offenders).toEqual([])` with nothing asserting `files.length`. Rename `apps/price-feed/src` and that root contributes nothing, silently. Pre-conversion the same rename threw.
- **`apps/web/src/routes/route-move.test.ts:193`** — `expect(extra, extra.join("\n")).toEqual([])` over `allWebSources()`, no non-empty check.

Notably the other converted guards *do* have one — `orders-stay-off-the-wire.test.ts`, `suppression-seam.test.ts` and `rung-state-seam.test.ts` all assert a non-empty scanned set — so this is an inconsistency within the PR, not a repo-wide gap.

Cheapest fix covers all callers at once: throw from `repoRelativeDir` when the resolved directory doesn't exist. It restores the exact loudness that was lost, in the one place that already exists to reject bad `dir` arguments, and it costs one `existsSync`. Adding a `files.length` floor to the two guards above is worth doing regardless — it's the convention the rest of this PR follows.

ops/testkit/repo-sources.testkit.ts
```diff
@@ -0,0 +1,306 @@
...
+ * against a literal `/^(?:apps|packages)\//`, which is the same mistake in miniature
+ * as the `SKIPPED_DIRS` denylist this whole file exists to retire: a second copy of a
+ * fact the repo already states, kept in sync by memory alone. It rots the same way
+ * and it fails the same way — SILENTLY, by shrinking the census rather than erroring,
+ * and a guard handed a short package list passes while guarding less than it claims.
+ * `apps/web/src/projection/contract.test.ts` already refused the hardcoded version
+ * for exactly this reason and kept its own yaml-derived group list, which left two
+ * censuses in one repo disagreeing about where groups come from. The manifest wins.
+ *
+ * ONLY THE `<dir>/*` GLOB SHAPE IS UNDERSTOOD, which is the only shape this repo
```
AmetAlvirde: This paragraph says an unrecognised shape "THROWS rather than being skipped", and that a missing manifest throws too, "because there is no useful partial answer, only a confident wrong one". Agreed with the principle — but there is a third path that returns the confident wrong answer silently: **anything that isn't a block-sequence `packages:` key yields `[]`**, with no throw at all.

```
workspaceGroupsIn('packages: ["packages/*", "apps/*"]')  →  []
workspaceGroupsIn('Packages:\n  - packages/*\n')          →  []
```

Flow style is valid YAML and valid pnpm. Reformatting `pnpm-workspace.yaml` — by hand, or by a formatter — takes this path, and `workspaceGroups()` doesn't assert non-empty either, so `workspacePackageDirs()` then returns an empty census. A guard handed a zero-package list is precisely the failure this function was rewritten to prevent, and the docstring's promise is what would stop a reader from checking.

`if (groups.length === 0) throw` closes it in one line and makes the docstring true. (The duplicated copy in `contract.test.ts` has the identical gap; its own inline oracle would red, and `module-graph.test.ts` would red on unresolved specifiers — but neither error names the cause, so the reader debugs the wrong thing.)

AmetAlvirde:
## Addendum — a second pass found four more, and one of them changes my bottom line

I went back over the two new `ops/testkit` modules with a sharper eye on their *own* failure modes rather than on the conversions. Everything below I reproduced on this checkout with `tsx` before writing it; the details are in four new inline comments.

**The parser can over-exclude after all** (`gitignore-dir-globs.ts:56`). The "no wildcard" character class `[!*?[\]/]` omits `{`, `}` and `\`. Git has no brace expansion, so `{tmp,scratch}/` is one literal directory name to git — but picomatch, which vitest's `exclude` runs through, expands it:

```
gitignoredDirGlobs("{tmp,scratch}/")  →  [ '**/{tmp,scratch}/**' ]
picomatch("**/{tmp,scratch}/**")("apps/web/src/tmp/a.test.ts")  →  true
```

One `.gitignore` line silently deletes every test under any `tmp/` or `scratch/` in the tree and the suite still passes. This is the precise failure the module's header promises it cannot have, in the module the PR singles out as the only one that fails toward green. The existing fixtures pin `*`, `?`, `[bc]`, `!` and `**` — every metacharacter that *is* in the class and none that isn't, which is how it survived.

**And the guard on the parser is blind to it** (`gitignore-dir-globs.test.ts:186`). `dirNameOfGlob` omits the same three characters, so it returns `"{tmp,scratch}"`, the "unrecognised exclude glob" assertion passes, and `matchesAny` then does a literal `indexOf` that never matches — so "excludes no test file that git considers part of this branch" stays green while picomatch eats the tests. The two character classes are hand-kept copies of one set, which is the drift pattern this PR exists to end. This is also the answer to whether the discovery guard is theatre: the *eaten* assertion is well-designed and load-bearing in principle, but as written it shares a bug with its subject, so on this specific class of mistranslation it guards nothing.

**Missing scan roots stopped being loud** (`repo-sources.testkit.ts:177`). `sourceFiles({ dir: "apps/price-feed/srcTYPO" })` returns `[]`; the `readdirSync` walks it replaced threw `ENOENT`. Two converted guards have no false-pass floor to notice — `de-prototype.test.ts:49` (three hardcoded roots, three `toEqual([])`, no `files.length` check) and `route-move.test.ts:193`. Three others in the same PR *do* have one, so it's an internal inconsistency, and one `existsSync` throw in `repoRelativeDir` fixes every caller at once.

**`workspaceGroupsIn` has a silent-empty path its docstring denies** (`repo-sources.testkit.ts:206`). The doc says an unrecognised shape throws because "there is no useful partial answer, only a confident wrong one" — but flow-style YAML returns `[]` with no throw: `workspaceGroupsIn('packages: ["packages/*", "apps/*"]')` → `[]`. That is a zero-package census, silently. One `if (groups.length === 0) throw` makes the docstring true.

### Revised bottom line

I said "mergeable with fixes" and I'd now put the emphasis harder on the *fixes*. The brace hole plus its blind guard is a genuine instance of this PR's own thesis reappearing inside the fix — a hand-maintained character set, kept in two copies, that only skips what someone remembered to add. It is not a reason to rethink the approach; it is a reason not to merge it as the finished article, because the module carrying it is the one the PR asks the reader to trust most.

Merge blockers, revised:

1. The brace/backslash hole and the matching hole in `dirNameOfGlob` — both one-line fixes; ideally one shared constant, or match with `picomatch` itself (resolvable if added as a root devDependency; it's already in the store at 4.0.4).
2. The duplication ruling (my earlier comment on `repo-sources.testkit.ts:216`) — nothing holds the shared parser to the manifest.
3. The two false comments (`contract.test.ts:105` and `:162`).

Then cheap and worth doing in-PR: the `existsSync` throw, the non-empty group check, floors on the two guards that lack them, and `ops/**` into a tsconfig. The nested-`.gitignore` gap and the #353 matcher notes stay as I ruled them — fix or descope explicitly, but stop claiming the class is closed.

Still comment-only: not approving, not requesting changes as a formal state, not merging.

ops/testkit/repo-sources.testkit.ts
AmetAlvirde: @/private/tmp/claude-501/-Users-amet-Dev-notes/e53a6971-82f6-4351-97c4-12a49cd595ea/scratchpad/replies/3800289758

ops/testkit/repo-sources.testkit.ts
AmetAlvirde: Taken as your first option: the header now names all three sources, says plainly that `.git/info/exclude` and `core.excludesFile` are per-machine and uncommitted, and **rules the risk accepted** with the reasoning — this checkout's `info/exclude` is entirely redundant with the root `.gitignore`, the guards also run in CI on a clean clone where only the committed file applies, and the alternative rebuilds by hand the thing this walker exists to retire.

I declined `--exclude-per-directory=.gitignore` for a reason that got stronger during this round: test **discovery** now asks git the same question the walker does (see the top-level thread on the parser). Having the two use different exclude semantics would put them back to disagreeing about what "not this repo's source" means, which is the defect I was fixing one layer up.

The submodule note is in the header too, as a note for whoever adds the first one rather than as a finding.

apps/web/src/projection/contract.test.ts
AmetAlvirde: Fixed by deletion — the copy this comment describes is gone, so the false claim goes with it. You were right that it was the worst possible place for a verifiable falsehood: it opened the one paragraph asking a reader to trust a non-obvious design.

What replaces it states only the independence argument, and states it next to the inline oracle rather than at the top of the file.

apps/web/src/projection/contract.test.ts
AmetAlvirde: Fixed, and further than the comment. You were right that the reason was invalidated by a diff two directories over.

Rather than restate the justification for a hand-rolled duplicate, I removed the duplicate: the per-group census now calls `workspacePackageDirs()`. Leaving the false reason out but keeping the private re-derivation would have left the same third-wheel shape one level down.

Membership verified rather than assumed, since that is this PR's own standard: old logic and new both yield the same 6 package directories, empty diff both ways.

ops/testkit/gitignore-dir-globs.ts
AmetAlvirde: Fixed rather than descoped — and by deleting the parser, so the file this comment sits on no longer exists.

Test discovery now asks git directly: `git ls-files --others --ignored --exclude-standard --directory --no-empty-directory -z`. Nested `.gitignore` files, negations, anchored paths, `info/exclude` and the global excludes file are handled **by git** instead of re-implemented. `apps/web/.tanstack/**` and `apps/web/.vercel/**` are both in the contributed globs now, and pinned by a named assertion.

Your second question — a nested entry being anchored to its own directory, where `**/<name>/**` is broader — is answered by construction rather than by omission: git reports repo-root-relative paths, so the globs come out anchored (`apps/web/.vercel/**`, never `**/.vercel/**`). That is strictly narrower than the old form and no longer a guess about which directory an entry meant.

The header no longer claims the class is closed one level deep, because it no longer is.

ops/testkit/gitignore-dir-globs.test.ts
AmetAlvirde: Agreed on all of it. `FLOOR` is dropped — your diagnosis was exactly right, `testFiles` comes from `git ls-files`, which is completely independent of the exclude globs, so it could not detect over-exclusion at all. The PR body's claim that it caught the `src/` mutation is removed; you were right that the `excludes no test file...` assertion is what caught that, and that assertion survives as the load-bearing one.

Its honest replacement is the opposite direction, which nothing was pinning: **the mechanism is not a no-op.** Demonstrated to be non-redundant rather than assumed — with the glob source stubbed to `[]`, the eaten-test assertion stays green and only the new one reds. A count floor would not have caught that either.

On your separate point that none of the assertions measures what vitest *actually* collected: still declining to shell out to `vitest list` in-suite, for the reason you confirmed. It is covered out-of-suite instead, and reported in the PR body — a four-way measurement (worktree live or not × excludes on or off) plus an authored test file planted inside `.claude/` and observed collected only when the excludes are disabled.

apps/tui/src/startup.test.ts
AmetAlvirde: You were right, and the fix is the one you named. The matcher now uses the Error-instance form, so it asserts the production behaviour actually worth asserting: `prepareStartup` propagates an ingest rejection **verbatim**, without wrapping it. Retitled to say that.

Proved by mutation, and the mutation is the point: wrapping the rejection in `startup.ts` as `new Error("startup failed: " + cause.message)` reds the new form — and leaves the **old** substring form fully green. So the old assertion was testing the stub, exactly as you said.

The PR body's claim that all four matchers were verified against altered production text is corrected; it is now three, with this one described separately.

Scoped as a deliberate local exception with a comment saying why — the other three matchers and the `--as-of` family's substring convention are untouched. I am not filing the repo-wide convention issue: with this matcher fixed, the remaining cap is a style choice rather than a defect, and filing it would queue someone to change a convention nobody has complained about.

ops/testkit/gitignore-dir-globs.ts
AmetAlvirde: This was the finding that changed the shape of the fix. You are right that it is this PR's own thesis reappearing inside the fix, and right that a metacharacter denylist rots exactly the way `SKIPPED_DIRS` did.

I took neither one-line option. The parser is **deleted**; discovery now asks git what it ignores. That removes the character class rather than completing it, and takes the nested-`.gitignore` gap with it.

Where a metacharacter can still appear — a real directory genuinely named `{a,b}` — the fix is your allowlist instinct inverted: **escape, don't detect.** Every entry is escaped before it becomes a glob, so correctness no longer depends on a character list being complete. Pinned with authored fixtures, including the exact case you found: `{tmp,scratch}/` produces `\{tmp,scratch\}/**`, which picomatch matches against the literal braced path and **not** against `tmp/` or `scratch/`.

`picomatch` is now a root devDependency, as you suggested, and is the oracle in the guard.

ops/testkit/gitignore-dir-globs.test.ts
AmetAlvirde: Fixed, and this is the finding I'd most want on record, because "the checker shares an implementation with the thing it checks" is the general form of the whole PR.

`dirNameOfGlob` is gone with the rest of the hand-rolled matcher. The guard's oracle is **picomatch** — vitest's real engine — so it can no longer agree with the subject by construction.

Your `:153` alignment point is fixed in the same move: the file list comes from `git ls-files`, and "is this a test file" now comes from vitest's own `defaultInclude` run through picomatch, instead of a hardcoded `[".test.ts", ".test.tsx"]`. `.spec.ts`, `.test.mts` and `.test.js` are visible to the guard now; they were not before.

Verified by the mutation the old guard could not see: dropping `{`/`}` from the escape set reds the brace fixture. Under the retired guard, the same defect was invisible to both parser and check.

ops/testkit/repo-sources.testkit.ts
AmetAlvirde: Fixed exactly as you proposed — the `existsSync` throw lives in `repoRelativeDir`, so it covers every caller at once, in the place that already exists to reject a bad `dir`. The message names the failure mode rather than just the path, so the reader does not debug the wrong thing.

Both floors added too, since you were right that this was an inconsistency inside one PR rather than a repo-wide gap: `de-prototype.test.ts` now asserts a non-empty scanned set per root as well as in aggregate, and `route-move.test.ts` asserts one over `allWebSources()`.

Mutation-verified in both directions. Typo one of `de-prototype`'s three roots and the whole file now fails to collect on the new throw. Stub a root that *does* exist to return `[]` and only the new floor reds — all three marker guards stay green, which is the false pass you described, demonstrated.

ops/testkit/repo-sources.testkit.ts
AmetAlvirde: Fixed: `if (groups.length === 0) throw`, with a message that names the **shape** as the thing to check, since as you note the reader would otherwise debug the contents.

Both of your reproducers are pinned as authored fixtures — flow style and a mis-cased key — along with the recognised shape still parsing in manifest order. Verified by removing the throw: exactly that fixture reds.

The fixture lives in `contract.test.ts` rather than a new file, because `ops/testkit/` has no unit-test file for the shared parser and that is where the parser is already the subject of an assertion. There is a comment saying so.

The duplicated copy you mention as having the identical gap no longer exists — see the thread on `repo-sources.testkit.ts:216`.

AmetAlvirde:
## Fix round complete — all eleven threads answered and resolved

Both blockers and every cheap item are addressed. Five commits: `5287f96`, `59f8ec1`, `141d5fb`, `e2b8ccf`, `de24191`. The PR body is rewritten; several of its claims were among the things you disproved.

**The three blockers were one defect wearing three coats,** and it was this PR's own thesis turned back on itself: *a checker that shares an implementation with the thing it checks is not a checker.* The discovery guard shared its parser's metacharacter class. The workspace-group oracle checked a copy instead of the shared parser. Both are fixed by removing the shared implementation, not by patching it.

**The parser is deleted.** Discovery asks `git ls-files --others --ignored --exclude-standard --directory --no-empty-directory -z` instead of translating `.gitignore` text. That closes the brace/backslash hole and the nested-`.gitignore` gap in one move, makes the globs anchored to real paths, and puts discovery and the source walker on the same question to the same tool. Where a metacharacter can still reach a glob, entries are **escaped rather than detected** — correctness no longer depends on a character list being complete. The guard's oracle is picomatch, your suggestion, now a root devDependency.

**Every guard this branch adds or touches was mutation-tested** — the mutation named, applied, red observed, restored:

| guard | mutation | result |
| --- | --- | --- |
| eaten-test-files | glob collapsed to its first path segment | RED |
| mechanism-not-a-no-op | glob source stubbed to `[]` | RED (eaten-test assertion stayed **green** — they are not redundant) |
| brace escaping | `{`/`}` dropped from the escape set | RED (invisible to the retired guard) |
| `ops/**` typecheck gate | type error planted in the new module | RED |
| workspace-group oracle | shared parser → `["apps","packages"]` / `["packages"]` | RED on order / RED on membership |
| missing scan root | one of three roots typo'd | RED |
| non-empty floors | a root stubbed to `[]` | RED, with all three marker guards green — the false pass, demonstrated |
| empty-census throw | the throw removed | RED |
| startup matcher | production wraps the rejection | RED — and the **old** substring form stayed fully green |

**Discovery re-measured four ways** on the final branch (157 test files now, vs 156 at the original measurement):

| | excludes on | off |
| --- | --- | --- |
| no worktree | 157 | 157 |
| one probe worktree live | **157** | **314** |

157 of those 314 are the worktree's own copies. The top row matters as much as the bottom: with no worktree live the excludes remove **nothing** that was ever legitimately collectable.

**`pnpm verify` green** on the clean tree, no other lane merged: `154 passed | 3 skipped (157)`, `2283 passed | 46 skipped (2329)`, seven `tsc` invocations, plus the openTUI startup smoke.

Two things deliberately **not** done, both stated in the body rather than claimed as fixed: the six-extension list still omits `.mjs`/`.cjs` (predates this branch), and no repo-wide `toThrow` convention issue is filed — with the one broken matcher fixed, the remaining cap is a style choice.

Thanks for the second pass. The addendum's four findings were the ones that changed the shape of the fix rather than its details.

PR #363, opened 2026-08-18
fix(tui): a contentless durable log survives migration, and the funding interview refuses a missing terminal

### LINKED ISSUE #345: migrate-legacy-log overwrites an existing-but-empty durable log while reporting it did nothing
Found by lane C while writing #294's `migrate-legacy-log.test.ts`. Filing rather than fixing, under #294's tests-only scope discipline — the same rule that produced #329.

## The defect

`migrateLegacyLog` rewrites an **existing-but-empty** `events.jsonl` to a single newline byte, prints `No durable log to migrate`, and exits 0. It reports "no log" about a log it has just overwritten.

## Why it gets through

`readOptional` returns `undefined` **only on ENOENT**. An existing file whose contents are `""` therefore returns `""`, not `undefined`, so it slips past the early return at `apps/tui/src/event-store.ts:209-211`:

```ts
const raw = await readOptional(paths.log);
if (raw === undefined) {
  return { migratedCount: 0, unchangedCount: 0, outputPath: paths.log };
}
```

`""` is falsy but it is not `undefined`, so the guard does not fire, and `writeLogImage` still runs and commits an image over the file.

## Why it matters more than the empty case suggests

The consequence is scoped to a log that is already empty, so no *events* are lost today. What is lost is the file's own identity: an operator who has just created `events.jsonl` and not yet written to it gets it silently rewritten by a command that told them it did nothing. The failure mode is the one this repo treats as most serious — a durable-log write that reports the opposite of what it did — and the guard that is supposed to prevent it is one character away from correct.

## Pinned, not fixed

`migrate-legacy-log.test.ts` carries two **`KNOWN DEFECT`** tests that pin the observable behavior, state plainly that it is wrong, and record that **they must be inverted when this is fixed**. They were written before this issue existed and cite no number; the citation is added in the same change that fixes this.

## Suggested fix

Distinguish "absent" from "empty" at the guard rather than at `readOptional` — an existing empty log should take the same early return as a missing one, and `readOptional`'s ENOENT-only contract should stay as it is, since other callers depend on it.

### LINKED ISSUE #346: import-orders-cli's funding interview is unreachable without a TTY, and fails with a raw readline error
Found by lane C while writing #294's `import-orders-cli.test.ts`. Filing rather than fixing, under #294's tests-only scope discipline.

## The defect

`import-orders-cli.ts` constructs its readline interface at `:28`, at module scope, before any prompt is issued. `createInterface` **eagerly consumes stdin**. With a piped (non-TTY) stdin, the stream has therefore already ended by the time the first `ask` runs, and `rl.question` rejects with `readline was closed` — printed verbatim to the operator with no interpretation.

## Two consequences

1. **Piping answers to `pnpm orders:import` cannot work at all.** Not "works awkwardly" — the first prompt rejects.
2. **`no-reserve-declared` (`apps/tui/src/import-orders.ts:588`) is unreachable non-interactively.** The funding interview is the only path that reaches it, so a refusal the domain layer implements has no way to fire outside a TTY.

## Why it is worth a ticket rather than a shrug

The command is interactive by design and an operator at a terminal is unaffected today, which is why this has gone unnoticed. It matters because it silently rules out the two things a shell command is normally expected to support — scripting it, and testing it end to end through its real entry point — and it announces that with a readline internal error rather than a refusal in the repo's own voice. Any future attempt to automate an import will hit this first and will read as a bug in the caller.

## Suggested direction

Construct the interface lazily, at first `ask`, rather than at module scope; and give a non-TTY stdin an explicit, named refusal instead of letting a readline internal surface. Whether the interview should additionally accept piped answers is a separate product question and is **not** assumed here.

AmetAlvirde:
Two defects in the TUI's durable-log tooling, file-disjoint and shipped together because they share a package and a test run.

Closes #345 and closes #346.

## #345 — the migration overwrote a log it reported as absent

`migrateLegacyLog` rewrote an existing-but-empty `events.jsonl` to a single newline byte, printed `No durable log to migrate`, and exited 0. `readOptional` returns `undefined` **only** on ENOENT, so `""` is falsy but not `undefined` and slipped past the guard: `loadGenesis` ran, the line loop yielded nothing, and `writeLogImage` was still called with `"\n"` — while `migratedCount + unchangedCount === 0` sent the shell down its "no log" branch. The report was a lie about a write that happened, on the one tool in the repo that rewrites the whole durable log.

The fix is at the guard, not at `readOptional`:

```ts
if (raw === undefined || raw.trim().length === 0) {
```

**`readOptional` is deliberately unchanged.** Its ENOENT-only contract is what lets its other callers distinguish "absent" from "present but empty", and folding `""` into `undefined` there would have quieted this path by destroying the distinction that separates an empty log from an unreadable one. "Nothing to migrate" is a policy this migration owns, so it lives here. That reasoning now survives in the guard comment.

**The empty and blank-lines-only cases are one seam, not two.** The predicate is the line loop's own rule hoisted above the write — the loop trims each line and skips the empty ones, so "no content once blank lines are discarded" is exactly the set of inputs that can never yield a record. An `=== ""` check would have fixed half the defect and left a red pin.

### The two `KNOWN DEFECT` pins are inverted, and that is the deliverable

`migrate-legacy-log.test.ts` carried two tests that deliberately characterized the wrong behaviour, and whose own comments said they must be inverted when this was fixed. Both now assert the log is left byte-for-byte untouched — `""` stays `""`, `"\n\n\n"` stays `"\n\n\n"` — with the same stdout sentence, empty stderr, exit 0, and no litter. The block comment above them is rewritten to record the fixed seam rather than deleted, because the `readOptional` argument is what stops someone "fixing" this in the wrong place later. Both now cite #345, which they previously did not.

## #346 — the funding interview was unreachable without a TTY

`import-orders-cli.ts` constructed its readline interface at module scope, before any prompt was issued. `createInterface` eagerly consumes stdin, so on a piped non-TTY stdin the stream had already ended by the first `ask` and `rl.question` rejected with `readline was closed` — a readline internal, printed verbatim to the operator.

Two changes:

1. **The interface is built at the first question, never at startup**, and memoized so the several questions of one interview share one interface. The `finally` is now `rl?.close()`: a run that never prompted does not construct one on the way out just to close it.
2. **A stdin that is no terminal gets told so, once, in the repo's own voice** rather than leaking a readline internal.

**The refusal returns rather than throws, and that is load-bearing.** Throwing would unwind through the shell's outer `catch` and leave `no-reserve-declared` (`import-orders.ts:588`) unreachable forever — the domain's refusal for a batch nobody funded, reached only when `declareFunding` receives an empty batch answer. Returning `""` hands the domain the one answer a run with no terminal can honestly give, and the domain then refuses in its own voice. Two sentences, one story: the shell names *why* there is no answer, the flow names *what it did* about it. Exit 1, nothing written.

This makes `no-reserve-declared` assertable end-to-end **for the first time** — a refusal the domain layer has always implemented with no way to fire outside a TTY. It is now pinned: the named refusal on stderr, the domain refusal on the flow's channel, exit 1, no `orders.jsonl`, no `.tmp`/`.lock` litter.

**No piped-answer support was added.** #346 calls that a separate product question and does not assume it. It never became necessary — the empty-answer composition covers the case without it. If it is ever wanted, it is its own issue.

The `imported-partial` branch still exits 0 and is untouched. Its fifteen-line rationale names a re-trigger — *"if this import is ever automated or piped, the exit code becomes the only surface left"* — and one sentence was added recording that the trigger has **not** fired: a piped run now refuses at the first question, which is the opposite of automating the import.

## Every new and inverted test was proven able to fail

Mutation applied to production code, red observed, restored:

| Mutation | What went red |
| --- | --- |
| Guard reverted to `raw === undefined` | Both inverted pins, on the file-contents assertion (`expected '\n' to be ''` / `to be '\n\n\n'`) |
| Guard narrowed to `raw === undefined \|\| raw === ""` | Blank-lines pin only; the empty-log pin stayed green — the predicted split, which is what proves the two pins hold different things |
| `isTTY` check deleted | The non-TTY case: `expected 13 to be 1` — the unhandled `readline was closed` rejection returns |
| Non-TTY `ask` throws instead of returning `""` | The `no-reserve-declared` assertion precisely — the refusal never reaches the flow |
| Notice moved to shell startup | The "stays silent when nothing is asked" case |

**One honest negative, recorded rather than papered over.** Moving `createInterface` itself back to module scope *with the guard intact* leaves every case green: the `isTTY` guard returns before `rl` is touched, and a spawned stdin is never a terminal, so no test standing outside a TTY can observe the interface's laziness. Laziness is a structural claim here — it is what lets the `finally` close only what was built — not a tested one. What the new cases do hold down is the guard, the notice's placement, and the return-vs-throw composition. Said plainly because a guard nobody can turn red is worth exactly what it can prove.

## Fixtures

Every fixture is authored — invented ids, an invented instrument, round prices, round balances — and `NUMISMA_DATA_DIR` is set to a throwaway directory on every spawned run. Nothing is seeded from, or shaped by, real output.

## Note for reviewers running the suite in a worktree

`ops/testkit/gitignored-path-globs.test.ts` fails two cases inside any fresh worktree and passes in the main checkout. **It is unrelated to this PR** and reproduces on a clean `a5a0cad`. The cause is that the guard asserts against ambient disk contents: it expects `git ls-files --others --ignored` to report `.claude/` and `apps/web/.tanstack/`, and `--others` only lists paths that physically exist, which in a fresh worktree they do not. The walker's behaviour is correct — a worktree with no build output and no nested worktrees genuinely has nothing to exclude. The test is what is environment-dependent. Filed upward to the board rather than fixed here, since it is another lane's file.

### DIFF
--- a/apps/tui/src/event-store.ts
+++ b/apps/tui/src/event-store.ts
@@ -259,6 +259,19 @@ export async function migrateLegacyLog(
   // `undefined` would quiet this path by destroying the distinction that separates an
   // empty log from an unreadable one. "Nothing to migrate" is a policy this migration
   // owns.
+  //
+  // IT RETURNS ABOVE `loadGenesis`, AND THAT PLACEMENT IS THE POINT, not an oversight —
+  // it is a real behaviour change and it is deliberate. An empty log in a data dir whose
+  // `genesis.json` is missing or corrupt now ends 0 with "No durable log to migrate",
+  // where before #345 it ended 1 naming `genesis.json`. That is the guard's own thesis
+  // applied consistently: an ABSENT log has always returned from here, above the genesis
+  // read, so a CONTENTLESS one — which #345 says is the same answer — has to leave through
+  // the same door or the two shapes are not really one seam. Genesis is loaded to
+  // cross-reference records against; with no records there is nothing to cross-reference,
+  // and demanding the operator repair a file this run was never going to read would be a
+  // refusal about nothing. Pinned in `migrate-legacy-log.test.ts` — the case that plants an
+  // empty log in a dir with no genesis at all — which is what goes red if this guard ever
+  // drifts below the read.
   if (raw === undefined || raw.trim().length === 0) {
     return { migratedCount: 0, unchangedCount: 0, outputPath: paths.log, discarded: [] };
   }
--- a/apps/tui/src/event-store.ts
+++ b/apps/tui/src/event-store.ts
@@ -225,6 +225,10 @@ export interface MigrationReport {
  * deliberate one-time reconstruction, NOT a runtime append path — append-only still
  * holds going forward. Idempotent: re-running a clean log migrates nothing.
  *
+ * A log with nothing in it to migrate — absent, empty, or blank lines only — returns a
+ * zero report and touches no disk at all, so the caller's "no durable log" sentence is
+ * true of the bytes as well as of the counts (#345; the guard carries the reasoning).
+ *
  * IT ALSO REPORTS WHAT THE FOLD DROPPED, and this is the path where that matters most:
  * #293 names the legacy migration as the one the ingest gates cannot cover. The
  * cross-reference below runs the fold (`buildEventReference`), so its discards are in
@@ -236,7 +240,26 @@ export async function migrateLegacyLog(
   cashLegs: Map<string, SuppliedCashLeg>,
 ): Promise<MigrationReport> {
   const raw = await readOptional(paths.log);
-  if (raw === undefined) {
+  // A CONTENTLESS LOG IS THE SAME ANSWER AS A MISSING ONE (#345): nothing to migrate,
+  // nothing written, and the file left byte-for-byte as the operator left it.
+  //
+  // Two shapes reach here as "contentless" and they are ONE seam, not two cases: an
+  // existing empty log, and one holding only blank lines. The loop below trims each line
+  // and skips the empty ones, so neither shape can ever yield a record — but without this
+  // guard both still fell through to `writeLogImage` and were replaced by a single
+  // newline byte, while `migratedCount + unchangedCount === 0` sent the shell down its
+  // "No durable log to migrate" branch. The report was a lie about a write that happened,
+  // on the one tool in the repo that rewrites the whole durable log. Hence the predicate
+  // is the loop's own rule hoisted above the write — no content once blank lines are
+  // discarded — rather than an `=== ""` check that would fix half of it.
+  //
+  // THE GUARD BELONGS HERE, NOT IN `readOptional`. That helper maps ENOENT alone to
+  // `undefined`, and its other callers read it for exactly that: "absent" and "present
+  // but empty" are different facts about the disk. Widening it to fold `""` into
+  // `undefined` would quiet this path by destroying the distinction that separates an
+  // empty log from an unreadable one. "Nothing to migrate" is a policy this migration
+  // owns.
+  if (raw === undefined || raw.trim().length === 0) {
     return { migratedCount: 0, unchangedCount: 0, outputPath: paths.log, discarded: [] };
   }
 
--- a/apps/tui/src/import-orders-cli.test.ts
+++ b/apps/tui/src/import-orders-cli.test.ts
@@ -171,7 +171,13 @@ interface ShellRun {
   stderr: string;
 }
 
-describe("import-orders-cli — the shell's own contract (argv, exit codes, env, readline)", () => {
+// NOT "readline", WHICH THIS TITLE USED TO CLAIM. Every case here is a `spawnSync` with a
+// piped stdin, so `process.stdin.isTTY` is falsy in all of them and the `isTTY` guard in
+// `ask` returns before `createInterface` is ever called. No case in this file constructs a
+// readline interface, and therefore none can observe one being closed. What IS pinned is
+// the no-terminal branch in front of it — the notice, its placement at the first
+// unanswerable question, and the empty answer the domain then refuses on.
+describe("import-orders-cli — the shell's own contract (argv, exit codes, env, no-terminal stdin)", () => {
   const createdDirs: string[] = [];
 
   afterEach(async () => {
@@ -226,10 +232,18 @@ describe("import-orders-cli — the shell's own contract (argv, exit codes, env,
   }
 
   /**
-   * The readline lifecycle assertion, made on EVERY path: the process ended on its own.
-   * `rl.close()` lives in a `finally`, so a path that lost it would leave stdin open and
-   * the process alive until the runner's timeout killed it — which is a signal, not a
-   * status.
+   * THE TERMINATION ASSERTION, MADE ON EVERY PATH: the process ended on its own rather than
+   * being killed by the runner's timeout, which arrives as a signal instead of a status.
+   *
+   * IT IS NOT A READLINE-LIFECYCLE ASSERTION, WHICH IS WHAT THIS DOC USED TO IMPLY. The
+   * `finally { rl?.close() }` in the shell is never exercised from here: no case in this
+   * file has a TTY on stdin, so `rl` is `undefined` on every one of them and deleting the
+   * whole `finally` block leaves all of them green. The close is correct — verified by hand
+   * against a real pty — and it is what keeps a REAL interview from leaving stdin open, but
+   * both halves of its lifecycle, the `createInterface` and the `close`, are observable only
+   * from a terminal, and a spawned test cannot stand in one. What this function actually
+   * holds down is that no path here HANGS — which, on a piped stdin, is a claim about the
+   * no-terminal branch returning promptly rather than about an interface being closed.
    */
   function expectExited(run: ShellRun): void {
     expect(run.signal).toBeNull();
@@ -294,8 +308,19 @@ describe("import-orders-cli — the shell's own contract (argv, exit codes, env,
     // And the unreadable row was reported on the error channel, where per-row problems go.
     expect(run.stderr).toContain(`${csv}:2 skipped`);
     // The operator was never taken into the interview: every readable rung was already on
-    // file, so there was nothing to attribute. The funding prompt's own text is the proof.
-    expect(run.stdout).not.toMatch(/Funding reserve for this batch/);
+    // file, so there was nothing to attribute.
+    //
+    // THE PROOF IS THE MISSING NO-TERMINAL NOTICE, not the missing prompt text. Asserting
+    // that `Funding reserve for this batch` never reaches stdout was the obvious probe and
+    // it went vacuous under #346: that string exists only as the argument handed to `ask`
+    // (`import-orders-funding-declaration.ts:68`), and a spawned run takes the no-terminal
+    // branch, which returns "" WITHOUT writing the question anywhere. The absence proved
+    // nothing — it holds on a run that asked and on a run that did not. The notice does
+    // discriminate: `ask` writes it the first time it is called on a stdin that is no
+    // terminal, so stderr staying clean of it is the observable fact that NO QUESTION WAS
+    // PUT AT ALL. Its twin is the no-terminal block below, where a run that DOES reach the
+    // funding question writes exactly that notice — same probe, opposite sign.
+    expect(run.stderr).not.toMatch(/No terminal on stdin/);
   });
 
   it("exits 0 on imported and leaves no temp sibling beside the sidecar it wrote", async () => {
@@ -354,9 +379,11 @@ describe("import-orders-cli — the shell's own contract (argv, exit codes, env,
     const dir = await dataDir();
     const csv = await exportFile(dir, [FRESH_ROW]);
 
-    // `resolveOrdersPath()` throws inside the try, AFTER the readline was constructed —
-    // so this path exercises both the catch (message to stderr, exit 1) and the `finally`
-    // that must still close the prompt.
+    // `resolveOrdersPath()` throws INSIDE the try, so this path exercises the outer catch:
+    // the message reaches stderr and the run exits 1. It does NOT exercise the `finally`'s
+    // `rl?.close()` in any meaningful sense — stdin here is a pipe, no question was ever
+    // put, and `rl` is `undefined`, so the optional call is a no-op. What the `finally`
+    // costs on this path is exactly the nothing the `?.` was written to cost.
     const run = runImport([csv], { dataDir: "data" });
 
     expectExited(run);
@@ -423,8 +450,17 @@ describe("import-orders-cli — the shell's own contract (argv, exit codes, env,
           "the batch — and there is nowhere to conduct it, so every question goes " +
           "unanswered. Run it from a terminal.",
       );
-      // ONCE, not once per question. Repeating it would bury the flow's refusal in copies.
-      expect(run.stderr.match(/No terminal on stdin/g)).toHaveLength(1);
+      // WHAT IT HONESTLY DOES NOT PIN, in the voice this file uses for that: the notice is
+      // written ONCE PER RUN rather than once per question, and no spawn can see it. This
+      // used to carry a `toHaveLength(1)` count over the notice, which could not fail — a
+      // no-terminal run puts EXACTLY ONE question. `import-orders.ts:586` asks the funding
+      // batch question, gets "", and `:588` refuses the whole import; every later question
+      // (the override pass, the rung walk) sits past that refusal, and no export shape,
+      // argv, or data-dir state reaches one. Delete the `toldThereIsNoTerminal` flag
+      // entirely, write the notice unconditionally, and all nine cases here stay green.
+      // The flag is kept anyway — it is a boolean, and the rung walk on the far side of the
+      // funding wall can put thirty questions, which is the shape that makes the difference
+      // between one notice and thirty legible. Counting a one-element set proved none of it.
       // AND THE FLOW'S OWN SENTENCE, on the flow's own channel, in the flow's own voice —
       // `no-reserve-declared`, asserted end-to-end for the first time. Two refusals, two
       // layers, one story: the shell says why there was no answer, the domain says what it
@@ -433,8 +469,19 @@ describe("import-orders-cli — the shell's own contract (argv, exit codes, env,
         "REFUSED — no funding reserve was declared for this batch",
       );
       expect(run.stderr).toContain(`Nothing was written to ${join(dir, "orders.jsonl")}.`);
-      // THE INTERNAL IS GONE. This is the regression the whole change exists to kill, and a
-      // readline constructed anywhere on this path would bring it straight back.
+      // THE INTERNAL IS GONE FROM THIS PATH — the piped, non-TTY one, which is the path
+      // every unattended caller and every test in this file takes, and where a readline
+      // constructed anywhere would bring `readline was closed` straight back.
+      //
+      // SAID PRECISELY, BECAUSE THE BROADER CLAIM IS FALSE. This is not "no readline
+      // internal ever reaches the operator again". At a REAL terminal, Ctrl-D on the
+      // funding question still surfaces Node's own `Aborted with Ctrl+D` on stderr with
+      // exit 1 — no shell sentence, no `no-reserve-declared`. Reproduced by hand on this
+      // commit (`printf '\004' | script -q /dev/null tsx …`). A BLANK LINE at the same
+      // terminal refuses correctly, in both voices, so this is a gap in the TTY branch
+      // rather than a regression of the fix: what #346 killed is a piped run reporting a
+      // readline internal instead of naming the missing terminal, and that is dead. The
+      // Ctrl-D case is tracked separately; nothing here should be widened to cover it.
       expect(run.stderr).not.toMatch(/readline was closed/);
       // Refused means refused: no sidecar, and no staging sibling of one either.
       expect(await readdir(dir)).not.toContain("orders.jsonl");
--- a/apps/tui/src/import-orders-cli.test.ts
+++ b/apps/tui/src/import-orders-cli.test.ts
@@ -23,7 +23,9 @@
  * door, the same shape `record-fill-cli.test.ts` uses. `input` is always supplied so the
  * readline prompt gets EOF or an authored answer instead of hanging, and every runner call
  * carries a `timeout` so a shell that failed to close its prompt surfaces as a KILLED
- * process rather than as a stuck suite.
+ * process rather than as a stuck suite. A spawn's stdin is a PIPE, never a terminal, which
+ * is not merely an incidental property of this harness — since #346 it is the condition the
+ * last block below tests, and the reason `no-reserve-declared` is assertable at all.
  *
  * EVERY FIXTURE IS AUTHORED — invented ids, an invented instrument, round decade prices,
  * round balances. Nothing is seeded from, or shaped by, a real export, a real ladder or a
@@ -368,9 +370,10 @@ describe("import-orders-cli — the shell's own contract (argv, exit codes, env,
     // so running the same refusal under two different values shows the resolution
     // FOLLOWING the env var rather than merely happening to sit somewhere plausible.
     //
-    // THE OTHER TWO RESOLVERS SIT BEHIND THE PROMPT and cannot be reached from a spawned
-    // run — see the interview-wall case below, and #346, for why — so what is pinned here
-    // is the reachable one plus the negative that matters: no run of this shell falls back
+    // THE OTHER TWO RESOLVERS SIT BEHIND THE PROMPT and STILL cannot be reached from a
+    // spawned run: #346 made a piped run refuse the interview rather than answer it, which
+    // moved the wall without removing it — see the no-terminal block below — so what is
+    // pinned here is the reachable one plus the negative that matters: no run falls back
     // to the real accumulus root. `resolvePlansPath` and `resolveEventStorePaths` share
     // `resolveDataDir` with `resolveOrdersPath`, and that agreement is pinned at the
     // resolver (`packages/preferences/src/plans-reliable.test.ts`, the data-dir drift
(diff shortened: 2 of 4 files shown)


apps/tui/src/import-orders-cli.test.ts
```diff
@@ -390,43 +393,84 @@ describe("import-orders-cli — the shell's own contract (argv, exit codes, env,
...
+      expectExited(run);
+      expect(run.status).toBe(1);
+      // THE SHELL'S SENTENCE — a consequence, not a rule, and not a readline internal.
+      expect(run.stderr).toContain(
+        "No terminal on stdin: this import is an interview — it asks which reserve funds " +
+          "the batch — and there is nowhere to conduct it, so every question goes " +
+          "unanswered. Run it from a terminal.",
+      );
+      // ONCE, not once per question. Repeating it would bury the flow's refusal in copies.
+      expect(run.stderr.match(/No terminal on stdin/g)).toHaveLength(1);
```
AmetAlvirde: **MEDIUM — this assertion cannot fail, and the guard it is pointed at cannot be observed. A third decorative guard.**

`toldThereIsNoTerminal` (`import-orders-cli.ts:39`, `:58-59`) can never be read as `true`, because **a run with no terminal calls `ask` exactly once, ever**:

1. the first and only reachable `ask` is `declareFunding`'s batch question (`import-orders.ts:586` → `import-orders-funding-declaration.ts:68`);
2. it returns `""`, so `declareFunding` returns `undefined` at `:70`;
3. `import-orders.ts:587-589` returns `reject(...)` immediately — `declareRungPicks` and the override pass are never reached.

There is no export shape, no argv, no data-dir state that reaches a second question without a terminal. So the mutation *"delete the flag and write the notice unconditionally"* leaves all nine cases green, and `expect(run.stderr.match(/No terminal on stdin/g)).toHaveLength(1)` passes for a reason that has nothing to do with the flag: only one notice was ever *possible*.

The doc comment at `import-orders-cli.ts:52-54` states the justification in the present tense — *"repeating it would bury the flow's refusal under copies of the shell's"* — describing a run this flow structurally cannot produce.

This is falsifiable in exactly the way the board's last two decorative guards were: there is no input that turns it red.

Two honest ways out, either fine:
- **Keep the flag, pin it.** It needs an `ask` called twice with no terminal, which means lifting `ask` out of the top-level `else` so a unit test can drive it — the same extraction `declareFunding` got, and for the same reason.
- **Keep the flag as forward-defence, downgrade both claims.** Say in the comment that today's flow asks once and the flag is insurance against a flow that asks more; and either drop `:427` or relabel it as "the notice is emitted, once" rather than as a pin on the memo.

What must not stand is the current pairing: a comment asserting the flag is load-bearing next to an assertion that would pass without it.

apps/tui/src/import-orders-cli.ts
```diff
@@ -95,6 +137,9 @@ if (!csvPath) {
     process.stderr.write(`${error instanceof Error ? error.message : String(error)}\n`);
     process.exitCode = 1;
   } finally {
-    rl.close();
+    // ONLY IF ONE WAS EVER BUILT. `?.` is the whole point of the lazy construction above:
+    // a run that never prompted must not construct an interface here just to close it,
+    // which would consume stdin on the way out for no reason at all.
+    rl?.close();
```
AmetAlvirde: **MEDIUM — `rl?.close()` is now unreachable from every test in the repo, and the file's coverage claims were not updated to say so.**

`runImport` always passes `input: options.input ?? ""` to `spawnSync` (`import-orders-cli.test.ts:213`), so the child's stdin is a pipe and `process.stdin.isTTY` is falsy in **all nine** cases. `ask` therefore returns at `:57` before `:68` runs, `rl` stays `undefined`, and this `finally` is a no-op on every path the suite can drive.

Concrete mutation: **delete the whole `finally` block** — 9/9 green.

The PR body discloses one honest negative (module-scope `createInterface` survives). This is a second one it does not: before #346 the interface *was* built on every run, so the `finally` really was exercised; after #346 nothing in the suite constructs one, and the close became structural rather than tested. Two places still claim otherwise:

- `import-orders-cli.test.ts:228-232`, `expectExited`'s doc: *"`rl.close()` lives in a `finally`, so a path that lost it would leave stdin open and the process alive until the runner's timeout killed it."* That is no longer the mechanism — losing the close changes nothing observable, because there is nothing to close.
- the `describe` title at `:174`, which still advertises **readline** as one of the four things this file's contract covers.

Not a defect in the shipped code — the `?.` is right, and I verified a real piped run exits cleanly with no fd held open. It is a coverage claim that quietly stopped being true in this diff. Either say so where the PR says the other negative out loud, or reword `expectExited`'s rationale to what it actually still proves (the process ended on its own).

apps/tui/src/import-orders-cli.ts
```diff
@@ -29,7 +29,46 @@ if (!csvPath) {
...
+   * "" hands the domain the one answer a run with no terminal can honestly give, and the
+   * domain then refuses in its own voice. Two sentences, one story: the shell names WHY
+   * there is no answer, the flow names WHAT IT DID about it.
+   *
+   * The notice is written ONCE per run, not once per question: the operator learns there is
+   * no terminal from the first unanswerable question, and repeating it would bury the
+   * flow's refusal under copies of the shell's.
+   */
+  const ask = (question: string): Promise<string> => {
+    if (!process.stdin.isTTY) {
```
AmetAlvirde: **MEDIUM — the readline internal is not gone; it moved to the only input mode this change leaves supported.** Ctrl-D at a real terminal still prints a readline-authored sentence and bypasses the domain refusal entirely.

`process.stdin.isTTY` splits the world in two and this fix covers one half. On the other half — a real terminal, which #346's own remedy is *"Run it from a terminal"* — the natural "I have no answer" gesture is Ctrl-D, and Node 24 rejects the pending `rl.question` through `kQuestionReject`. The outer catch at `:136-138` prints the message verbatim, exactly as it printed `readline was closed` before.

Reproduced against this branch (`d68d8f4`), real pty, authored fixture:

```
$ printf '\004' | script -q /dev/null \
    ./node_modules/.bin/tsx apps/tui/src/import-orders-cli.ts $D/open-orders-export.csv
Funding reserve for this batch: Aborted with Ctrl+D
$ echo $?
1
$ ls $D            # no orders.jsonl
events.jsonl  genesis.json  open-orders-export.csv
```

`Aborted with Ctrl+D` is Node's string, not this repo's. Neither sentence the PR promises is printed: not the shell's *why there is no answer*, not the flow's `REFUSED — no funding reserve was declared for this batch`. `no-reserve-declared` never fires, and the operator is told what **readline** did, not what **the import** did — which is the whole thesis of #346, unmet on the half of the world #346 tells operators to use.

For contrast, the same fixture with Enter instead of Ctrl-D behaves exactly as designed:

```
Funding reserve for this batch:
REFUSED — no funding reserve was declared for this batch
Nothing was written to $D/orders.jsonl.   # exit 1
```

So this is a *gap*, not a regression — pre-#346 Ctrl-D behaved the same. But the test at `:436-438` calls the internal *"the regression the whole change exists to kill"*, and it is killed only for `!isTTY`. Two options:

- **Close the gap here.** `rl.question(question).catch(() => "")` at `:69` routes the Ctrl-D case into the same empty-answer composition the non-TTY case uses, and `no-reserve-declared` then fires in both halves — the composition this PR already argues for, applied to the half it skipped.
- **Or file it and narrow the claim**, so `:436` says the internal is gone *from a run with no terminal* rather than gone outright.

I would not merge on the reading that #346 is fully closed while `pnpm orders:import` at a terminal still answers Ctrl-D in readline's voice.

apps/tui/src/import-orders-cli.ts
```diff
@@ -29,7 +29,46 @@ if (!csvPath) {
...
+        toldThereIsNoTerminal = true;
+        process.stderr.write(
+          "No terminal on stdin: this import is an interview — it asks which reserve funds " +
+            "the batch — and there is nowhere to conduct it, so every question goes " +
+            "unanswered. Run it from a terminal.\n",
+        );
+      }
+      return Promise.resolve("");
+    }
+    rl ??= createInterface({ input: process.stdin, output: process.stdout });
```
AmetAlvirde: **MEDIUM (out of this PR's scope, but the defect is live) — `record-fill-cli.ts` still has the pre-#346 shape, and still leaks `readline was closed` to the operator.**

`apps/tui/src/record-fill-cli.ts:34` is the line this file used to be:

```ts
const rl = createInterface({ input: process.stdin, output: process.stdout });
```

module scope, eager, bound at `:95` as `ask: (question) => rl.question(question)`, closed at `:106`. Same construction, same consumption of a piped stdin, same internal reaching the operator. Reproduced on this branch against an authored data dir:

```
$ NUMISMA_DATA_DIR=$D ./node_modules/.bin/tsx apps/tui/src/record-fill-cli.ts < /dev/null
Resting rungs:
  [0] bitget:TESTUSDT:buy:300:2026-01-02T10:00:00
      TESTUSDT buy 10 @ 300 = 3000 USD against reserve-authored
readline was closed
```

That is the exact string this PR exists to remove, still printed by a sibling shell — and `record-fill-cli.test.ts` does not characterize it, so nothing in the suite knows.

Worth naming here for two reasons. First, this file's test header cites `record-fill-cli.test.ts` as the shape it copied (`import-orders-cli.test.ts:22-23`), so the two shells are explicitly kept in step, and after this merge they are not. Second, `record-fill` is the *thirty-answer* interview — the `ask` count where "once per run, not once per question" would actually have been observable, which is where the `toldThereIsNoTerminal` flag would stop being decorative.

**Not asking for it in this PR** — #346 names the import and the diff is rightly file-disjoint. Asking for a follow-up issue, and for the PR body's *"the internal is gone"* framing to say which shell it is gone from.

apps/tui/src/event-store.ts
```diff
@@ -236,7 +240,26 @@ export async function migrateLegacyLog(
...
+  // is the loop's own rule hoisted above the write — no content once blank lines are
+  // discarded — rather than an `=== ""` check that would fix half of it.
+  //
+  // THE GUARD BELONGS HERE, NOT IN `readOptional`. That helper maps ENOENT alone to
+  // `undefined`, and its other callers read it for exactly that: "absent" and "present
+  // but empty" are different facts about the disk. Widening it to fold `""` into
+  // `undefined` would quiet this path by destroying the distinction that separates an
+  // empty log from an unreadable one. "Nothing to migrate" is a policy this migration
+  // owns.
+  if (raw === undefined || raw.trim().length === 0) {
```
AmetAlvirde: **LOW — the guard changes one more thing than "no write", and it is neither in the PR body nor under a test: a contentless log now skips `loadGenesis` too.**

The early return jumps over `const genesis = await loadGenesis(paths.genesis)` on the next line. So a data dir with an empty (or blank-lines-only) `events.jsonl` and a **missing or corrupt `genesis.json`** changes verdict:

- before — guard not taken → `loadGenesis` throws → shell `.catch` → **exit 1**, stderr naming `genesis.json`;
- after — **exit 0**, stdout `No durable log to migrate at …/events.jsonl.`

Reproduced on this branch:

```
$ D=$(mktemp -d); : > $D/events.jsonl        # no genesis.json at all
$ NUMISMA_DATA_DIR=$D pnpm exec tsx apps/tui/src/migrate-legacy-log.ts
No durable log to migrate at /tmp/…/events.jsonl.   # exit 0
```

Nothing pins it in either direction: `migrate-legacy-log.test.ts:506` ("exits 1 when the data dir has no genesis to migrate against") plants a log **with content**, so it stays green and never noticed the split.

I think this is the *right* answer — nothing to migrate means nothing to validate against, and it makes the empty log behave exactly like the absent one, which is the guard's whole thesis. Flagging it because it is a second behaviour change riding on a one-line predicate, and the PR body's `touches no disk at all` describes only the write half. Cheapest resolution: one sentence in the guard comment (`… and does not read genesis either: there is nothing to validate against`) plus a sibling of `:506` with an empty log asserting exit 0. Otherwise the next person to widen this guard has no signal that the genesis read moved behind it.

apps/tui/src/import-orders-cli.ts
```diff
@@ -29,7 +29,46 @@ if (!csvPath) {
...
+    if (!process.stdin.isTTY) {
+      if (!toldThereIsNoTerminal) {
+        toldThereIsNoTerminal = true;
+        process.stderr.write(
+          "No terminal on stdin: this import is an interview — it asks which reserve funds " +
+            "the batch — and there is nowhere to conduct it, so every question goes " +
+            "unanswered. Run it from a terminal.\n",
+        );
+      }
+      return Promise.resolve("");
```
AmetAlvirde: **LOW (latent, not a present defect) — `""` is "the one honest answer" at exactly one call site; at every other prompt in this flow it means "yes, take the default".**

The composition works, and I verified it does. But it is load-bearing on an *ordering* fact that nothing states or asserts: that the first question a run can reach is the one where blank means refuse.

Everywhere else in the same interview, blank means proceed:

- `import-orders-funding-declaration.ts:74` — `isAffirmative("")` is `false`, i.e. a deliberate "no, no per-order overrides";
- `import-orders-rung-picks.ts:161` — `answer === "" || isAffirmative(answer)` → **accept every proposed rung**, and that module's own comment at `:156-159` says so explicitly: *"A BLANK LINE IS ACCEPTANCE … It is NOT the refusal a blank funding answer is"*;
- `import-orders-rung-picks.ts:183` — blank takes the proposed default and breaks.

So if the funding declaration ever stops being the first question — a pre-flight prompt added ahead of it, `declareFunding` reordered, or a future caller of `ask` — a run with no terminal would stop refusing and start silently ratifying every default, then **write** durable lines nobody declared. That is the opposite of what this notice promises the operator ("every question goes unanswered"), and no test would catch the day it flips: the nine cases here all key off exit 1 and an absent `orders.jsonl`, both of which come from `declareFunding` alone.

Nothing to fix in the shipped behaviour. Two cheap options:
- narrow the claim at `:47-51` — `""` is the honest answer *to the funding declaration*, and the refusal composition depends on it being asked first; or
- make the ordering explicit rather than incidental, e.g. have `importBitgetOpenOrders` reach `declareFunding` before any other `ask` by construction, with a comment at `import-orders.ts:586` saying the non-TTY refusal rests on it.

Recording it because the two-sentence story in this doc comment is stated as a property of `ask`, and it is really a property of who calls `ask` first.

AmetAlvirde (commented):
## Review — PR #363 (`pr-review` / `review` discipline, Opus 5 high)

**Verdict: the two shipped behaviour changes are correct.** I could not construct any input where the #345 guard writes when it should not, or skips a log that has content; and I could not construct any input where the non-TTY `ask` hangs, leaks, or lets a line reach disk. The two inverted `KNOWN DEFECT` pins assert the right thing. Nothing here blocks merge on correctness of the shipped code.

**What I did find is three claims the change makes that its tests cannot hold**, plus one residual instance of the very defect class #346 names, still live in the repo. Six findings inline, one below (not line-anchorable).

Findings by severity: 3 medium, 3 low. No high, no critical — stated plainly rather than padded.

---

### L1 (low, not line-anchorable — the line is outside the diff)

`apps/tui/src/import-orders-cli.test.ts:298`

```ts
// The operator was never taken into the interview: every readable rung was already on
// file, so there was nothing to attribute. The funding prompt's own text is the proof.
expect(run.stdout).not.toMatch(/Funding reserve for this batch/);
```

**This PR silently destroyed that assertion's evidentiary value, and the comment above it is now false.** Before #346, the prompt text reached stdout because `rl.question(query)` wrote the query to `output`. After #346, a non-TTY `ask` returns before `rl` is ever built, so it writes *no query at all* — the notice goes to **stderr** and the question text is never emitted anywhere. Verified: in a real piped run the entire stdout is empty while the interview *is* entered.

So `run.stdout` cannot contain `Funding reserve for this batch` in **any** spawned run, whether or not the interview was entered. The assertion can no longer distinguish "never asked" from "asked and refused", which is precisely the distinction its comment claims it proves.

Concrete fix: the post-#346 probe for "the interview was never entered" is the one the new sibling case already uses —

```ts
expect(run.stderr).not.toMatch(/No terminal on stdin/);
```

— which *can* fail, because the `imported-partial` fixture is one `FRESH_ROW` away from entering the interview.

---

### What I checked and found clean

Recorded so the next reader knows what has already been ruled out, and does not pay for it twice.

- **The `trim()` predicate does not diverge from the line loop's skip rule — for any input.** `raw.trim().length === 0` ⟺ every character of `raw` is whitespace ⟺ every element of `raw.split("\n")` trims to `""` ⟺ the loop yields no record. `\n` is itself whitespace, so the split cannot separate the two. Checked the exotic members of the ECMAScript `WhiteSpace`/`LineTerminator` set the loop also trims — NBSP `\u00a0`, BOM `\uFEFF`, `\u2028`/`\u2029`, `\v`, `\f`, `\u3000`: guard and loop agree on all of them. A lone `\u0000` is not whitespace to either, and still aborts at `line 1 is not valid JSON` as before. **No divergence exists.**
- **The guard changes nothing for a log that is not contentless.** `touched === 0` ⟺ contentless, both before and after: the only non-blank `continue` in the loop is the `unresolved.push(...)` arm, which throws at the end, so any log with content either yields `migrated.length >= 1` or aborts. The shell's `touched === 0` branch and the guard's predicate cover exactly the same set of logs. (One consequence *outside* the write is real — see the `event-store.ts:262` comment.)
- **No infinite loop from the empty answer.** `declareRungPicks` has a genuine `for (;;)` re-ask loop at `import-orders-rung-picks.ts:181`, which under a permanently-`""` `ask` would spin forever — but `""` hits `break` on the first iteration (`:183`), and the loop is unreachable anyway because `:161` accepts the batch on `""`. Verified by reading, not assumed.
- **No resource leak, no hang, no stray fd from the lazy interface.** On the non-TTY path `rl` is never constructed, stdin is never read, nothing holds the loop open, and the process exits on its own. Verified end to end.
- **The interactive path still works across several questions on one memoized interface.** Driven through a real pty: `Funding reserve for this batch:` → `reserve-authored`, `Override…? [y/N]` → Enter, exit **0**, one `orderPlaced` line in `orders.jsonl`, no litter. The `rl ??=` memo is genuinely exercised there.
- **Everything between `ask` returning `""` and `declareFunding` returning `undefined`.** `(await ask(...)).trim() === ""` → `undefined` (`import-orders-funding-declaration.ts:69-71`) → `return reject(io, "no-reserve-declared", …)` (`import-orders.ts:587-589`). No write precedes it: `records` is still empty, observations are built but not appended, and `reject` returns before `appendOrders`. Confirmed on disk — no `orders.jsonl`, no `.tmp`, no `.lock`.
- **A blank line at a real terminal still reaches the domain refusal.** Driven through a pty: Enter at the funding prompt → `REFUSED — no funding reserve was declared for this batch`, exit 1, nothing written. The TTY and non-TTY routes to `no-reserve-declared` agree. (Ctrl-D does not — see the `import-orders-cli.ts:57` comment.)
- **Both inverted pins are correct and both can fail**, on the mutations the PR body lists. I did not re-run them.
- **The suite is green here:** `import-orders-cli.test.ts` 9/9, `migrate-legacy-log.test.ts` 14/14. `ops/testkit/gitignored-path-globs.test.ts` is another lane's and out of scope; not looked at.

Reviewed read-only in `.claude/worktrees/lane-c` at `d68d8f4`; no source or test was edited, and every claim above was reached by reading or by running the real shells against throwaway fixtures.

apps/tui/src/import-orders-cli.test.ts
AmetAlvirde: **Accepted and fixed** in 8f70adc.

Confirmed the mutation independently before acting on it: deleting the `toldThereIsNoTerminal` flag and writing the notice unconditionally leaves the file **9/9 green**. The assertion could not fail.

The assertion is deleted. The flag is **kept** — it is correct and cheap, and the thirty-answer rung walk in the twin shell is exactly where once-per-run would become observable. What replaces the assertion is a disclosure in this file's existing "what this does not pin" voice, stating that a no-terminal run puts exactly one question (`import-orders.ts:586` → `declareFunding` → `""` → `undefined` → `:587` returns `reject`), so the once-per-run behaviour is untestable from a spawn rather than merely untested.

The `ask` docblock in `import-orders-cli.ts` was overstating the same thing and has been softened to match.

apps/tui/src/import-orders-cli.ts
AmetAlvirde: **Accepted and fixed** in 8f70adc.

Verified independently: deleting the entire `finally { rl?.close() }` leaves the file **9/9 green**. `runImport` always passes `input: … ?? ""`, so `isTTY` is falsy in all nine cases and `rl` never gets built.

The **code stays** — it is correct, and you confirmed it on a real pty. What was wrong was the advertising, so that is what changed:

- the describe title `…, env, readline)` → `…, env, no-terminal stdin)`, with a note that no case here constructs a readline;
- the `expectExited` doc rewritten as a termination assertion, disclosing that both halves of the readline lifecycle are TTY-only;
- the relative-`NUMISMA_DATA_DIR` case's inline comment, which claimed the readline "was constructed" and that the case exercises the `finally` — both false — corrected.

apps/tui/src/import-orders-cli.ts
AmetAlvirde: **Accepted as a documentation defect and fixed; the behaviour change is deliberately deferred.**

You are right that the claim was false as written. `Aborted with Ctrl+D` at a real terminal is Node's own string reaching the operator, which is the defect class #346 names. In 8f70adc the claim is scoped to the piped/non-TTY path and the TTY Ctrl-D gap is recorded explicitly, including that a blank line at a TTY refuses correctly — so it is a gap, not a regression.

**Declining `rl.question(q).catch(() => "")` in this PR**, for a specific reason rather than scope reflex: nothing in this suite can stand at a pty, so that fix would ship **unpinnable** — a fourth assertion that cannot fail, in a PR whose entire review outcome was two assertions that could not fail. Reproducing it needs the `script -q /dev/null` harness you used, which is a real increment and cross-cutting (the twin shell has the same hole).

A follow-up covering TTY Ctrl-D together with a pty harness is going to the board owner for filing. Resolving here because the false claim — the part that belongs to this PR — is fixed.

apps/tui/src/import-orders-cli.ts
AmetAlvirde: **Confirmed live, and deliberately not fixed here.**

`record-fill-cli.ts` does still hold the module-scope eager `createInterface` and does still print `readline was closed` verbatim on a piped run. Agreed on all counts, including your sharpest point: it is the thirty-answer interview, so it is the shell where once-per-run is actually observable — which makes it the better home for that pattern than this one.

Not fixing it in #363: it is file-disjoint from #345/#346, and this lane's merge is sequenced against other lanes on the board. Folding an unrelated shell into it would widen the diff past what the two issues close and put `closingIssuesReferences` at risk.

Going to the board owner as a follow-up, paired with the TTY Ctrl-D gap above since both want the same pty harness. Resolving as acknowledged-and-routed, not as dismissed.

apps/tui/src/event-store.ts
AmetAlvirde: **Accepted and fixed** in 8f70adc — with one correction to the framing.

The behaviour change is real and was genuinely undisclosed, so the finding stands. But it is not incidental: an **absent** log has always returned from above the genesis read, so a contentless one leaving through the same door is what makes the "one seam" claim in #345 true rather than a coincidence. Exit 0 on empty-log-plus-missing-genesis is the correct answer, not a side effect to be tolerated.

What was actually missing is that nothing pinned it — `migrate-legacy-log.test.ts:506` plants a log **with content**, as you noted. So:

- a new case, `needs no genesis at all to report an empty log as nothing to migrate (exit 0)`, on a bare temp dir with no genesis: exit 0, the exact sentence, `stderr` empty and free of `/genesis/`, log byte-for-byte `""`, no genesis conjured, no litter;
- the guard's comment now discloses that it returns above `loadGenesis` and why that is consistent.

Mutation-verified twice, independently: moving the guard below `loadGenesis` gives **1 failed | 14 passed**, and the one red is this new case — every other case in the file stays green, which is the point.

apps/tui/src/import-orders-cli.ts
AmetAlvirde: **Accepted and documented** in 8f70adc.

Checked all four readings before writing it down, and they are as you describe: `import-orders-funding-declaration.ts:68` refuses on an empty batch answer, but `:74` reads `""` as "no overrides", `import-orders-rung-picks.ts:161` as "accept every proposed rung", and `:183` as "take the default".

So the empty answer is honest at exactly one call site, and the refusal rests on an unstated ordering fact. The `return Promise.resolve("")` site now names all four readings, states that only funding-asked-first makes the empty answer safe, and spells out the consequence of a reorder: a no-terminal run would silently ratify defaults and **write**.

No behaviour change — this is latent, and the guard against it is that the next person to reorder the interview reads why they must not.

Issue #357, opened 2026-08-17
Price feed: the lost-day signal is computed and warned, but reaches no one outside the TUI


AmetAlvirde:
Split out of #356, which asked for two things: (a) a missed window surfaced as failure rather than success, and (b) a supported backfill path. **(b) is being specced now** as the `--as-of` recovery increment. This issue carries what remains of (a), restated to match what the code actually shows.

## This is NOT a request to reverse D8

[[2026-08-12-calendar-aware-instrument-coverage-grill]] (#266) locked **D8 — warn only, no suppression, no exit-code change**, with `exitCode` declared as the literal type `0` so "exit 0 when days are lost" is testable, and the command's own contract stated as *"A lost day is reported; a broken job is failed."* D7 promotes that from preferable to mandatory: having accepted ~10 holiday false positives a year, a non-zero exit would fail the daily automation ten times a year.

#356's wording ("reports success at every surface") reads as an ask to escalate to failure. **That reading is rejected here.** If D8 is to be re-opened it deserves its own grill against #266, not a side door.

## The actual gap: delivery, not detection

Detection already works. `venueDark` is a live second array on `GapReport`:

- `packages/event-store/src/gap-report.ts:186,211,322,345` — `VenueDarkDay`, the array, and the walk that fills it
- `packages/event-store/src/gap-report.ts:410` — `formatVenueDarkDays`
- `packages/event-store/src/gap-report-io.ts:122` — serialized as its own key, deliberately never folded into `lost`

So on a day a whole venue owed marks and produced none, the signal **is computed and it is rendered as a warning**. The defect is that it reaches a human only if that human opens the TUI.

## Two mechanisms keep it from arriving

1. **The file half is written by the job whose absence it reports.** `ops/price-feed/run-daily-fetch.sh` step 5 runs `pnpm gap-report -- --write`. On a fully-missed evening the wrapper never fires, so `gap-report.json` is never refreshed and keeps describing the last good day. The report that would name the loss is authored by the run that didn't happen. (The TUI's live-derived channel is exempt — the wrapper comment says so explicitly — but that is a pull surface, not a push.)

2. **Nothing pushes.** `run-daily-fetch.sh` already documents this for the exit code: *"A FAILED JOB REACHES NO ONE EITHER … launchd RECORDS the exit code and `launchctl print` will show it on request, but nothing is PUSHED — no notification, no badge, no mail."* The same is true of the venue-dark warning.

## Evidence this is real, not theoretical

[...]

PR #364, opened 2026-08-18
fix(engine,preferences): refuse a blank data dir instead of silently using the real ledger

### LINKED ISSUE #348: an empty-string data dir silently resolves to the real private ledger, while a relative one is loudly refused
Found while running the open-issue board: lane C hit the test-helper form of this, and lane E identified that the same hole is **structural in production code**, not lane-local. Filing rather than fixing — no lane's scope covered it.

## The defect

Two resolvers treat an **empty string** as "not configured" and fall through to the real private ledger, while treating a *relative* path as a loud error.

`resolveSidecarPath` (`packages/preferences/src/sidecar-io.ts:36-49`):

```ts
if (dataDir === undefined || dataDir.trim() === "") {
  return join(resolveDataDir(), fileName);   // <- silent fallthrough
}
const raw = dataDir.trim();
if (!isAbsolute(raw)) {
  throw new Error(`a sidecar data directory must be an absolute path ...`);
}
```

`resolveDataDir` (`data-dir.ts`) repeats the shape for the env var — `if (fromEnv && fromEnv.trim() !== "")` — and its `else` returns `accumulusDataDirDefault()`.

So `resolveSidecarPath("plans.jsonl", "")` and `NUMISMA_DATA_DIR=""` both resolve to the **real accumulus data directory**.

## Why this is worth a ticket

The asymmetry is the problem, not the fallback itself. A defaulting resolver is fine. What is not fine is that **the same function is careful about one malformed input and silent about another** — a relative path gets a paragraph-long refusal explaining split-brain ledgers, and an empty string gets the live ledger with no signal at all.

Empty string is not an exotic input. It is what a `?? ""` default produces, what an unset shell variable expands to under `${VAR:-}`, and what a config reader returns for a present-but-blank key. Each of those reads naturally as *"no directory configured"* — and each silently selects the operator's real private data.

The failure mode is the worst available: it **fails green**. A test with this bug passes, because the real ledger is a valid readable directory. Nothing throws, nothing warns, and the only evidence is a write that landed somewhere nobody looked.

## Observed, not theoretical

- Lane C's cancel-order test runner defaulted `NUMISMA_DATA_DIR` to `""` and resolved to the live ledger. **No leak occurred** — caught during review and made a compile error in that branch — but it was green while wrong.
- Lane E's build introduces a **new** durable artifact. The same mistake there would have *created a file* in the live ledger rather than merely reading one; a new file that nothing would flag as foreign.

## Suggested direction

Treat empty as malformed rather than absent, in both resolvers: `undefined` means "not configured, use the default"; `""` (or whitespace) means "someone tried to configure this and got it wrong" and should be refused in the same voice as the relative-path case. The existing message is the right model — it explains the consequence rather than just naming the rule.

Whether the default itself should require an explicit opt-in is a larger question and is **not** proposed here.

AmetAlvirde:
Every data-dir resolver treated an empty string as "not configured" and fell through to the real accumulus ledger, while treating a *relative* path as a loud error. The asymmetry was the defect, not the fallback: a deployment that meant to point at a scratch directory and got the expansion wrong wrote to the operator's real private ledger, silently.

Closes #348.

## The new rule, everywhere

- **`undefined` = not configured** → use the default. Unchanged, and every existing test depends on it.
- **`""` or whitespace-only = someone tried to configure this and got it wrong** → refuse, loudly, in the same voice as the relative-path case (explain the *consequence*, don't just name the rule).

Applied in four TypeScript resolvers and one shell wrapper. The two-line change was never the work; the sweep was.

## ⚠️ This reverses a documented `MUST`

`sidecar-io.ts` carried this, and it is no longer the rule:

> An explicit `""` MUST land here and not on `resolve("")`, which is the process's CWD: a durable, git-tracked artifact written relative to whatever directory a script happened to start in is a split-brain ledger waiting to happen, and `""` is exactly the value an unset shell variable expands to at the one call site most likely to produce it.

**That reasoning is not wrong — it is incomplete.** It weighed exactly two arms and correctly killed the wrong one. It never considered a third.

The docstring is **rewritten to record all three arms, not deleted** — the CWD argument is load-bearing and has to survive so nobody reintroduces `resolve("")`:

1. **`""` → `resolve("")` → the process's CWD.** Rejected, and the rejection still stands. Never reintroduce it. The throw is now what keeps this arm shut, since `""` no longer has a landing spot.
2. **`""` → the default, silently.** What the code did until now. Its argument — that `""` is what an unset shell variable expands to at the call site most likely to produce it — is true, and incomplete: it treats a *misconfigured* knob as an *absent* one. Arm 2 cannot see the failure where a wrong expansion aims writes at the real ledger. That is why it lost, not because its CWD reasoning was wrong.
3. **`""` → refuse.** Wins. The relative-path arm already refuses a present-but-unusable value for exactly this reason, and empty is the same class of mistake with a worse blast radius.

`context/adr/ADR-006` and its `INDEX.md` row both stated the contract as "empty/whitespace fall-through to the default" and are corrected, so an accepted ADR does not go false.

## ⚠️ The first sweep used an inverted criterion, and that is why it missed things

The original sweep looked for `${VAR:-}` as the hazard idiom and treated a *non-empty* default as safe. **That is backwards.**

- `${VAR:-anything}` — the **colon** form — substitutes when the variable is unset **or empty**. This is the hazard: it silently swallows a misconfigured blank.
- `${VAR-anything}` — **no colon** — substitutes only when the variable is **unset**. This is the form that matches the new contract.

The wrong criterion is precisely why `ops/price-feed/run-daily-fetch.sh` was cleared on the first pass. The corrected rule was applied to a re-sweep, and it found the site immediately. Recording it here because the next person to sweep this repo will reach for the criterion in this body.

## The complete sweep, re-run

**No production TypeScript call site can produce `""`.** Every `resolveOrdersPath` / `resolvePlansPath` / `resolveReconciliationsPath` / `resolvePreferencesPath` / `resolveEventStorePaths` call in app code is a bare no-arg call. There is no `--data-dir` flag, no JSON config carrying a `dataDir`, no persisted sidecar field. The env var is the sole external knob.

But the first sweep was scoped to *callers*, and the defect was in *resolvers*:

| resolver | before | now |
| --- | --- | --- |
| `resolveDataDir` (`packages/engine/src/data-dir.ts`) | `""` → default | refuses |
| `resolveSidecarPath` (`packages/preferences/src/sidecar-io.ts`) | `""` → default | refuses |
| `resolvePreferencesPath` (`packages/preferences/src/preferences.ts`) | `""` → **`<cwd>/preferences.jsonl`** | refuses |
| `resolveEventStorePaths` (`packages/event-store/src/event-store.ts`) | `""` → **`<cwd>/events.jsonl`** | refuses |

The bottom two took a **default parameter** — `dataDir = resolveDataDirDefault()` — and a JS default parameter fires on `undefined` and on nothing else. An explicit `""` sailed straight past it into `resolve("")`, the process CWD. That is **arm 1**, live, in the resolver that owns `genesis.json` and the append-only `events.jsonl`. Both are now explicit `undefined` checks rather than default parameters, which is the only shape that can tell the two cases apart.

Verified at runtime before the fix, not inferred:

```
resolvePreferencesPath('')       => <cwd>/preferences.jsonl
resolveEventStorePaths('').log   => <cwd>/events.jsonl
```

**A fifth site laundered a blank past all of them.** `apps/price-feed/src/paths.ts` called `resolve(dataDir)` *before* delegating:

```ts
export function resolvePriceFeedPaths(dataDir: string): PriceFeedPaths {
  const base = resolve(dataDir);                                  // `""` becomes an absolute CWD path HERE
  const { genesis, log, inbox } = resolveEventStorePaths(base);   // arrives already disguised
```

`resolve("")` is the process CWD — absolute, and perfectly valid-looking. So a blank reached the hardened `resolveEventStorePaths` wearing a shape its guard could not recognise, the guard never fired, and the pair produced exactly the `<cwd>/events.jsonl` that resolver's own error message says must never happen. The guard now sits **ahead of** `resolve()`, which is the only placement that works.

**This closes a hole production cannot currently reach, and that is stated deliberately rather than left for a reader to discover.** Both callers — `cli.ts:14` and `fetch-prices.ts:167` — pass `config.dataDir`, and `DEFAULT_CONFIG.dataDir = resolveDataDir()` (`config.ts:83`) is already absolute and already blank-refused. `dataDir` is a required parameter here with no default, so there is no `undefined` → default arm either. The warrant is not a live bug: it is that **the PR would otherwise ship an error message a sibling function falsifies**, and a contract contradicted by live code in the same repo is the precise defect #348 is about. The fix makes the resolver's own contract hold for any caller, including ones that do not exist yet.

**On whitespace, and the over-correction to avoid.** `resolve("   ")` does *not* collapse to the CWD — it yields `<cwd>/   /events.jsonl`, a real directory named three spaces. That is a distinct hazard, not a cosmetic variant of the empty case. But the predicate is `trim() === ""`, so a path merely *containing* a space (`/tmp/numisma authored root 348`) still passes, and that is pinned by its own test. A guard that rejected spaces outright would be a new bug wearing this fix's clothes.

**`launcher.testkit.ts`** — `options.env.NUMISMA_DATA_DIR ?? ""`, the exact idiom this issue makes illegal, is replaced with a narrowing of a value `assertIsolated` already proves real. It was already dead code; manufacturing the empty string the resolvers now refuse is not something to leave lying around.

## The shell plane was running the outlawed arm

`ops/price-feed/run-daily-fetch.sh` read the knob as `${NUMISMA_DATA_DIR:-$HOME/Dev/accumulus/data}`. Both failure directions were live:

| `NUMISMA_DATA_DIR` | shell, before | TypeScript, after |
| --- | --- | --- |
| unset | real ledger | real ledger — agree |
| `""` | **real ledger** (arm 2) | **refuses** |
| `"   "` | **literal `"   "`, resolved against CWD** (arm 1) | **refuses** |

The whitespace row is the sharper one: `:-` does **not** substitute for a non-empty blank, so `"   "` survived verbatim and both `HEARTBEAT_FILE` and the `git -C "$DATA_DIR"` commit step resolved it against whatever directory the scheduler started in. **That is #348's own asymmetry surviving in the other language** — the same input loudly refused by the TypeScript resolvers and silently accepted as a relative path by the wrapper that drives them.

Now: `${VAR-default}` (unset only), plus an explicit refusal for either blank spelling, exiting **78** (`EX_CONFIG`, chosen against the codes this wrapper already speaks — 1, 124, 127, 143 — so the log distinguishes "broken knob" from "the run tried and failed").

**Where that refusal surfaces.** The inherited stderr, which under launchd is the plist's `StandardErrorPath` — `~/Library/Logs/numisma/launchd.price-feed.err.log`. Not the per-run `price-feed-<stamp>.log`, because that file does not exist yet: the check must run before `HEARTBEAT_FILE` is derived, which is itself above the `mkdir -p "$LOG_DIR"` / `LOG_FILE=` / `exec … tee` block. `LOG_DIR` comes from its own env var and does not depend on `DATA_DIR`, so a blank data dir does not blind the logging — the per-run log is simply not open yet.

**There is deliberately no heartbeat for this failure.** The heartbeat is written *into* `$DATA_DIR`, and the whole finding is that we do not have a trustworthy one. Inventing a fallback location so something gets written is the exact defect #348 kills: guessing a path for a knob the operator got wrong. A future edit that relocates this check to get a breadcrumb out of it would reintroduce the bug.

## ⚠️ Consequence: this converts a silent-corruption failure into a silent-absence one

Worth stating plainly rather than discovering later.

Before: a blank or whitespace `NUMISMA_DATA_DIR` meant the scheduled job **ran** and wrote ledger data to a directory nobody configured. After: it refuses and does nothing.

**A refusing run writes no heartbeat, so the previous run's heartbeat stays in place — and it says the job was clean.** #357's delivery channel reads `exitCode` and `lastStep` from the last heartbeat at step 0, so a refused run reports healthy. That is the *silently-absent job* case, which #357 explicitly scoped **out** (step 0 cannot see it, and the alternative was a second LaunchAgent that delivery grill D2 rejected on cost). This change creates an instance of a gap that is known, deliberate, and currently unowned.

**It is still the right trade: silent absence beats silent corruption, and absence is recoverable.** And it is bounded — see below.

## The 18:00 unattended job cannot trip the refusal

A refusal firing in the daily job would be worse than the silent default it replaces, so this is a proof, not an assumption.

**The proof is the rule itself: the refusal can only fire on a *misconfigured* variable. An unset one still takes the default.** Nothing about the scheduled path is changed by this PR unless someone has actively set `NUMISMA_DATA_DIR` to a blank value — which is precisely the misconfiguration #348 asks us to refuse. This holds regardless of machine state, which is why it is the argument that matters.

Corroborated on this machine: the **installed** plist (`~/Library/LaunchAgents/com.numisma.pricefeed.daily.plist`, not just the `ops/` template) carries the absolute literal `/Users/amet/Dev/accumulus/data`; `~/.config/numisma/price-feed.env` contains zero `NUMISMA_` lines; there is no `export NUMISMA_DATA_DIR=` anywhere in the repo; and `DATA_DIR` is read *before* the env file is sourced, so a blank injected there could not reach it anyway.

## Every new test was proven to fail

No guard here is decorative. Each mutation was applied, the red observed, and the tree restored.

| mutation | result |
| --- | --- |
| engine guard reverted to the pre-fix `if (fromEnv && fromEnv.trim() !== "")` | **3 red** — both refusal tests + the message test; the `undefined` tests stayed green |
| engine default arm broken (`env.NUMISMA_DATA_DIR ?? ""`) | **4 red** — all default tests; the refusal tests stayed green |
| sidecar blank-refusal reverted | **2 red** — refusal tests; defaults green |
| sidecar default arm → `join(process.cwd(), …)` | **2 red** — defaults; refusal green |
| sidecar made to *return* `join(process.cwd(), fileName)` instead of throwing | **1 red**, and the failure message names the CWD path — the assertion that used to be tautological |
| `resolveEventStorePaths` blank refusal removed | **3 red** of 13; the 10 remaining, including `undefined` → default, stayed green |
| `resolvePreferencesPath` blank refusal removed | **3 red** of 63; the 60 remaining, including `undefined` → default, stayed green |
| `resolvePriceFeedPaths` guard removed entirely | **3 red** of 5 |
| `resolvePriceFeedPaths` guard **moved below `resolve()`** and tested against `base.trim()` | **the same 3 red** — see below |
| shell wrapper reverted to the colon form with the refusal dropped | red on the appended wrapper cases |

**The placement mutation is the strongest evidence here.** Moving the `paths.ts` guard *below* `resolve()` and testing `base.trim()` instead is the **plausible wrong fix** — it reads as correct in review and does exactly nothing, because `resolve("")` is already a non-blank absolute path. It reds the same three tests as deleting the guard outright. **These tests catch placement, not merely presence**, which is the property that separates a real guard from a decorative one.

The two arms are independently pinned in every resolver: breaking either one reds only its own tests, and the `undefined` → default discrimination stays green throughout. That property is what makes these guards load-bearing rather than decorative.

## Deliberately out of scope

`resolvePreferencesPath` and `resolveEventStorePaths` also silently accept a **relative** path that `resolveDataDir` and `resolveSidecarPath` refuse loudly. Same family as #348, different rule — filed as **#369** rather than folded in here. That issue also notes the four resolvers disagree about `~` today, which needs a deliberate ruling before anyone writes the shared validation table.

## Verification

`pnpm typecheck` — clean, all seven projects.

Full suite on the merged head, with `origin/main` (`7269a9f`) merged in:

```
Test Files  157 passed | 4 skipped (161)
Tests       2435 passed | 21 skipped (2456)
```

**Zero failures.**

An earlier run on the pre-merge head showed a single red — `run-daily-fetch.test.ts > case 4 — an EXTERNAL stop before the ceiling`, a 2000ms budget on a fake missed on run 5 of 12 under full parallel load. It did not reproduce in two isolated re-runs of that file (55 passed, twice), and CI passed the same commit. It is a known load-sensitive timing flake in tests that shell out to a real wrapper; the sighting and its instrumentation are recorded in **#372** rather than re-argued here. This lane's change was ruled out mechanically before that call was made: the edit alters behaviour only when `NUMISMA_DATA_DIR` is blank, case 4 runs against a real temp data dir, and a broken guard would kill the wrapper before *any* step rather than on one run of twelve.


packages/engine/src/data-dir.ts
```diff
@@ -29,7 +29,15 @@ function accumulusDataDirDefault(): string {
  * plane can drift onto a divergent ghost ledger.
  *
  * Resolution rule (must stay byte-identical across planes):
```
AmetAlvirde: **HIGH — missed caller site: the shell plane still resolves a blank to the real ledger, and now disagrees with this resolver.**

This line requires the rule to "stay byte-identical across planes." `ops/price-feed/run-daily-fetch.sh:59` is a plane, and after this PR it no longer matches:

```sh
DATA_DIR="${NUMISMA_DATA_DIR:-$HOME/Dev/accumulus/data}"
```

I verified all three inputs against both planes:

| `NUMISMA_DATA_DIR` | shell `DATA_DIR` | `resolveDataDir` |
|---|---|---|
| unset | `/Users/amet/Dev/accumulus/data` | `/Users/amet/Dev/accumulus/data` — agree |
| `""` | **`/Users/amet/Dev/accumulus/data`** | **THROWS** — diverge |
| `"   "` | **`"   "`** (CWD-relative!) | **THROWS** — diverge |

Two distinct defects:

1. **`""` → the real ledger.** `${VAR:-default}` substitutes when `VAR` is unset *or empty* — so the wrapper is still running **arm 2**, the arm this PR argues is wrong. Under a blank, `HEARTBEAT_FILE="$DATA_DIR/job-heartbeat.json"` (`:90`) writes into the real accumulus tree and step 3 runs `git -C "$DATA_DIR" add -u` / `commit` (`:672`–`:681`) against it, while node refuses. The blast radius #348 exists to close is still open on the shell side; the planes merely fail in *different* directions now.
2. **`"   "` → a relative path.** `:-` does *not* substitute for whitespace-only, so `DATA_DIR` becomes the literal `"   "` and every `git -C "   "` / heartbeat write resolves against the process CWD — precisely the split-brain hazard arm 1 of the `sidecar-io.ts` docstring forbids.

**Why the sweep missed it.** The PR body's criterion was that the hazard idiom is `${VAR:-}` and that "every `NUMISMA_*` expansion in the wrapper uses the non-empty-default form." That criterion is inverted: a *non-empty* default is exactly what silently swallows an empty value. `${VAR:-anything}` is the hazard; `${VAR-anything}` (no colon) is the unset-only form that would match the new TS contract.

Not a blocker for the 18:00 job today — I confirmed the installed plist carries the absolute literal `/Users/amet/Dev/accumulus/data` and `~/.config/numisma/price-feed.env` has zero `NUMISMA_` lines, so the refusal cannot fire there. But that safety is machine state, not a code invariant, and nothing guards it.

packages/preferences/src/plans-reliable.test.ts
```diff
@@ -90,13 +90,51 @@ describe("resolvePlansPath — ADR-006's invariant, at every door", () => {
...
+    // and not just as `toThrow`, because the failure mode this guards is a RETURN of
+    // `resolve("")/plans.jsonl` — which a throw-shaped assertion alone would not name.
+    let produced: string | undefined;
+    try {
+      produced = resolvePlansPath("");
+    } catch {
+      produced = undefined;
+    }
+    expect(produced).toBeUndefined();
+    expect(produced).not.toBe(join(process.cwd(), "plans.jsonl"));
```
AmetAlvirde: **LOW — this assertion is tautological; it cannot fail.**

`expect(produced).toBeUndefined()` on the line above has already passed, so `produced` is `undefined` here — and `undefined` can never `.toBe()` a string. There is no source mutation that turns this line red while leaving line 116 green.

The comment above claims it guards "a RETURN of `resolve("")/plans.jsonl` — which a throw-shaped assertion alone would not name." But in that scenario the `catch` never runs, `produced` holds the path, and **line 116 fails first**. Line 116 is doing the whole job; this line is decorative.

If the intent is to name the CWD failure mode in the output, invert the order so the specific assertion is the one that reports:

```ts
expect(produced).not.toBe(join(process.cwd(), "plans.jsonl"));
expect(produced).toBeUndefined();
```

Every other test added in this PR is load-bearing — I checked each against the mutation that reds it.

packages/engine/src/data-dir.test.ts
```diff
@@ -17,12 +19,41 @@ describe("resolveDataDir — the single durable-ledger data-root resolver", () =
...
+    }
+    expect(message).toMatch(/not .?unset.?/i);
+    expect(message).toMatch(/REAL default ledger/);
+    expect(message).toMatch(/Unset NUMISMA_DATA_DIR/);
+    expect(message).toMatch(/absolute path/);
+  });
+
+  it("a GENUINELY unset knob still defaults — blank refusal must not swallow `undefined`", () => {
+    expect(resolveDataDir({})).toBe(ACCUMULUS_DEFAULT);
+    expect(resolveDataDir({ NUMISMA_DATA_DIR: undefined })).toBe(ACCUMULUS_DEFAULT);
```
AmetAlvirde: **INFO — no symmetry guard between the two resolvers.**

This file calls itself the anti-drift contract test, and `data-dir.ts:31` demands byte-identical behaviour across planes, but nothing asserts that `resolveDataDir` and `resolveSidecarPath` actually *agree*. They do agree today on `undefined` / `""` / `"   "` / relative — I checked all four — so this is a durability suggestion, not a defect.

Given the shell divergence flagged on `data-dir.ts`, a small shared table (input → default | throw) exercised against both resolvers would make a future one-sided edit fail here rather than in production. Cheap, and it is the property the two docstrings already claim in prose.

AmetAlvirde (commented):
Review pass (fresh reviewer, no authorship). The two-resolver change itself is correct, symmetric, and well-documented — the reversed `MUST` is handled properly: all three arms are recorded, and the CWD-hazard argument survives with an explicit `NEVER reintroduce resolve("")`. The ADR/INDEX corrections are required, not scope creep. `undefined`-still-defaults is pinned on both sides, and no new test can reach a real ledger (every fixture is `mkdtemp` under `tmpdir()`).

One substantive gap: **the caller sweep stopped at TypeScript, and the shell plane still implements the exact arm this PR outlaws.** Details inline.

packages/preferences/src/plans-reliable.test.ts
AmetAlvirde: Fixed in `d8dca09`.

Your diagnosis was right, and the suggested two-line inversion turned out not to be sufficient on its own: the two `toThrow` assertions sit *above* the value-capture block, so on a CWD-returning mutation they fire first and report `expected [Function] to throw an error` — which still never names CWD.

So the value capture moved to the **top** of the test, ahead of the `toThrow` pair, with explicit assertion messages:

```ts
const cwdFlavoured = join(process.cwd(), "plans.jsonl");
let produced: string | undefined;
try { produced = resolvePlansPath(""); } catch { produced = undefined; }
expect(produced, `resolvePlansPath("") must never resolve against the process CWD (${cwdFlavoured})`).not.toBe(cwdFlavoured);
expect(produced, 'resolvePlansPath("") must throw rather than return any path at all').toBeUndefined();
```

No line in the block is unfailable now, and the three failure modes report distinctly: the CWD line reds on a `resolve("")` return, `toBeUndefined` reds on a return of any *other* path (including the pre-#348 silent default), and the `toThrow` pair still pins the resolver's wording for both blank spellings.

Proven by mutation — `sidecar-io.ts` made to return `join(process.cwd(), fileName)` instead of throwing:

```
AssertionError: resolvePlansPath("") must never resolve against the process CWD
(/Users/amet/.../lane-d/plans.jsonl): expected '…' not to be '…'
```

Restored after; `git diff` on `sidecar-io.ts` empty.

packages/engine/src/data-dir.ts
AmetAlvirde: Fixed in `1c27fd4`. You were right on both arms, and the diagnosis of *why* the sweep missed it was the more valuable half.

`ops/price-feed/run-daily-fetch.sh` now reads the knob with the **unset-only** form and refuses either spelling of blank:

```sh
DATA_DIR="${NUMISMA_DATA_DIR-$HOME/Dev/accumulus/data}"
if [[ -z "${DATA_DIR//[[:space:]]/}" ]]; then
  # … loud message on stderr …
  exit 78
fi
```

Stripping all whitespace and testing what remains catches `""` and `"   "` with one predicate. Exit **78** is `EX_CONFIG`, chosen against the codes this wrapper already speaks — 1, 124, 127, 143 — so a reader can tell "the environment handed me a broken knob" apart from "the run tried and failed".

**Your inverted-criterion point is now recorded in the PR body**, since that is the part that generalizes: `${VAR:-anything}` is the hazard because the colon form substitutes on unset *or empty*; `${VAR-anything}` is the unset-only form matching the TS contract. The body states plainly that the original miss traces to the wrong criterion, because the next person to sweep this repo will reach for the criterion written there.

Two things the fix documents in place, both of which came out of chasing this:

- **Where the refusal surfaces.** The inherited stderr — under launchd, the plist's `StandardErrorPath`, `~/Library/Logs/numisma/launchd.price-feed.err.log`. Not the per-run log, which does not exist yet: the check must precede `HEARTBEAT_FILE`, which itself precedes the `mkdir -p "$LOG_DIR"` / `LOG_FILE=` / `exec … tee` block. `LOG_DIR` comes from its own env var and does not depend on `DATA_DIR`, so a blank data dir does not blind the logging.
- **There is deliberately no heartbeat for this failure.** The heartbeat lives *inside* `$DATA_DIR`, which is exactly what is broken. Inventing a fallback location so something gets written would be the defect #348 kills — guessing a path for a knob the operator got wrong. The comment says so, so a future edit does not "helpfully" relocate the check to get a breadcrumb out of it.

Your note that the safety of the 18:00 job rested on machine state rather than a code invariant also landed: the PR body now proves it from the rule instead — the refusal can only fire on a *misconfigured* variable, and an unset one still takes the default — with the installed-plist evidence kept only as corroboration.

packages/engine/src/data-dir.test.ts
AmetAlvirde: Not built here — tracked instead, in **#369**. Reasoning, since you asked for the property rather than a specific implementation.

**The concrete drift you were pointing at got closed, just not by a table.** Your check found the two resolvers agreeing, but the sweep that followed found two *more* resolvers that did not, and a fifth site that laundered a blank past all of them:

| resolver | `""` before | now |
| --- | --- | --- |
| `resolveDataDir` | default | refuses |
| `resolveSidecarPath` | default | refuses |
| `resolvePreferencesPath` | **`<cwd>/preferences.jsonl`** | refuses |
| `resolveEventStorePaths` | **`<cwd>/events.jsonl`** | refuses |
| `resolvePriceFeedPaths` | `resolve()` first, so the blank arrived **disguised** as absolute | refuses ahead of `resolve()` |

The bottom two took a default *parameter*, and a JS default fires on `undefined` and nothing else, so `""` sailed past into `resolve("")`. All five now agree on `undefined` and on both spellings of blank.

**Why the table is deferred rather than written now.** Writing it requires a decision this PR should not make quietly: **the resolvers do not agree on `~` either.** `resolveDataDir` expands it; `resolveSidecarPath` rejects it as non-absolute, since `isAbsolute("~/x")` is false. A shared input→outcome table has to state a tilde row, and transcribing today's behaviour would either bless the disagreement or silently change one side. That is a deliberate ruling, and it belongs with the relative-path arm — which is the *other* rule these same resolvers disagree about, and which is also out of scope here.

Both live in **#369**, which proposes exactly your shape — route every resolver through one shared validation, exercised by one table, so a future one-sided edit reds a test instead of surprising production — and records the tilde ruling as a prerequisite rather than a detail.

Resolving this thread because the finding has a home, not because it was dismissed. If you would rather the table land here despite the tilde question, say so and I will reopen it.

PR #366, opened 2026-08-18
fix(engine): split the fused zero-cost fallback — canonical-order attribution, and a discard for a lot-less leg

### LINKED ISSUE #329: the zero-cost tier-attribution fallback picks lot order, and invents c1 for empty lots
Split out of #299 item 4 under that issue's scope discipline ("fix the assertions, not the code under test — if any of the four turns out to hide a real defect, that becomes its own issue"). #299 adds a **pinning** test for this branch; the pin freezes the behavior described below deliberately, without endorsing it.

`tierWeightedDeltas` (`packages/engine/src/events/fold.ts:36-58`) has a degenerate fallback:

```ts
const present = TIERS.filter((tier) => costByTier.has(tier));
if (present.length === 0 || totalCost === 0) {
  // Degenerate (zero-cost lots): attribute everything to the first lot's Tier.
  return [{ tier: lots[0]?.tier ?? "c1", amount: sign * total }];
}
```

Two things are wrong with it, and the two halves of the condition are wrong in different ways.

## 1. The fallback reads lot insertion order; every other path reads canonical Tier order

`present` is built by filtering `TIERS` (`fold.ts:30`, `["c1","c2","c3"]`), so the proportional path at `:51-56` emits Tiers in canonical order regardless of how the caller ordered `lots`. The fallback instead returns `lots[0].tier` — **array insertion order**.

So for zero-cost lots the Tier that receives the entire cash delta depends on how the caller happened to order the lots array, not on anything about the position. Two callers holding the same position with the same lots in a different order get their reserve credited to different Tiers. Nothing else in this file behaves that way.

## 2. `present.length === 0` is exactly the empty-lots case, and it mints a `c1` attribution from nothing

Any lot at all puts a key in `costByTier` (`:41`), even a zero-cost one — a zero-cost lot yields `present` non-empty with `totalCost === 0`. So `present.length === 0` is reachable **only when `lots` is empty**, and in that case `lots[0]?.tier ?? "c1"` always resolves to the `?? "c1"` arm.

The result is a full-magnitude `c1` delta for a cash movement with no provenance whatsoever, which `applyReserveDelta` then books as real c1 capital. The two halves of the condition are therefore not one case: `totalCost === 0` is "lots exist, all free", `present.length === 0` is "no lots at all". Only the first is what the comment describes.

## Why file it rather than fix it in #299

#299 is a test-only pass by construction. Changing this would change folded reserve balances, which is a behavior change needing its own reasoning about what the right attribution *is* — proportional-by-quantity, canonical-first, or refusing the empty-lots case outright.

## Sequencing

`fold.ts` is rewritten by **#293 / spec #323** (the fold discard channel). That rewrite is the natural place to decide this — in particular, the empty-lots arm looks less like an attribution question and more like a **discard**: a delta with no provenance is exactly the kind of silent drop #323's channel exists to report. Worth deciding there rather than patching here.

The `#299` pin at `packages/engine/src/fold.test.ts` will go red when this is fixed. That is intended — it is the guard for the rewrite, and whoever changes the behavior updates the pin in the same change.

AmetAlvirde:
`tierWeightedDeltas` had one degenerate fallback covering two conditions that are wrong in different ways. This splits them and fixes each on its own terms.

Closes #329.

Three commits: the two original slices, then a third that makes the discard **total** after review found the second slice was losing real money. The third commit is the important one — the story of this PR is in it, not around it.

## The two arms

```ts
const present = TIERS.filter((tier) => costByTier.has(tier));
if (present.length === 0 || totalCost === 0) {
  return [{ tier: lots[0]?.tier ?? "c1", amount: sign * total }];
}
```

**`totalCost === 0` — lots exist, all free.** The whole delta went to `lots[0].tier`: **array insertion order**. Two callers holding the same position with the same lots ordered differently credited different tiers. Nothing else in this file behaves that way — the proportional path iterates `present`, which is already canonical. It now attributes to `present[0]`, the canonically first Tier holding a lot.

An even split across `present` was considered and rejected: with every cost weight zero there is no cost-basis weight to split by, so an even split would invent a weighting the data does not carry. `present[0]` is order-independent — which is exactly the defect being cured — and it degenerates to the old answer whenever one Tier holds the lots, the common shape.

**`present.length === 0` — no lots at all.** Any lot puts a key in `costByTier`, so this arm is reachable **only when `lots` is empty**, which means `lots[0]?.tier ?? "c1"` **always** resolved to the `?? "c1"` fallback. It minted a full-magnitude `c1` delta for a cash movement with no provenance whatsoever, and `applyToReserve` booked it as real c1 capital. That is not a bad attribution; it is an invented one. It is now a **discard**.

## The discard is TOTAL — and the first attempt at it was worse than the bug

The first version of this change dropped only the *cash leg* and let the rest of the arm run. Review caught it, and it was a genuine regression rather than a rough edge.

On `PositionClosed`, that shape recorded the skip and then still pushed a closed-book row carrying the full `settlement.proceeds` and still ran `positions.delete`. On `PositionTrimmed`, it still pushed the partial row and still mutated `trimming.lots`. **Realized P&L rose by the full proceeds while no reserve was ever credited** — the cash simply vanished. The invented-`c1` row this PR removes was wrong, but it at least *conserved the money*; a partial discard does not.

The file already forbade this in its own words. From the R8 comment inside the close arm:

> either the close applies IN FULL (cash leg, closed-book row, retirement) or it is recorded as discarded and NOTHING happens … the code would announce the drop and then act on its behalf.

That sentence was written to forbid precisely the shape the first attempt introduced. So the discard is now total: **no closed-book row, no retirement, no lot mutation, no registration.** `applyTieredLeg` returns a boolean — *may the rest of this event run?* — and each lot-derived arm `break`s on `false`.

This is deliberately the **same** semantics as the `position-absent` `else` branches already in these arms, which discard the whole event too. One discard concept, not two.

**The trade, stated plainly:** a lot-less close now leaves the position **open** rather than retiring it. That is the point. Compare the two failure modes — the old behaviour retired the position, credited full proceeds to realized P&L, and credited no reserve, *invisibly*; the new one leaves a position visibly open with a discard record naming exactly why. **Loud and inspectable beats silent and balanced-looking.**

## The discard goes through ADR-020's existing channel

`foldEvents` already returns `{data, skipped}` with `SkippedFoldEvent`, `SKIP_DETAIL` and `dedupeFoldSkips`. `tierWeightedDeltas` returns `[]`; the arm records the drop with the event's own locator and applies nothing. No new mechanism, and the `(eventId, reason)` dedupe key needed no change.

The four lot-derived call sites route through one `applyTieredLeg` closure. The four explicit-tier legs (Deposit, Withdraw, both Transfer legs) deliberately keep calling `applyToReserve` directly: their tiers come from the event itself rather than from lots, so they have no provenance that could be missing.

`PositionAddedTo` is the fourth lot-derived site and its discard branch is **unreachable** — `[event.lot]` is a one-element array literal, so the empty list cannot occur. It stays routed anyway, with the unreachability stated in code and the return deliberately ignored, because it *is* a lot-derived leg by kind: moving it to the explicit-tier path would silently lose the provenance check the day that site learns to add more than one lot.

### Why a new reason and not `reserve-absent`

The vocabulary was closed at `position-absent` / `reserve-absent`. This is a third kind, and it joins as **`provenance-absent`**.

Mapping it onto `reserve-absent` was the alternative and it was rejected as **dishonest, not merely coarse**. That reason's pinned prose reads *"The fold has no record of the reserve this event's cash leg names"* — which on this path is **a false statement**. The reserve exists and was found. What is missing is the cost-basis provenance the cash leg would inherit. A discard channel whose whole value is that it is stable and quotable cannot afford a notice that lies.

The same standard applies to the new member's own prose, and it is why the total discard matters beyond correctness: `provenance-absent`'s detail ends *"No balance moved."* Under the partial discard that sentence was **false** on the close and trim arms. It is true now.

**Precedence is a decision, not an accident:** when a leg has no lots, `provenance-absent` wins and the reserve is never consulted. There is nothing to apply, so whether the reserve exists is moot; also recording `reserve-absent` would report a miss that never happened. One event, one finding, and it is tested.

## The #299 pin was inverted, not preserved

`fold.test.ts` carried a pin that deliberately froze the old insertion-order behaviour so it could not change silently. It is **designed to go red on this fix**, and it did. It is inverted in the same change: both fixtures now expect `c1`, and the describe no longer reads "pinned, not endorsed" — it states the ruled behaviour and why.

The fixtures' deliberate c3-then-c1 and c2-then-c1 orderings are kept, because that ordering is what lets the test tell canonical order from insertion order. A new test turns the defect into a property: the same zero-cost lots in two array orders must produce the identical delta list.

## Where this can actually fire — corrected twice

The original reachability analysis in this PR was **wrong, and then wrong again after the first correction.** Both versions are worth recording, because the conclusion moved each time.

It first claimed the ingest gate admits a lot-less funding leg the fold discards, citing `crossref.ts:1354` and `:1480`. Then it narrowed to *"only `:1480` is reachable"*, since `:1354` passes `[event.lot]`.

**Neither is reachable.** Both call sites run *after* `parseEvent` — `crossref.ts:621` calls itself the *"Second ingest gate (after `parseEvent`)"*, and `crossReferenceOpen` takes an already-parsed `PositionOpenedEvent`. `parseLots` (`parse.ts:526-528`) rejects an empty `position.lots` outright. So no parsed `PositionOpened` can carry zero lots, and no sufficiency check is ever skipped. There is no guard that stopped guarding.

The one route that survives is a **`PositionTrimmed` whose removals name tiers the position holds no lots in** — `splitTierRemoval` returns `removed: []` when `tierTotal === 0` (`fold.ts:933`). That is also gate-closed (`parse.ts:569` requires a positive quantity, `crossref.ts:1245` rejects on sufficiency), so it is reachable **only** through callers that fold the durable log directly, bypassing both gates: `loadFoldedReview` at `packages/event-store/src/event-store.ts:225` calls `foldEvents` with no parse and no crossref.

**So: migration-shaped and historical, not user-injectable through ingest today.** That bounds the blast radius honestly without pretending the arm is dead — it is exactly why this ships as a recorded discard rather than as deleted code.

Issue #367 was filed on the original false premise and has been **corrected in the open**, with the retraction kept as a comment rather than edited away. It now asks the residual question, which is not about this arm at all: ingest-time validation does not constrain what the durable log already contains.

## Why this is landable without a replay

Two separate measurements, each licensing a different arm. They are **not** one figure.

- **`totalCost === 0`:** zero lots with `cost === 0` across the live ledger — 465 events plus 19 genesis lots, every cost nonzero.
- **`present.length === 0`:** zero calls into `tierWeightedDeltas` with an empty `lots` array, across a full replay of all 465 events.

The second one needed its own measurement because its precondition — no lots at all — never appears in the JSON, so no grep over event shapes could have answered it. Both arms are unreachable on real data today and **this change moves no real balance.**

The roadmap cites 452 events; the correct current figure is **465**. The log grew, the conclusion did not.

**How thin the live evidence is, stated deliberately.** Only 2 events in the whole ledger reach `tierWeightedDeltas` at all, and there is not one `PositionOpened`, `LotAdded` or `PositionTrimmed` event in the log — three of the four lot-derived call sites have never fired on real data. That strengthens the "no balance moves" claim and equally weakens real-data measurement as evidence about future behaviour. The change stands on its reasoning; the measurement bounds today's blast radius, not tomorrow's.

## What the blotter does now

An earlier draft of this body claimed `byTier` could now under-sum `totalRealizedPnlUsd`, because a lot-less close would land in `rows`, `byTempo` and the total but not in `byTier`. **That is no longer true, and the reason is the total discard:** a discarded close produces no row at all, so nothing lands anywhere and every blotter identity still balances. The under-sum was an artifact of the partial discard, and it went away with it.

What a reader should expect instead: a discarded close or trim leaves the position **open**, contributes nothing to the closed book, and appears in `skipped` as `provenance-absent`. `discardedEventCount` is the one live surfaced change.

## This PR knowingly leaves the discard vocabulary split

Making the provenance discard total exposes that **`reserve-absent` has the same partial-discard shape**, and always has. When the settlement reserve is missing, the close arm records the skip and then still pushes the row and still deletes the position — realized P&L rises with no reserve credited, which is the exact failure this PR just fixed, arrived at through a different miss.

**The inconsistency is a sharper problem than the defect**, and it is the reason this is called out rather than left to be discovered. `FoldSkipReason` is a deliberately closed vocabulary — `contracts.ts` says widening it is a spec change, not a build decision. After this PR it has three members and **two different meanings of "discarded"**:

| reason | what the fold does |
|---|---|
| `position-absent` | records the skip, applies nothing |
| `provenance-absent` | records the skip, applies nothing |
| `reserve-absent` | records the skip, **then pushes the row and deletes the position** |

Nothing in the type, the prose, or `SKIP_DETAIL` tells a reader which is which — the distinction lives only in the shape of the call sites. So the next person to add a reason code will inherit one of the two behaviours by whichever site they read first, without knowing a choice was being made. It shows up in the prose too: both members end *"No balance moved."*, and this PR made that true for `provenance-absent` while it remains **false** for `reserve-absent` on the close arm.

**Left unfixed deliberately.** It is pre-existing on `main`, outside #329's scope, and unreachable through ingest for the same reason the rest is (the cross-ref existence gate rejects an unknown reserve first). Fixing it here would widen a PR that has already reversed one premise mid-flight and retracted another. It is documented in the helper's docstring as *inherited rather than chosen*, so the next reader does not mistake it for a decision made here, and it is filed as **#371** — flagged, not smuggled in.

## What did *not* need changing, and why that is load-bearing

There is **no reason→label map anywhere** in the TUI or web discard surfaces — every consumer renders `reason` and `detail` straight off the record. A third vocabulary member therefore reaches every surface with **no code change**, which is why widening the union showed up nowhere in the full suite. That is a property of the ADR-020 design worth stating, not a gap in the search.

One thing it does **not** cover: the existing prose pin in `fold-discard-channel.test.ts:161-169` loops only over the skips its own fixture produces, so it never reaches `provenance-absent`. This PR writes a **separate** prose pin rather than assuming inherited coverage. If anyone later consolidates the two, the new member's prose silently falls out of every pin — so consolidate only by making the loop enumerate the union.

## Tests

Every test added or inverted here was **made to fail on purpose before being trusted** — mutation named, applied, red observed, restored. This board has already shipped two decorative guards that stayed green on the exact regression they existed to catch, and the first version of this PR nearly shipped a third: its four discard tests all pinned a lot-less *open*, a shape `parseLots` forbids, so they exercised an unreachable route while the reachable one went untested. The tests are now re-pointed at the trim route that can actually occur, and driven through `foldEvents` directly — the same door `loadFoldedReview` uses.

The discrimination matters as much as the red. Each mutation below was re-run independently rather than taken on report:

- **Restoring the old invented-`c1` return** reds **all six** discard tests and leaves **all four** zero-cost tests green. The two arms are independently guarded.
- **Recording the skip and letting the arm continue** — the partial discard this PR's third commit fixes — reds **exactly three** tests: the trim conservation test, the emptied-close test, and the lot-less-open test. The precedence and prose tests stay green under it, because they only read the skip record; the conservation tests are what guard the defect.
- **Swapping the two checks in `applyTieredLeg`** reds **only** the precedence test — 1 failed, 41 passed — so that test discriminates rather than duplicating its neighbours.

The parser-shape test earns its place as a tripwire, not a duplicate: it pins that `parseEvent` rejects `lots: []`, which is *the reason* the lot-less open is unreachable. If someone relaxes `parseLots`, that route goes live and this test reddens to say so.

## Suite

Full suite: **2287 passed, 48 skipped, 2 failed.**

One further observation, recorded rather than tidied away: an earlier full-suite run showed a **third** failure, `apps/price-feed/src/schedule-window.test.ts > the launchd fetch window > refuses a mark hour that is not an hour`. It passes 15/15 in isolation and did not reproduce on a second full run. It shells out to a real wrapper script, so parallel contention is the likely mechanism, and this PR's diff is confined to the engine's fold, which that assertion never reaches. Called a **flake, not a pass**, and filed as **#372** so the next lane in that harness has a prior observation to check against.

Both standing failures are in `ops/testkit/gitignored-path-globs.test.ts` — *"still excludes agent worktrees under .claude/ — the bug this closed"* and *"reaches nested .gitignore files, anchored to their own directory"* — and are **not from this change.** That guard derives its globs from `git ls-files --others --ignored`, which lists only ignored paths that exist on disk, so any fresh linked worktree fails them while the main checkout passes 8/8. Pre-existing from #362 and owned by #368, which merges ahead of this one. Nothing in `ops/` is modified here, and every other test in the repo passes.

### DIFF
--- a/packages/engine/src/contracts.ts
+++ b/packages/engine/src/contracts.ts
@@ -85,12 +85,21 @@ export interface FundReviewData {
 }
 
 /**
- * Why the fold could not apply an event. CLOSED AT TWO MEMBERS — a third reason is a
- * spec change, not a build decision. `position-absent`: the verb named a position the
- * fold has no record of (never opened in this window, or already retired).
+ * Why the fold could not apply an event. A CLOSED VOCABULARY — widening it is a spec
+ * change, not a build decision; it was widened once, deliberately, to three (#329).
+ *
+ * `position-absent`: the verb named a position the fold has no record of (never opened
+ * in this window, or already retired).
+ *
  * `reserve-absent`: a cash leg named a reserve the fold has no record of.
+ *
+ * `provenance-absent`: a cash leg derived from position lots had NO LOTS to derive
+ * from, so there is no cost-basis provenance for the leg to inherit. THIS IS NOT
+ * `reserve-absent` AND THE TWO MUST NEVER BE CONFLATED: here the reserve exists and was
+ * found — what is missing is the attribution the leg would have carried into it.
+ * Reporting a reserve miss that never happened would make the notice lie.
  */
-export type FoldSkipReason = "position-absent" | "reserve-absent";
+export type FoldSkipReason = "position-absent" | "reserve-absent" | "provenance-absent";
 
 /**
  * One event the fold read and then dropped — the Discard Channel's record (PRD #323,
--- a/packages/engine/src/events/fold.ts
+++ b/packages/engine/src/events/fold.ts
@@ -49,6 +49,9 @@ const SKIP_DETAIL: Record<FoldSkipReason, string> = {
   "reserve-absent":
     "The fold has no record of the reserve this event's cash leg names, so that leg was " +
     "read and then dropped. No balance moved.",
+  "provenance-absent":
+    "This event's cash leg has no cost-basis provenance to inherit — the fold has no " +
+    "lots to attribute it across — so the leg was read and then dropped. No balance moved.",
 };
 
 /**
@@ -91,7 +94,10 @@ export function dedupeFoldSkips(skipped: readonly SkippedFoldEvent[]): SkippedFo
  *
  * ZERO-COST LOTS get one delta, not a split: with every cost weight zero there is
  * no cost-basis weight to split by, so the whole delta lands on the canonically
- * first Tier that holds a lot (#329). */
+ * first Tier that holds a lot (#329).
+ *
+ * NO LOTS AT ALL RETURNS AN EMPTY LIST — the leg has no provenance to inherit, so
+ * there is nothing to attribute and the caller must discard it (#329). */
 function tierWeightedDeltas(lots: PositionLot[], total: number, sign: 1 | -1): TierDelta[] {
   const costByTier = new Map<CapitalTier, number>();
   let totalCost = 0;
@@ -102,8 +108,14 @@ function tierWeightedDeltas(lots: PositionLot[], total: number, sign: 1 | -1): T
   }
   const present = TIERS.filter((tier) => costByTier.has(tier));
   if (present.length === 0) {
-    // No lots at all. Left exactly as it was; ruled on separately (#329).
-    return [{ tier: lots[0]?.tier ?? "c1", amount: sign * total }];
+    // NO LOTS ⇒ NO PROVENANCE ⇒ NO DELTA. Any lot at all puts a key in `costByTier`,
+    // so this arm is reachable only when `lots` is empty. It used to return
+    // `[{ tier: lots[0]?.tier ?? "c1", … }]`, which — the `lots[0]` being undefined by
+    // construction — ALWAYS took the `?? "c1"` fallback and minted a full-magnitude c1
+    // delta for a cash movement carrying no provenance whatsoever; `applyToReserve`
+    // then booked it as real c1 capital. Not a bad attribution: an invented one. The
+    // empty list is the discard, and the caller records it (#329).
+    return [];
   }
   if (totalCost === 0) {
     // Zero-cost lots: no cost-basis weight to split by, so the whole delta goes to
@@ -128,7 +140,10 @@ function tierWeightedDeltas(lots: PositionLot[], total: number, sign: 1 | -1): T
 }
 
 /** Open → debit the funding Reserve by the cash leaving, split across the
- * position lots' own Tiers (each Tier pays for what it bought). */
+ * position lots' own Tiers (each Tier pays for what it bought).
+ *
+ * RETURNS `[]` FOR A LOT-LESS OPEN — no lots, no provenance, nothing to attribute.
+ * Callers must treat the empty list as a discard, never as a no-op worth booking. */
 export function reserveDeltasForOpen(lots: PositionLot[], amount: number): TierDelta[] {
   return tierWeightedDeltas(lots, amount, -1);
 }
@@ -141,7 +156,10 @@ export function reserveDeltasForOpen(lots: PositionLot[], amount: number): TierD
  * native and USD weights are proportionally identical, so the split is exact. When
  * `buildClosedPosition` reuses this helper to apportion proceeds against USD cost
  * basis, that native-vs-USD basis mismatch is the source of the documented
- * mixed-`entryFx` per-tier caveat; see that helper's note. Totals stay exact. */
+ * mixed-`entryFx` per-tier caveat; see that helper's note. Totals stay exact.
+ *
+ * RETURNS `[]` FOR A LOT-LESS CLOSE — no lots, no provenance, nothing to attribute.
+ * Callers must treat the empty list as a discard, never as a no-op worth booking. */
 export function reserveDeltasForClose(lots: PositionLot[], proceeds: number): TierDelta[] {
   return tierWeightedDeltas(lots, proceeds, 1);
 }
@@ -264,6 +282,32 @@ export function foldEvents(
       reason,
       detail: SKIP_DETAIL[reason],
     });
+  };
+  /**
+   * Apply one LOT-DERIVED cash leg, or record why it could not be applied.
+   *
+   * PRECEDENCE IS A DECISION: `provenance-absent` wins and the reserve is never
+   * consulted. With no deltas there is nothing to apply, so whether the reserve exists
+   * is moot — recording `reserve-absent` as well would report a reserve miss that never
+   * happened. One event, one finding, on this path.
+   *
+   * Only lot-derived legs route through here. The explicit-tier legs (Deposit,
+   * Withdraw, both Transfer legs) call `applyToReserve` directly: their deltas are
+   * never empty, and routing them here would imply a discard that cannot occur.
+   */
+  const applyTieredLeg = (
+    reserveId: string,
+    deltas: TierDelta[],
+    event: PortfolioEvent,
+    order: number,
+  ): void => {
+    if (deltas.length === 0) {
+      recordSkip(event, order, "provenance-absent");
+      return;
+    }
+    if (!applyToReserve(reserves, reserveId, deltas)) {
+      recordSkip(event, order, "reserve-absent");
+    }
   };
   let usdMxn = genesis.review.usdMxn;
   let latestAsOf = genesis.review.asOf;
@@ -391,9 +435,12 @@ export function foldEvents(
           costAnchors.set(position.instrumentId, anchor);
         }
         // Cash leg: debit the funding reserve. Sufficiency was proven at ingest.
-        if (!applyToReserve(reserves, event.funding.reserveId, reserveDeltasForOpen(position.lots, event.funding.amount))) {
-          recordSkip(event, order, "reserve-absent");
-        }
+        applyTieredLeg(
+          event.funding.reserveId,
+          reserveDeltasForOpen(position.lots, event.funding.amount),
+          event,
+          order,
+        );
         break;
       }
       case "PositionClosed": {
@@ -402,15 +449,12 @@ export function foldEvents(
         // closed position's mix, BEFORE retiring the asset leg. Honest-by-
         // construction: the asset cannot be removed without recording the cash.
         if (closing) {
-          if (
-            !applyToReserve(
-              reserves,
-              event.settlement.reserveId,
-              reserveDeltasForClose(closing.lots, event.settlement.proceeds),
-            )
-          ) {
-            recordSkip(event, order, "reserve-absent");
-          }
+          applyTieredLeg(
+            event.settlement.reserveId,
+            reserveDeltasForClose(closing.lots, event.settlement.proceeds),
+            event,
+            order,
+          );
           // Realized P&L: compute proceeds − cost
           // basis, tag it, and push a finished row onto the closed book INSTEAD of
           // silently dropping the position. Descriptive only — the profit already
@@ -461,15 +505,12 @@ export function foldEvents(
           // The mark the removed units are valued at for the NAV-honesty disclosure:
           // the latest PriceMarked seen so far, else the position's standing mark.
           const markPrice = latestMark.get(trimming.instrumentId) ?? trimming.markPrice;
-          if (
-            !applyToReserve(
-              reserves,
-              event.settlement.reserveId,
-              reserveDeltasForClose(removed, event.settlement.proceeds),
-            )
-          ) {
-            recordSkip(event, order, "reserve-absent");
-          }
+          applyTieredLeg(
+            event.settlement.reserveId,
+            reserveDeltasForClose(removed, event.settlement.proceeds),
+            event,
+            order,
+          );
           closedPositions.push(
             buildClosedPosition(
               trimming,
@@ -529,15 +570,12 @@ export function foldEvents(
               anchor.price = blended;
             }
           }
-          if (
-            !applyToReserve(
-              reserves,
-              event.funding.reserveId,
-              reserveDeltasForOpen([event.lot], event.funding.amount),
-            )
-          ) {
-            recordSkip(event, order, "reserve-absent");
-          }
+          applyTieredLeg(
+            event.funding.reserveId,
+            reserveDeltasForOpen([event.lot], event.funding.amount),
+            event,
+            order,
+          );
         } else {
           recordSkip(event, order, "position-absent");
         }
--- a/packages/engine/src/events/fold.ts
+++ b/packages/engine/src/events/fold.ts
@@ -302,6 +302,12 @@ export function foldEvents(
    * is moot — recording `reserve-absent` as well would report a reserve miss that never
    * happened. One event, one finding, on this path.
    *
+   * AND THE COST IS MEASURABLE, NOT MERELY SEMANTIC: `dedupeFoldSkips` keys on
+   * (`eventId`, `reason`), so two reasons for one event are two DISTINCT records that
+   * both survive dedupe — inflating `discardedEventCount` and emitting two
+   * `formatFoldDiscards` lines for a single drop. The channel would over-report the
+   * very thing it exists to report exactly once.
+   *
    * A RESERVE MISS IS NOT A DISCARD and returns `true`. That asymmetry is INHERITED,
    * not introduced here: a missing reserve has always let the arm carry on, and the
    * cross-ref existence gate rejects an unknown reserve before the fold ever runs, so
--- a/packages/engine/src/events/fold.ts
+++ b/packages/engine/src/events/fold.ts
@@ -87,7 +87,11 @@ export function dedupeFoldSkips(skipped: readonly SkippedFoldEvent[]): SkippedFo
 /** Cost-basis weight of each Tier in a position's lots (the provenance the cash
  * leg inherits). Returns signed `amount` per present Tier summing exactly to
  * `total`, with any float residual folded into the last Tier so the reserve's
- * authoritative `amount` lands exact. */
+ * authoritative `amount` lands exact.
+ *
+ * ZERO-COST LOTS get one delta, not a split: with every cost weight zero there is
+ * no cost-basis weight to split by, so the whole delta lands on the canonically
+ * first Tier that holds a lot (#329). */
 function tierWeightedDeltas(lots: PositionLot[], total: number, sign: 1 | -1): TierDelta[] {
   const costByTier = new Map<CapitalTier, number>();
   let totalCost = 0;
@@ -97,9 +101,20 @@ function tierWeightedDeltas(lots: PositionLot[], total: number, sign: 1 | -1): T
     totalCost += cost;
   }
   const present = TIERS.filter((tier) => costByTier.has(tier));
-  if (present.length === 0 || totalCost === 0) {
-    // Degenerate (zero-cost lots): attribute everything to the first lot's Tier.
+  if (present.length === 0) {
+    // No lots at all. Left exactly as it was; ruled on separately (#329).
     return [{ tier: lots[0]?.tier ?? "c1", amount: sign * total }];
+  }
+  if (totalCost === 0) {
+    // Zero-cost lots: no cost-basis weight to split by, so the whole delta goes to
+    // `present[0]` — the canonically first Tier holding a lot, in c1/c2/c3 order.
+    // Keying on `lots[0].tier` instead made attribution depend on ARRAY INSERTION
+    // ORDER, so two callers with the same lots ordered differently credited
+    // different Tiers (#329). An even split across `present` was rejected: with
+    // every weight zero it would invent a weighting the data does not carry.
+    // `present[0]` is order-independent and matches the old answer whenever
+    // exactly one Tier holds lots — the common shape.
+    return [{ tier: present[0]!, amount: sign * total }];
   }
   const deltas: TierDelta[] = [];
   let allocated = 0;
--- a/packages/engine/src/events/fold.ts
+++ b/packages/engine/src/events/fold.ts
@@ -284,30 +284,48 @@ export function foldEvents(
     });
   };
   /**
-   * Apply one LOT-DERIVED cash leg, or record why it could not be applied.
+   * Apply one LOT-DERIVED cash leg. RETURNS WHETHER THE REST OF THE EVENT MAY RUN.
+   *
+   * THE DISCARD IS TOTAL, WHICH IS WHY THIS IS A GATE AND NOT A VOID CALL. A
+   * lot-derived leg with no deltas has no cost-basis provenance, and an event whose
+   * cash leg cannot be attributed is dropped WHOLE: no closed-book row, no retirement,
+   * no lot mutation, no registration. `false` means the arm must do nothing at all.
+   * Recording the drop and then letting the arm carry on would raise realized P&L by
+   * the full proceeds while crediting NO reserve — the cash would simply vanish, which
+   * is strictly worse than the invented-`c1` attribution this replaced (that at least
+   * conserved the money). The `position-absent` `else` branches in these same arms
+   * already discard the whole event; this is deliberately the SAME semantics, not a
+   * second kind of discard.
    *
    * PRECEDENCE IS A DECISION: `provenance-absent` wins and the reserve is never
    * consulted. With no deltas there is nothing to apply, so whether the reserve exists
    * is moot — recording `reserve-absent` as well would report a reserve miss that never
    * happened. One event, one finding, on this path.
    *
-   * Only lot-derived legs route through here. The explicit-tier legs (Deposit,
-   * Withdraw, both Transfer legs) call `applyToReserve` directly: their deltas are
-   * never empty, and routing them here would imply a discard that cannot occur.
+   * A RESERVE MISS IS NOT A DISCARD and returns `true`. That asymmetry is INHERITED,
+   * not introduced here: a missing reserve has always let the arm carry on, and the
+   * cross-ref existence gate rejects an unknown reserve before the fold ever runs, so
+   * on the gated path it cannot happen at all. Only the provenance case gates.
+   *
+   * Only lot-derived legs route through here. The explicit-tier legs (`Deposit`,
+   * `Withdraw`, both `Transfer` legs) call `applyToReserve` directly: their tiers come
+   * from the event itself rather than from lots, so they have no provenance that could
+   * be missing, and routing them here would imply a discard that cannot occur.
    */
   const applyTieredLeg = (
     reserveId: string,
     deltas: TierDelta[],
     event: PortfolioEvent,
     order: number,
-  ): void => {
+  ): boolean => {
     if (deltas.length === 0) {
       recordSkip(event, order, "provenance-absent");
-      return;
+      return false;
     }
     if (!applyToReserve(reserves, reserveId, deltas)) {
       recordSkip(event, order, "reserve-absent");
     }
+    return true;
   };
   let usdMxn = genesis.review.usdMxn;
   let latestAsOf = genesis.review.asOf;
@@ -403,6 +421,22 @@ export function foldEvents(
     switch (event.type) {
       case "PositionOpened": {
         const { position } = event;
+        // Cash leg FIRST, because it gates the arm. Sufficiency was proven at ingest.
+        // A lot-less open has no provenance for the debit to inherit, and the discard
+        // is total: no registration, no cost anchor, no entry-VWAC fallback mark. The
+        // reserve and the position/close maps are disjoint, so hoisting the leg above
+        // the registration changes nothing an observer can see on the applying path —
+        // it only makes the drop able to stop the arm before it starts.
+        if (
+          !applyTieredLeg(
+            event.funding.reserveId,
+            reserveDeltasForOpen(position.lots, event.funding.amount),
+            event,
+            order,
+          )
+        ) {
+          break;
+        }
         const entryPrice = weightedAverageCost(position.lots);
         positions.set(position.id, {
           id: position.id,
@@ -434,47 +468,49 @@ export function foldEvents(
           pushClose(anchor);
           costAnchors.set(position.instrumentId, anchor);
         }
-        // Cash leg: debit the funding reserve. Sufficiency was proven at ingest.
-        applyTieredLeg(
-          event.funding.reserveId,
-          reserveDeltasForOpen(position.lots, event.funding.amount),
-          event,
-          order,
-        );
         break;
       }
       case "PositionClosed": {
         const closing = positions.get(event.positionId);
+        if (!closing) {
+          recordSkip(event, order, "position-absent");
+          break;
+        }
         // Cash leg: credit the settlement reserve with proceeds, tiered by the
         // closed position's mix, BEFORE retiring the asset leg. Honest-by-
         // construction: the asset cannot be removed without recording the cash.
-        if (closing) {
-          applyTieredLeg(
+        //
+        // AND IT GATES THE REST OF THE ARM. A lot-less position has no provenance for
+        // the proceeds to inherit, so the close is discarded WHOLE — the position stays
+        // open, no row is booked, nothing is retired. Booking the row while the cash
+        // leg dropped would raise realized P&L by the full proceeds against a reserve
+        // that never moved: the R8 failure below, in its other direction.
+        if (
+          !applyTieredLeg(
             event.settlement.reserveId,
             reserveDeltasForClose(closing.lots, event.settlement.proceeds),
             event,
             order,
-          );
-          // Realized P&L: compute proceeds − cost
-          // basis, tag it, and push a finished row onto the closed book INSTEAD of
-          // silently dropping the position. Descriptive only — the profit already
-          // landed in the Reserve above; the blotter records how the fund got here.
-          const settlementCurrency =
-            reserves.get(event.settlement.reserveId)?.currency ?? closing.currency;
-          closedPositions.push(
-            buildClosedPosition(closing, closing.lots, event.asOf, event.settlement.proceeds, settlementCurrency, usdMxn),
-          );
-          // R8 (ledger item 19): THE DELETE LIVES HERE NOW, not after the block. Moving
-          // it in is provably behavior-preserving — `Map.delete` on a key whose `get`
-          // just returned `undefined` is a no-op by definition — and it makes the arm a
-          // clean pair: either the close applies IN FULL (cash leg, closed-book row,
-          // retirement) or it is recorded as discarded and NOTHING happens. Leaving the
-          // delete outside while adding the `else` would be strictly worse than either
-          // shape: the code would announce the drop and then act on its behalf.
-          positions.delete(event.positionId);
-        } else {
-          recordSkip(event, order, "position-absent");
+          )
+        ) {
+          break;
         }
+        // Realized P&L: compute proceeds − cost
+        // basis, tag it, and push a finished row onto the closed book INSTEAD of
+        // silently dropping the position. Descriptive only — the profit already
+        // landed in the Reserve above; the blotter records how the fund got here.
+        const settlementCurrency =
+          reserves.get(event.settlement.reserveId)?.currency ?? closing.currency;
+        closedPositions.push(
+          buildClosedPosition(closing, closing.lots, event.asOf, event.settlement.proceeds, settlementCurrency, usdMxn),
+        );
+        // R8 (ledger item 19): THE DELETE LIVES INSIDE THE APPLYING PATH, not after the
+        // block — reached only once BOTH gates above have passed. The arm is a clean
+        // pair: either the close applies IN FULL (cash leg, closed-book row,
+        // retirement) or it is recorded as discarded and NOTHING happens. Announcing
+        // the drop and then acting on its behalf — pushing the row, retiring the asset
+        // — is the shape this guards against, whichever gate did the announcing.
+        positions.delete(event.positionId);
         break;
       }
       case "PositionTrimmed": {
@@ -505,12 +541,23 @@ export function foldEvents(
           // The mark the removed units are valued at for the NAV-honesty disclosure:
           // the latest PriceMarked seen so far, else the position's standing mark.
           const markPrice = latestMark.get(trimming.instrumentId) ?? trimming.markPrice;
-          applyTieredLeg(
-            event.settlement.reserveId,
-            reserveDeltasForClose(removed, event.settlement.proceeds),
-            event,
-            order,
-          );
+          // Cash leg, AND the gate on the rest of the arm. `splitTierRemoval` yields no
+          // removed lots when the removal names a tier the position holds nothing in, so
+          // the proceeds have no provenance to inherit and the trim is discarded WHOLE:
+          // no partial row, and `trimming.lots` is left exactly as it was. Pushing the
+          // row while the credit dropped would book realized P&L against a reserve that
+          // never moved; mutating the lots on top of that would retire the asset for
+          // cash the fund never received.
+          if (
+            !applyTieredLeg(
+              event.settlement.reserveId,
+              reserveDeltasForClose(removed, event.settlement.proceeds),
+              event,
+              order,
+            )
+          ) {
+            break;
+          }
           closedPositions.push(
             buildClosedPosition(
               trimming,
@@ -570,6 +617,16 @@ export function foldEvents(
               anchor.price = blended;
             }
           }
+          // ROUTED THROUGH THE GATE THOUGH THE DISCARD IS UNREACHABLE HERE, AND THE
+          // RETURN IS DELIBERATELY IGNORED. The array literal `[event.lot]` always
+          // holds exactly one lot, so `reserveDeltasForOpen` always yields exactly one
+          // delta and `provenance-absent` cannot fire on this site — there is nothing
+          // left of the arm to gate anyway, the append happened above. It stays routed
+          // because this IS a lot-derived leg: its tier comes from `event.lot.tier`, so
+          // it belongs on the lot-derived path by kind, not by whether today's argument
+          // happens to be non-empty. Calling `applyToReserve` directly instead would
+          // put a lot-derived leg on the explicit-tier path and quietly lose the
+          // provenance check the day this site learns to add more than one lot.
           applyTieredLeg(
             event.funding.reserveId,
             reserveDeltasForOpen([event.lot], event.funding.amount),
--- a/packages/engine/src/fold.test.ts
+++ b/packages/engine/src/fold.test.ts
@@ -9,7 +9,9 @@ import {
   buildCompositionReport,
   buildEventReference,
   crossReferenceEvent,
+  EVENT_SCHEMA_VERSION,
   foldEvents,
+  parseEvent,
   parseFundReview,
   reserveDeltasForClose,
   reserveDeltasForOpen,
@@ -19,6 +21,7 @@ import {
   type PositionLot,
   type PositionOpenedEvent,
   type PositionClosedEvent,
+  type PositionTrimmedEvent,
   type PriceMarkedEvent,
 } from "./index.js";
 import { describe, expect, it } from "vitest";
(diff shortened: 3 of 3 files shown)


packages/engine/src/events/fold.ts
```diff
@@ -387,15 +449,12 @@ export function foldEvents(
...
-          if (
-            !applyToReserve(
-              reserves,
-              event.settlement.reserveId,
-              reserveDeltasForClose(closing.lots, event.settlement.proceeds),
-            )
-          ) {
-            recordSkip(event, order, "reserve-absent");
-          }
+          applyTieredLeg(
```
AmetAlvirde: **The discard drops a real cash CREDIT here, and the arm applies anyway.**

The PR's justification — "it minted c1 capital for a cash movement with no provenance" — holds for the *debit* direction (open/add): not booking a debit leaves the reserve overstated but no asset was booked either. On a **close it inverts**: `settlement.proceeds` is money that actually arrived. Returning `[]` means it is never credited to any reserve, so the reserve is understated by the full proceeds from here to the end of the fold.

Worse, the discard is not a discard of the *event* — only of the cash leg. The arm still runs `buildClosedPosition(...)` (whose `totalRealizedPnlUsd` counts the full `proceedsUsd` against a zero cost basis) and still runs `positions.delete(...)`. So the blotter books a realized gain that no reserve ever received.

That also falsifies this arm's own stated invariant, three lines below: *"either the close applies IN FULL (cash leg, closed-book row, retirement) or it is recorded as discarded and NOTHING happens."* After this change the close can be recorded as discarded **and** retire the position **and** push the row.

Scenario: fold a `PositionClosed` on a lot-less position with `proceeds: 250` against a 1000 reserve. Before: reserve 1250, blotter +250. After: reserve 1000, blotter still +250, one `provenance-absent` skip whose prose says no balance moved.

packages/engine/src/events/fold.ts
```diff
@@ -250,6 +283,32 @@ export function foldEvents(
...
+   * PRECEDENCE IS A DECISION: `provenance-absent` wins and the reserve is never
+   * consulted. With no deltas there is nothing to apply, so whether the reserve exists
+   * is moot — recording `reserve-absent` as well would report a reserve miss that never
+   * happened. One event, one finding, on this path.
+   *
+   * Only lot-derived legs route through here. The explicit-tier legs (Deposit,
+   * Withdraw, both Transfer legs) call `applyToReserve` directly: their deltas are
+   * never empty, and routing them here would imply a discard that cannot occur.
+   */
+  const applyTieredLeg = (
```
AmetAlvirde: **Finding 1 — Medium. The discard is partial: only the cash leg drops, and the `PositionClosed` arm now contradicts its own invariant.**

`applyTieredLeg` suppresses the reserve mutation and records the drop, but every *other* effect of the arm still runs:

- **`PositionOpened`** — the ghost position is still inserted into `positions`, and a `Close` anchor is still pushed for the instrument at `entryPrice = weightedAverageCost([])`, which is `0` (`fold.ts:979-986`). So a lot-less open now plants a zero-price point in that instrument's journey and books a zero-lot position, with no offsetting debit.
- **`PositionClosed`** — the closed-book row is still pushed and the position is still retired.
- **`PositionTrimmed`** — the row is still pushed and `trimming.lots = working` still lands.

The comment at `fold.ts:467-474` states the close arm is "a clean pair: either the close applies IN FULL (cash leg, closed-book row, retirement) or it is recorded as discarded and NOTHING happens." A `provenance-absent` close is now a **third state**: recorded as discarded *and* applied in full except for the cash. I accept that `reserve-absent` already had this shape and the comment was therefore already imprecise — but this PR widens the set of events that land in it, and leaves a comment standing that a reader will take as a guarantee.

**Recommendation:** amend that comment to name the third state explicitly, and state a ruling for whether a `provenance-absent` close/trim should also suppress the closed-book row. If the ruling is "no, the row stays", say so there — it is defensible (the row is descriptive), but it should be a decision on the page rather than a leftover.

packages/engine/src/events/fold.ts
```diff
@@ -49,6 +49,9 @@ const SKIP_DETAIL: Record<FoldSkipReason, string> = {
   "reserve-absent":
     "The fold has no record of the reserve this event's cash leg names, so that leg was " +
     "read and then dropped. No balance moved.",
+  "provenance-absent":
+    "This event's cash leg has no cost-basis provenance to inherit — the fold has no " +
+    "lots to attribute it across — so the leg was read and then dropped. No balance moved.",
```
AmetAlvirde: **Finding 2 — Medium. "No balance moved." is false on two of the four routed arms.**

The sentence is inherited verbatim from `reserve-absent`, where the event applies nothing at all. Here it is attached to arms that apply plenty (see finding 1): on a `PositionTrimmed` the removed lots leave the position and a realized row is booked while no cash arrives — NAV moves. On a `PositionClosed` the position is retired.

This is **not** a violation of the two pinned rules — no figure, no event content, and the added test correctly proves both. It is a truthfulness problem, on the channel whose stated value is that its notices are stable and quotable. The PR body itself argues, persuasively, that mapping onto `reserve-absent` was rejected because "a discard channel whose whole value is that it is stable and quotable cannot afford a notice that lies." The same standard applies to this sentence.

**Recommendation:** narrow it to what is actually true of every routed arm — e.g. `"...so the leg was read and then dropped. The reserve was not touched."` That keeps the reassurance, drops the overclaim, and stays inside both pinned rules.

packages/engine/src/events/fold.ts
```diff
@@ -446,15 +505,12 @@ export function foldEvents(
...
-              reserves,
-              event.settlement.reserveId,
-              reserveDeltasForClose(removed, event.settlement.proceeds),
-            )
-          ) {
-            recordSkip(event, order, "reserve-absent");
-          }
+          applyTieredLeg(
+            event.settlement.reserveId,
+            reserveDeltasForClose(removed, event.settlement.proceeds),
```
AmetAlvirde: **Finding 3 — Medium. There is a second route to the empty arm, and the gate admits it today.**

The body's reachability argument is that the arm is reachable "only when `lots` is empty," and then reasons about call sites as if the only candidate were a lot-less `PositionOpened`. That skips this site, where the argument is `removed` — a *derived* array, not a position's lots.

`splitTierRemoval` returns `removed: []` when the named tier holds nothing (`fold.ts:933`, documented as "a degenerate empty tier is a safe no-op"). And a zero-quantity removal on an unheld tier clears every gate:

- sufficiency (`crossref.ts:1245-1254`) tests `requested > available + 1e-9` — `0 > 0` is false, so it passes;
- full-retirement (`crossref.ts:1264`) sees `totalRequested === 0`, so it passes;
- settlement-magnitude (`crossref.ts:1276-1281`) computes `expected = 0 * price = 0` and is guarded by `if (expected > 0)`, so it never runs.

So `PositionTrimmed` with `removals: [{ tier: "c3", quantity: 0 }]` on a c1-only position reaches `reserveDeltasForClose([], proceeds)` with an arbitrary `proceeds`.

The empirical claim in the body is **unaffected** — the replay measured calls with an empty `lots` argument, which covers `removed` too. What is wrong is the *reasoning* about future reachability, and it matters because the behaviour delta on this route is larger than the body describes: previously such a trim credited the **full proceeds** to invented `c1`; it now credits **nothing** while still pushing a closed-book row for the full proceeds (finding 7).

**Recommendation:** name the `removed` route in the body and in the `tierWeightedDeltas` docstring — "reachable only when the argument array is empty, which is a lot-less position *or* a removal that resolves to no lots" — and add a test at this seam. It is the only one of the four routed sites where the arm is reachable through the gate.

packages/engine/src/fold.test.ts
```diff
@@ -884,3 +906,95 @@ describe("tier-weighted deltas — the zero-cost degenerate fallback (pinned, no
...
+      { reserveId, amount: 400 },
+    );
+  }
+
+  it("returns an empty delta list for empty lots on both legs", () => {
+    expect(reserveDeltasForOpen([], 400)).toEqual([]);
+    expect(reserveDeltasForClose([], 250)).toEqual([]);
+  });
+
+  it("records one provenance-absent skip AND leaves the reserve balance untouched", () => {
```
AmetAlvirde: **Finding 4 — Medium. This test under-discriminates on exactly the property it exists to assert — and it is why finding 1 went unnoticed.**

The comment is right that both halves matter. But the second half checks only `reserves[0].amount`, and "read and then dropped" is a claim about the **whole read model**, not about one reserve. `folded.data` is materially mutated by this same event:

- `data.positions` gains `ghost-core` — zero lots, `markPrice: 0`;
- `data.closes` gains `{ instrumentId: "btc-usd", asOf: "2026-06-02", price: 0 }`, because `seededInstruments` had not seen `btc-usd` and `weightedAverageCost([])` is `0`.

Neither is asserted, so the test is green on a fold that did quite a lot.

The repo already has the right shape for this, in the sibling channel test — `fold-discard-channel.test.ts:124-128` builds an `untouched` baseline from an empty log and asserts `expect(folded.data).toEqual({ ...untouched, review: { ...untouched.review, asOf } })`. Had this test been written that way, it would have gone red on the ghost position and surfaced finding 1 at authoring time rather than at review time.

Given the board's history with decorative guards, this is the one I would insist on. **Recommendation:** assert the full `data` shape against an `untouched` baseline. If the ghost position and the zero-price anchor are the ruled behaviour, the test should *document* them explicitly rather than not look.

packages/engine/src/events/fold.ts
```diff
@@ -250,6 +283,32 @@ export function foldEvents(
       detail: SKIP_DETAIL[reason],
     });
   };
+  /**
+   * Apply one LOT-DERIVED cash leg, or record why it could not be applied.
+   *
+   * PRECEDENCE IS A DECISION: `provenance-absent` wins and the reserve is never
+   * consulted. With no deltas there is nothing to apply, so whether the reserve exists
+   * is moot — recording `reserve-absent` as well would report a reserve miss that never
+   * happened. One event, one finding, on this path.
```
AmetAlvirde: **Finding 5 — Low. The precedence decision is right, and its strongest argument is missing.**

One-event-one-finding is correct here, and the two reasons given (nothing to apply; a reserve miss that never happened) both hold. There is a third and more concrete one the comment does not make: `dedupeFoldSkips` keys on `(eventId, reason)`, so recording both would survive dedupe as **two** distinct records — inflating `HeadDigest.discardedEventCount` (`durable-log.ts:101`) and emitting two lines from `formatFoldDiscards` for a single event. That is a measurable consequence, not just a semantic preference, and it is the argument that makes the ruling unarguable.

**Recommendation:** add that clause. No code change.

packages/engine/src/events/fold.ts
```diff
@@ -514,15 +570,12 @@ export function foldEvents(
...
-              reserves,
-              event.funding.reserveId,
-              reserveDeltasForOpen([event.lot], event.funding.amount),
-            )
-          ) {
-            recordSkip(event, order, "reserve-absent");
-          }
+          applyTieredLeg(
+            event.funding.reserveId,
+            reserveDeltasForOpen([event.lot], event.funding.amount),
```
AmetAlvirde: **Finding 6 — Low. Exclusions verified correct; this one's discard branch is dead.**

I checked all eight `applyToReserve` sites named in its docstring (`fold.ts:958-964`). Four route through `applyTieredLeg`, four stay direct, and the split is right: `Deposit`, `Withdraw` and both `Transfer` legs build their deltas from an explicit `tier` field and can never produce an empty list, so routing them would advertise a discard that cannot occur. **No lot-derived site was missed.**

One nit within the routed four: this site passes `[event.lot]` — always length 1 — so its `deltas.length === 0` branch is unreachable, exactly as the body notes for the mirror gate site at `crossref.ts:1354`. Routing it anyway for uniformity is the right call, but the `applyTieredLeg` docstring's framing ("only lot-derived legs route through here" / explicit-tier legs excluded because their discard "cannot occur") implies all four routed legs *can* discard, when only three can.

**Recommendation:** one clause — "the add leg routes here for uniformity; its single-lot array can never be empty."

packages/engine/src/contracts.ts
```diff
@@ -85,12 +85,21 @@ export interface FundReviewData {
...
+ * `position-absent`: the verb named a position the fold has no record of (never opened
+ * in this window, or already retired).
+ *
  * `reserve-absent`: a cash leg named a reserve the fold has no record of.
+ *
+ * `provenance-absent`: a cash leg derived from position lots had NO LOTS to derive
+ * from, so there is no cost-basis provenance for the leg to inherit. THIS IS NOT
+ * `reserve-absent` AND THE TWO MUST NEVER BE CONFLATED: here the reserve exists and was
+ * found — what is missing is the attribution the leg would have carried into it.
+ * Reporting a reserve miss that never happened would make the notice lie.
```
AmetAlvirde: **Finding 7 — Low. `byTier` under-sum claim verified exactly — and it generalises further than the body says.**

Traced through `buildClosedPosition` (`fold.ts:837-911`): with `lots: []`, the `costUsdByTier` loop never runs, `reserveDeltasForClose([], proceedsUsd)` now returns `[]` so `proceedsByTier` stays empty, the `TIERS` loop `continue`s on every tier, and `tierAttribution` is `[]` — while `realizedPnlUsd = proceedsUsd - 0 = proceedsUsd`. `rows`, `byTempo` and `totalRealizedPnlUsd` all still carry it. **The body's claim is exactly right**, including the judgement that the old balance was fictional.

Two things it misses:

1. The row now reports a full realized **profit that no reserve ever received**. The under-sum is the symptom; the blotter showing realized P&L with no matching cash movement is the thing a reader will actually hit.
2. The same applies on the **trim** arm (finding 3), where it is worse — the position survives, so the fund carries a live position whose partial row claims realized proceeds that never landed.

**Recommendation:** extend the body's blotter paragraph to the trim arm and to the no-matching-cash framing.

---

**Finding 8 — Info. Consumer sweep confirms the body; nothing breaks.**

- Exactly one exhaustive `Record<FoldSkipReason, …>` exists — `SKIP_DETAIL` at `fold.ts:44` — and it is widened here. `preferences.ts:86` has a same-named map over a *different* union (`PreferenceSkipReason`) and is untouched.
- The only two `switch` statements on a `.reason` in the TUI (`import-orders-unattributed-refusal.ts:74`, which has a `never` exhaustiveness check, and `plans-report.ts:145`) are over unrelated unions.
- `durable-log.ts:101`, `event-store.ts:293` and `:345`, `ingest-walk.ts:200/221/240`, `apps/tui/src/event-store.ts:349` all go through `dedupeFoldSkips` and render `reason`/`detail` verbatim. None rejects, gates or branches on a discard, so a third member reaches every surface with no code change — the body's read of the ADR-020 design is correct.

One live behaviour change the body does not list: a lot-less leg now increments `HeadDigest.discardedEventCount` and breaks the "a clean fold says nothing" silence in `unattendedFoldVerdict`, where previously it produced no discard at all. Benign and arguably the point, but it is a surfaced change.

I also confirmed the body's claim about the existing prose pin: `fold-discard-channel.test.ts:141-169` loops only over its own two-event fixture (`Deposit` + `PositionClosed`), so it genuinely never reaches `provenance-absent`, and the separate pin here is the right call rather than a duplicate.

packages/engine/src/fold.test.ts
```diff
@@ -864,7 +870,23 @@ describe("tier-weighted deltas — the zero-cost degenerate fallback (pinned, no
...
+    const c3First: PositionLot[] = [
+      { quantity: 4, cost: 0, tier: "c3" },
+      { quantity: 6, cost: 0, tier: "c2" },
+    ];
+    const c2First: PositionLot[] = [
+      { quantity: 6, cost: 0, tier: "c2" },
+      { quantity: 4, cost: 0, tier: "c3" },
+    ];
+
+    expect(reserveDeltasForClose(c3First, 250)).toEqual(reserveDeltasForClose(c2First, 250));
```
AmetAlvirde: **Finding 9 — Info. Test discrimination, checked claim by claim.**

This is the strongest addition in the PR and it genuinely discriminates. The fixtures use `c3`/`c2` rather than reusing the pins' `c1`, so `lots[0].tier` differs across the two orders (`c3` vs `c2`) while `present[0]` is `c2` in both — meaning it reds on a revert to `lots[0].tier` **independently** of the two inverted pin assertions, which key on `c1`. That independence is what makes it a property test rather than a third copy of the pin.

On the body's other two mutation claims, which I checked by reading rather than by mutating:

- *"Restoring the old invented-c1 return reds all four discard tests, leaves all four zero-cost tests green"* — holds; the restoration touches only the `present.length === 0` arm, which the zero-cost fixtures never enter.
- *"Swapping the two checks reds **only** the precedence test"* — holds, and it is worth spelling out why, because it is not obvious: a swap that keeps both checks calls `applyToReserve(reserves, id, [])` on an **existing** reserve, which returns `true` (`fold.ts:966-977` — the empty delta list is applied as a no-op), so control falls through to the length check and `provenance-absent` is still recorded. That is what leaves the skip-and-balance and prose tests green. The claim is right; the mechanism deserves a sentence, because a reader checking it will otherwise expect two reds.

The prose test is a faithful copy of the established pin and would fail on a figure or on leaked event content. The only real gap in the added coverage is finding 4.

AmetAlvirde (commented):
## Fresh review — PR #366

I reviewed this with no prior context on the board. **The two arms are individually correct and the reasoning behind each is sound.** `present[0]` is the right cure for the insertion-order defect; the discard is the right answer for a lot-less leg; the precedence decision is right and the four explicit-tier exclusions are correct. The `byTier` under-sum claim checks out against `buildClosedPosition` exactly as written.

What I found are four things the change does that the change does not say, and one test that under-discriminates on the very property the channel exists to assert.

**Suite:** ran `npx vitest run` in `lane-e`. **2284 passed, 48 skipped, 2 failed.** Both failures are `ops/testkit/gitignored-path-globs.test.ts` — *"still excludes agent worktrees under .claude/ — the bug this closed"* and *"reaches nested .gitignore files, anchored to their own directory"*. Out of scope per the brief; matches the body exactly. No other failure.

### Findings

| # | Sev | Summary |
|---|-----|---------|
| 1 | Medium | The discard is **partial** — only the cash leg drops; the arm's other effects still apply, and the `PositionClosed` in-code invariant now reads false |
| 2 | Medium | `SKIP_DETAIL`'s *"No balance moved."* is a false claim on the close and trim arms |
| 3 | Medium | Reachability analysis misses a second, **gate-admitted** route to the empty arm: `PositionTrimmed`'s `removed` |
| 4 | Medium | The flagship discard test asserts only the reserve amount, not that `data` is otherwise untouched — it would stay green through F1 |
| 5 | Low | Precedence reasoning is sound; its strongest argument (`dedupeFoldSkips` double-count) is missing from the comment |
| 6 | Low | Exclusions verified correct; `PositionAddedTo`'s discard branch is dead, and the helper docstring implies otherwise |
| 7 | Low | `byTier` under-sum verified — and it generalises to the trim arm, which the body does not mention |
| 8 | Info | Consumer sweep confirms the body: nothing breaks. One live surfaced change (`discardedEventCount`) |
| 9 | Info | Test discrimination checked claim by claim — the order-independence test is the strongest addition; inverting the #299 pin was right |

Detail is in the inline comments. Findings 1–4 are the ones I would want addressed before merge; 1 and 4 are the pair — 4 is why 1 was not noticed.

### On the #299 pin (asked directly)

**Inverting was right, not preserving.** That pin recorded *observed* behaviour and said so out loud ("pinned, not endorsed", "not an argument that ... is the right answer"). A pin that disclaims endorsement is a record, not a contract — preserving it would have frozen the defect it was written to expose. It also did its job: it went red on the fix, which is the whole reason it existed. The inverted fixtures correctly keep the deliberate `c3`-then-`c1` / `c2`-then-`c1` orderings, which is what still lets them tell canonical order from insertion order.

### On the new prose (asked directly)

It obeys **both** pinned rules — no figure (`not.toMatch(/\d/)` passes), no substring of event content. Register and shape match the two existing members. My only objection is its final sentence, which is finding 2.

packages/engine/src/events/fold.ts
```diff
@@ -446,15 +505,12 @@ export function foldEvents(
...
-          if (
-            !applyToReserve(
-              reserves,
-              event.settlement.reserveId,
-              reserveDeltasForClose(removed, event.settlement.proceeds),
-            )
-          ) {
-            recordSkip(event, order, "reserve-absent");
-          }
+          applyTieredLeg(
```
AmetAlvirde: **This is the one call site where the empty list is genuinely reachable — and it is the one where the partial apply is worst.**

`removed` is empty not only when the position is lot-less. `splitTierRemoval` returns `{ removed: [], remaining: … }` whenever `tierTotal === 0` (`fold.ts:933`), i.e. whenever a removal names a tier the position does not hold. The trim gate above (`crossref.ts:1245-1254`) rejects `requested > available`, which catches that shape for anything arriving through cross-ref — but any ungated `foldEvents` caller (the legacy migration and pre-rule history this file elsewhere calls out) can hand this arm a trim whose removals miss the held tiers, and `removed` comes back empty with **non-empty position lots**.

Consequences on that path, all after the `provenance-absent` skip is recorded:

- `event.settlement.proceeds` is never credited — real cash silently vanishes from the reserve;
- `buildClosedPosition(..., removed, ..., partial=true)` is still pushed, with zero cost basis, so the partial row reports the whole proceeds as realized gain, and `markVsFill.markValueUsd` is 0 (removed quantity 0) so `deltaUsd` equals the full proceeds;
- `trimming.lots = working` still executes.

The skip's prose then reads "the leg was read and then dropped. No balance moved" for an event that moved the closed book and rewrote the position. Either the whole trim should be discarded when `removed` is empty, or the cash leg should still be applied and only the attribution reported as missing.

packages/engine/src/events/fold.ts
```diff
@@ -49,6 +49,9 @@ const SKIP_DETAIL: Record<FoldSkipReason, string> = {
   "reserve-absent":
     "The fold has no record of the reserve this event's cash leg names, so that leg was " +
     "read and then dropped. No balance moved.",
+  "provenance-absent":
```
AmetAlvirde: `"…so the leg was read and then dropped. No balance moved."`

That sentence is accurate only on the `PositionOpened` arm. On the other three routes through `applyTieredLeg` the surrounding arm keeps going after the skip is recorded: `PositionClosed` still pushes a closed-book row and deletes the position; `PositionTrimmed` still pushes a partial row and rewrites `trimming.lots`; `PositionAddedTo` has already appended `event.lot` and possibly rewritten `markPrice` and the fold's cost anchor before the call.

The file's own rationale for fixed prose is that these lines "stay stable and quotable" on the unattended daily surface. A quotable line telling the reader nothing happened, on an event that retired a position and booked realized P&L, is the same class of defect as the `reserve-absent` prose the PR body correctly refuses to reuse here. Either drop the no-state-moved claim from this member's wording, or make the arms honest so the claim is true.

packages/engine/src/fold.test.ts
```diff
@@ -884,3 +906,95 @@ describe("tier-weighted deltas — the zero-cost degenerate fallback (pinned, no
     ]);
   });
 });
+
+describe("tier-weighted deltas — a lot-less cash leg is DISCARDED, not attributed (#329)", () => {
```
AmetAlvirde: **Every test in this describe exercises a shape the parser forbids.**

`lotlessOpen` builds a `PositionOpened` with `lots: []`, but `parseLots` rejects exactly that — `packages/engine/src/events/parse.ts:528`, *"Position requires at least one lot."* — and genesis positions are held to the same rule by `packages/engine/src/parse.ts:349` (*"must be a non-empty array"*). So a lot-less `PositionOpened` cannot reach `foldEvents` through any parsed path; it only exists when a test hand-builds the record.

Meanwhile the one call site that can produce an empty delta list from a *well-formed* event — `PositionTrimmed` whose removals all miss the held tiers, so `splitTierRemoval` returns `removed: []` (`fold.ts:933`) — has no test at all, and it is the site where the partial apply actually loses money (see the comment on `fold.ts:508`).

The mutation-testing discipline described in the PR body is sound, but it was pointed at the branch that cannot fire. Please add a fold-level test for a lot-less **trim** and a lot-less **close**, asserting what happens to `reserves`, `closedPositions` and `positions` — not just to `skipped`.

packages/engine/src/events/fold.ts
```diff
@@ -113,7 +140,10 @@ function tierWeightedDeltas(lots: PositionLot[], total: number, sign: 1 | -1): T
 }
 
 /** Open → debit the funding Reserve by the cash leaving, split across the
- * position lots' own Tiers (each Tier pays for what it bought). */
+ * position lots' own Tiers (each Tier pays for what it bought).
+ *
+ * RETURNS `[]` FOR A LOT-LESS OPEN — no lots, no provenance, nothing to attribute.
```
AmetAlvirde: **The reachability analysis behind #367 points at the wrong site.**

The PR body says of the two gate call sites: *"Of the two, only `:1480` is reachable"* — i.e. `PositionOpened` with empty `lots` is the live question for the sufficiency gate.

It is not reachable either. `crossReferenceOpen` runs *after* parsing, and `parseLots` already refuses an empty `position.lots` (`packages/engine/src/events/parse.ts:528`). So no parsed `PositionOpened` can carry zero lots, `reserveDeltasForOpen` at `crossref.ts:1480` never receives an empty array, and the "guard that stops guarding" filed as #367 cannot fire on that path.

Which matters because it moves where the risk actually lives: the only well-formed event that can drive `tierWeightedDeltas` to `[]` is a `PositionTrimmed` whose removals land on unheld tiers (`splitTierRemoval`, `fold.ts:933`), and that one never goes through a `reserveDeltas*` sufficiency check to begin with — it is a credit. Worth correcting in #367 before someone rules on a hole that isn't there and misses the one that is.

AmetAlvirde:
## Correction + amendment to my review above (`#pullrequestreview-4957139771`)

A second reviewer pass surfaced two parser facts I did not check before posting. One **refutes the worked example in my finding 3**, one **narrows my findings 1 and 4**, and both point at a stronger argument that neither of my comments made. Correcting on the record.

### 1. Retracted: my finding 3's `quantity: 0` example is wrong

I claimed a `PositionTrimmed` with `removals: [{ tier: "c3", quantity: 0 }]` clears every gate. It does not — `events/parse.ts:569` rejects it at the front door: *"Removal quantity must be a positive number."* And with `quantity > 0` on an unheld tier, the sufficiency gate at `crossref.ts:1245` computes `requested > available + 1e-9` → `q > 1e-9` → **reject**. My example is closed at two doors. I should have checked the removal parser before asserting reachability; apologies for the noise.

**What survives, and it is the important half:** the *mechanism* I named is real — `splitTierRemoval` returns `removed: []` whenever `tierTotal === 0` (`fold.ts:933`) — and it is reachable for callers that fold **without** the crossref gate. That is not hypothetical: `packages/event-store/src/event-store.ts:225` folds `load.events` straight from the durable log, and `apps/tui/src/event-store.ts:307` exists precisely for migrated events judged against `foldEvents(genesis, migrated)`. Events already durable from before a gate rule, or arriving by migration, reach `tierWeightedDeltas` with no sufficiency check between them and this arm.

### 2. Narrowed: a lot-less `PositionOpened` cannot be parsed at all

`events/parse.ts:526-528` rejects `lots: []` (*"Position requires at least one lot."*) and `parse.ts:349` rejects it for genesis positions. So the ghost-position and zero-price-anchor effects I described in **finding 1** are **fixture-only on the open arm** — real in `foldEvents`, unreachable through either door. Finding 1's substance holds for **close and trim**, which is where it matters; the open sub-case should be read as a fixture artefact, not a live defect.

This also sharpens **finding 4** into something worse than I filed. The gap is not merely that the test asserts too little — it is that `lotlessOpen` constructs **a shape the parser forbids**, so all four new discard tests pin the one route that cannot occur, while the only reachable producer (a gate-bypassed trim or close) is untested. That inverts the PR body's reachability narrative: it names `crossref.ts:1480` / `PositionOpened` as the live question, when parsing already forecloses it, and never examines the trim credit path, which reaches the arm with no sufficiency check at all.

### 3. The argument I missed: the debit/credit asymmetry flips the case on two arms

My review accepted "an invented attribution is worse than none" across all four routed sites. That is right for the **debits** (open funding, add funding): no lots means no provenance, and minting `c1` capital for money *leaving* is a fabrication. It **inverts** for the **credits** (close settlement, trim settlement). `settlement.proceeds` is money that actually arrived. Returning `[]` there does not decline to invent a balance — it **discards a real one**. The cash is dropped on the floor while the arm still books the consequences: `buildClosedPosition` records the full proceeds as realized gain against a zero cost basis, `positions.delete` still retires the position, and on a trim `markVsFill.deltaUsd` reports the entire proceeds as an off-mark deviation against a mark value of `0`.

So the fund's realized P&L rises by the proceeds while no reserve receives them. That is a worse failure than the invented `c1` this PR is fixing, and it is reachable on the one route the body does not analyse.

**This is my primary recommendation, replacing the ordering in my summary above:** treat the four routed sites as two kinds, not one. `provenance-absent` is the right answer for a lot-less **debit**. For a lot-less **credit**, dropping the delta is not conservative — the honest options are to credit the proceeds to `present[0]`/`c1` and record a *warning*, or to refuse the event whole (suppressing the closed-book row and the retirement too) rather than half-applying it. Either way it should not be the same reason code as the debit case, since the two now fail in opposite directions.

Findings 5–9 in my review stand as written; findings 2 and 7 are strengthened by the above.

packages/engine/src/events/fold.ts
AmetAlvirde: Accepted in full — this was the catch that changed the PR, and it was right.

Fixed in 9cd0481. `applyTieredLeg` now returns whether the rest of the event may run, and every lot-derived arm `break`s on `false`: no closed-book row, no retirement, no lot mutation, no registration. Your framing is the one that landed — a discarded *credit* is strictly worse than the invented `c1` it replaced, because the old row at least conserved the money.

The R8 comment in this arm already forbade the shape in its own words, which settled it: when the file documents the trap and the change walks into it, the file wins.

packages/engine/src/events/fold.ts
AmetAlvirde: Accepted — fixed in 9cd0481.

The discard is now total, and deliberately the same semantics as the `position-absent` branches already in these arms rather than a second kind of discard. The in-code invariant you quoted reads true again; the R8 comment is amended to say the delete sits inside the applying path, reached only once both gates pass.

Consequence accepted and stated in the body: a lot-less close leaves the position open rather than retiring it. That is the trade — loud and inspectable beats silent and balanced-looking.

packages/engine/src/events/fold.ts
AmetAlvirde: Accepted — fixed in 9cd0481. You and the other review converged on this arm independently, which is why it got the weight it did.

The trim arm now gates on the same boolean: no partial row, and `trimming.lots = working` never runs. Your point that the trim is the worse case holds — the position survives there, so the old shape left a live position carrying a partial row for proceeds that never landed.

packages/engine/src/events/fold.ts
AmetAlvirde: Accepted, and this one mattered more than its Low-ish surface suggests.

I refused to map this case onto `reserve-absent` precisely because that reason's prose would have asserted something false. Shipping a *new* member whose own prose was false on two of its four arms would have been the same error wearing the name I approved.

Not fixed by rewording: fixed by making the behaviour match the words. Under the total discard in 9cd0481, "No balance moved." is true on every arm that can record it. The sentence stays as written.

packages/engine/src/events/fold.ts
AmetAlvirde: Accepted — same resolution as the finding-2 thread. The prose is unchanged and is now true, because the behaviour moved to meet it rather than the sentence being softened to meet the behaviour. Fixed in 9cd0481.

packages/engine/src/events/fold.ts
AmetAlvirde: Accepted on the mechanism, with one correction you yourself posted later and I want recorded on the thread it belongs to.

The route is real: `splitTierRemoval` returns `removed: []` when `tierTotal === 0`. But it is **not** gate-admitted — `parse.ts:569` requires a positive removal quantity and `crossref.ts:1245` rejects a removal against an unheld tier on sufficiency. It is reachable only through callers that fold the durable log directly, bypassing both gates: `loadFoldedReview` at `event-store.ts:225` calls `foldEvents` with no parse and no crossref.

That makes it migration-shaped and historical rather than user-injectable, which is exactly why it ships as a recorded discard instead of deleted code. It is now the route the tests actually exercise, and the body says all of this.

packages/engine/src/events/fold.ts
AmetAlvirde: Accepted, and thank you for chasing this past my own correction.

I had already narrowed the framing once — from "both gate sites" to "only `:1480`" — and that was still wrong. Both sites run after `parseEvent`, and `parseLots` refuses an empty `position.lots`, so neither can ever receive an empty array. There is no guard that stopped guarding.

#367 was filed on the false premise and is **corrected in the open**: the retraction is a comment on the issue rather than a silent body edit, the title says the premise is retracted, and the body now asks the residual question instead — ingest-time validation does not constrain what the durable log already contains. The PR body records both wrong versions and the correction, because the conclusion moved twice.

packages/engine/src/fold.test.ts
AmetAlvirde: Accepted, and this is the pairing you called: finding 4 is *why* finding 1 was not noticed. A test that asserts only the reserve amount cannot see an arm that keeps going.

Fixed in 9cd0481. The four tests are replaced by six that assert conservation directly — no closed-book row, no position deleted, no lot mutation, realized P&L unmoved, reserve unchanged — and they are driven through `foldEvents` directly, the same door `loadFoldedReview` uses.

Proven rather than asserted: the mutation is the bug itself (record the skip, let the arm continue). It reds exactly three tests — the trim conservation test, the emptied-close test, and the lot-less-open test — while the precedence and prose tests stay green, since those only read the skip record. I re-ran that mutation myself rather than taking it on report.

packages/engine/src/fold.test.ts
AmetAlvirde: Accepted without reservation — this was a decorative guard, and this board has shipped two already.

All four pinned a lot-less *open*, which `parseLots` forbids, so they exercised a route that cannot occur while the reachable one went untested. Coverage in 9cd0481 is re-pointed at the trim route.

One parser-shape test survives, and it earns its place as a *tripwire* rather than a duplicate: it pins that `parseEvent` rejects `lots: []`, which is the very reason the open route is unreachable. If someone relaxes `parseLots`, that route goes live and the test reddens to say so.

packages/engine/src/fold.test.ts
AmetAlvirde: Acknowledged, and the claim-by-claim check is appreciated — particularly the confirmation that inverting the #299 pin was right rather than preserving it. A pin that disclaims endorsement is a record, not a contract; preserving it would have frozen the defect it existed to expose.

The order-independence test you rate strongest is unchanged. The rest of the describe has moved as per finding 4.

packages/engine/src/events/fold.ts
AmetAlvirde: Accepted and added in 8a8fe97 — you are right that this is the argument that makes the ruling unarguable, and it was the one I had left out.

The docstring now carries it: `dedupeFoldSkips` keys on `(eventId, reason)`, so recording both reasons would survive dedupe as two distinct records, inflating `HeadDigest.discardedEventCount` and emitting two lines from `formatFoldDiscards` for a single event. A measurable consequence beats a semantic preference, and precedence is a decision that deserves the concrete reason on the page.

packages/engine/src/events/fold.ts
AmetAlvirde: Accepted — documented in 9cd0481 rather than removed, and the docstring no longer implies the branch can fire.

The site stays routed with its return deliberately ignored, and the code now says why: `[event.lot]` is a one-element literal so the discard cannot occur, but the leg is lot-derived *by kind* — its tier comes from `event.lot.tier`. Moving it to the explicit-tier path would put a lot-derived leg among the explicit-tier ones and quietly lose the provenance check the day the site learns to add more than one lot. It is not wrapped in an `if`, because a branch that cannot be false is dead code pretending otherwise.

packages/engine/src/contracts.ts
AmetAlvirde: Accepted on the analysis, and the recommendation is now **moot in the best way** — the total discard removed the symptom rather than the body needing to describe it better.

With the arm discarded whole, a lot-less close pushes no row at all, so nothing lands in `rows`, `byTempo` or `totalRealizedPnlUsd`, `byTier` no longer under-sums, and every blotter identity still balances. Your point 1 — a row reporting realized profit no reserve ever received — was the sharper framing of the defect, and it is exactly what the fix eliminates. Point 2 on the trim arm is fixed by the same change.

The body's blotter section is rewritten accordingly: it now records that the earlier under-sum claim was an artifact of the partial discard and went away with it. Your finding-8 note about `discardedEventCount` being a surfaced change is carried there too.

PR #368, opened 2026-08-18
Make the discovery guard environment-independent: an authored fixture, not live repo state

### LINKED ISSUE #365: The discovery guard asserts against live repo state, so it fails in every worktree — and its nested-gitignore assertion is unexercisable there
`ops/testkit/gitignored-path-globs.test.ts` fails 2 of 8 in any linked worktree and passes 8/8 in the main checkout. It is the guard on test discovery — the one piece of machinery in the repo that fails toward green — and its verdict currently depends on which checkout it runs in.

```
 ❯ ops/testkit/gitignored-path-globs.test.ts (8 tests | 2 failed)
   × test discovery > still excludes agent worktrees under .claude/ — the bug this closed
     → expected false to be true // Object.is equality
   × test discovery > reaches nested .gitignore files, anchored to their own directory
     → expected [ …(6) ] to include 'apps/web/.tanstack/**'
```

## The production mechanism is correct; the test is over-specified

`gitignoredPathGlobs()` derives its globs from

```
git ls-files --others --ignored --exclude-standard --directory --no-empty-directory -z
```

which lists only ignored paths that **exist on disk**. A fresh linked worktree has no `.claude/`, no `apps/web/.tanstack/`, no `apps/web/.vercel/`, no `coverage/`, so those globs are never contributed — and correctly so. Nothing ignored on disk means nothing to exclude. Side by side, same commit:

```
$ git -C <worktree> ls-files --others --ignored --exclude-standard --directory --no-empty-directory
apps/price-feed/node_modules/   apps/tui/node_modules/   apps/web/node_modules/
node_modules/   packages/event-store/node_modules/   packages/preferences/node_modules/

$ git -C <main checkout> ls-files --others --ignored --exclude-standard --directory --no-empty-directory
.claude/   .vercel/   apps/web/.env   apps/web/.env.auth.local   apps/web/.env.push.local
apps/web/.tanstack/   apps/web/.vercel/   coverage/   node_modules/   … (+ the node_modules set)
```

The two failing assertions read `rootConfig.test.exclude` — i.e. live repository state — and assert that specific entries are present. They are assertions about the machine, not about the derivation.

## Why this is a regression and not a papercut

This is the exact defect class #355 was filed to retire: *a guard that gives different verdicts in different checkouts*. #362 fixed that for the repo-wide source walker and reintroduced it one layer up, inside the guard shipped to protect the fix.

The second-order consequence is the serious one. In a worktree the nested-`.gitignore` assertion is not merely failing — it is **unexercisable**. `apps/web/.tanstack/` cannot exist in a fresh worktree, so there is no state in which that assertion can pass there, and nothing verifies that the derivation reaches nested `.gitignore` files and anchors them to their own directory. This board does nearly all of its work in worktrees, so the guard cannot verify what it exists to verify exactly where it counts. The remaining green is the load-bearing "eats no test file" assertion, which is satisfied by an empty glob list — the failure mode the module's own header warns about.

## The fix

Test the derivation against an **authored fixture**, not the live repo: a temp directory with a known `.gitignore` layout — root entries, a nested `.gitignore` in a subdirectory, directories that exist and directories that do not — run the derivation against it, assert the output. That makes the nested-gitignore and `.claude/` behaviours exercisable in every checkout, and pins the "does not exist on disk, contributes nothing" behaviour as intent rather than as an accident of where the suite happened to run.

A live-repo smoke assertion may stay alongside the fixture-driven one only if it is written to be true in both a worktree and the main checkout.

Shipped in #362. Scope is `ops/testkit/` only.

AmetAlvirde:
Closes #365.

The guard on test discovery asserted against **live repository state**, so it gave a different verdict in every checkout. `gitignoredPathGlobs()` derives its globs from

```
git ls-files --others --ignored --exclude-standard --directory --no-empty-directory -z
```

which lists only ignored paths that **exist on disk**. Two assertions in `gitignored-path-globs.test.ts` read `rootConfig.test.exclude` and required specific entries — `apps/web/.tanstack/**`, `apps/web/.vercel/**`, and a `.claude/` match — to be among them. A fresh linked worktree has none of those directories, so those globs are never contributed, and the guard went red on correct code.

## The production mechanism was right; the test was wrong

Nothing ignored on disk means nothing to exclude. That is the correct answer, not a bug — and the derivation is otherwise sound: repo-root-anchored, nested-`.gitignore`-aware, escaping intact. The defect was entirely in what the guard chose to assert against.

That makes this the exact defect class **#355** was filed to retire — a guard whose verdict depends on which checkout it runs in. #362 fixed it for the repo-wide source walker and reintroduced it one layer up, inside the guard shipped to protect that fix.

## Three environments, three answers, one commit

| environment | ignored dirs on disk | verdict at `a5a0cad` |
|---|---|---|
| main checkout (local) | `.claude/`, `coverage/`, `.vercel/`, `apps/web/.tanstack/`, `apps/web/.vercel/`, `.env*` | 8/8 pass |
| linked worktree | `node_modules/` only (6 globs) | 2 fail |
| CI, fresh clone | 7 globs | 2 fail |

CI was already red on this before #362 merged, and has been red on `main` ever since — every open lane's PR check inherits it. On the runner, the only reason any `apps/web` ignored output exists at all is that `pnpm --filter @numisma/web build` runs immediately before `pnpm test`, so the verdict depended on **CI step ordering** as well as on the checkout. `.claude/` can never exist on a runner, so that assertion had no state in which it could pass there.

## The second-order problem: unexercisable, not merely failing

In a worktree `apps/web/.tanstack/` **cannot** exist, so the nested-`.gitignore` assertion had no passing state — nothing verified that the derivation reaches nested `.gitignore` files and anchors them to their own directory. This board does nearly all of its work in worktrees, so the guard could not verify what it exists to verify exactly where it counts. What stayed green was the load-bearing "eats no test file" assertion, which an empty glob list satisfies — the fail-toward-green mode the module's own header warns about.

## The fix: assert the derivation against an authored fixture

`gitignoredPathGlobs()` takes an optional derivation root (`gitignoredPathGlobs(root = REPO_ROOT)`); the default is unchanged, so `vitest.config.ts` is untouched. The git flags are unchanged — the mechanism was never the problem. The parameter is the seam that lets the guard drive the derivation against a repository it authored rather than the one it happens to be running in.

The tests build a hermetic fixture repo in a temp directory with a known layout: a root `.gitignore`, a **nested** `.gitignore` in a subdirectory, directories that exist and directories that do not, an empty ignored directory, and a literal `{tmp,scratch}` directory that pins escaping through a real git round-trip. The fixture sets its own `core.excludesFile`, so a developer's global excludes cannot change the answer.

This turns the two dead assertions into live ones and adds the coverage nothing had:

- nested `.gitignore` reached **and anchored to its own directory** — the assertion that was unexercisable in a worktree, now exercised in every checkout
- `.claude/` excluded, with a buried test path matching and a sibling path outside it not matching
- a rule for a directory that **does not exist on disk** contributes nothing — pinned as intent, since that correct behaviour is what produced the false red
- an ignored-but-empty directory contributes nothing (`--no-empty-directory`)
- escaping survives the real git round-trip, not just the pure function

The live-repo layer is now deliberately thin: the load-bearing "eats no test file git knows about" assertion stays, plus a smoke that holds in a worktree, the main checkout and CI alike. Nothing left in the suite asserts the presence of a directory that only some checkouts have.

## The guard is not decorative

A guard that cannot fail is what produced this, so the replacement was mutation-tested against the glob derivation. Each mutation was applied to `gitignored-path-globs.ts`, observed red, and restored (the restored file is byte-identical to the reviewed one, `shasum c006e540…`):

| mutation | red |
|---|---|
| ignore the injected root — `cwd: root` → `cwd: REPO_ROOT`, i.e. reintroduce the exact defect this PR fixes | 4 failed / 14 |
| drop `--directory --no-empty-directory` | 6 failed / 14 |
| stop escaping metacharacters — `entry.replace(GLOB_METACHARACTERS, …)` → `entry` | 5 failed / 14 |
| unanchor — emit `**/<last segment>/**` instead of the real path | 8 failed / 14 |

The first is the one that matters: reintroducing the environment dependence turns the fixture suite red, which is precisely what the old guard could not do.

## Verification

| run | result |
|---|---|
| this test, linked worktree (bare) | 14/14 |
| this test, worktree with the main checkout's ignored dirs reconstructed on disk | 14/14 |
| **control** — pre-fix code, same reconstructed shape | 8/8, matching the main checkout exactly, which confirms the reconstruction is faithful and the only variable is on-disk state |
| full suite, worktree | 153 files passed, 4 skipped; 2287 tests passed, 48 skipped; 0 failures |
| `pnpm typecheck` (six projects + `tsconfig.ops.json`) | clean |

The control run is the load-bearing one: the same commit that is 8/8 in a main-shaped tree is 2-failed in a bare worktree, and after this change it is 14/14 in both.

### DIFF
--- a/ops/testkit/gitignored-path-globs.test.ts
+++ b/ops/testkit/gitignored-path-globs.test.ts
@@ -41,7 +41,13 @@
  * that hold in BOTH the main checkout and a fresh worktree.
  */
 import { execFileSync } from "node:child_process";
-import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import {
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 
@@ -55,7 +61,7 @@ import {
   globForIgnoredEntry,
   globsForIgnoredEntries,
 } from "./gitignored-path-globs.ts";
-import { repoFiles } from "./repo-sources.testkit.js";
+import { REPO_ROOT, repoFiles } from "./repo-sources.testkit.js";
 
 /** Whether `path` (repo-root-relative, POSIX) is matched by any of `globs`. */
 function matchesAny(globs: readonly string[], path: string): boolean {
@@ -95,18 +101,63 @@ describe("test discovery", () => {
    * `gitignoredPathGlobs()` returning `[]` would make every other assertion here
    * pass while the original bug came straight back.
    *
-   * WHY THIS LAYER IS ONLY TWO ASSERTIONS WIDE. Anything more specific about the
-   * live repo is a claim about WHICH CHECKOUT the suite is running in.
-   * `node_modules/` is the one ignored directory that exists in the main checkout
-   * and in every linked worktree alike, so it is the only named path this layer
-   * may mention. Everything about the derivation's behaviour — nested
-   * `.gitignore` files, anchoring, escaping, empty directories, and the
+   * WHY NON-EMPTINESS IS THE STRONGEST CLAIM AVAILABLE HERE, AND WHY NO PATH IS
+   * NAMED. An earlier version also asserted `toContain("node_modules/**")`, which
+   * is the exact class this module retires: a claim about WHICH CHECKOUT the suite
+   * is running in. A pnpm store-linked or symlinked `node_modules`, a CI cache
+   * restored outside the tree, or a `--modules-dir` install emits no such entry,
+   * and the guard goes red on completely correct code. `> 0` has none of that
+   * fragility and is still not vacuous: running vitest at all requires an install,
+   * an install puts at least one ignored directory somewhere on disk, and git
+   * reports it — so this holds in the main checkout, in a fresh linked worktree,
+   * and on a cold CI runner alike.
+   *
+   * BUT IT IS NOT THE SAME STRENGTH AS THE TRACKED-TEXT GUARD BELOW, and the
+   * difference is worth being honest about. `.gitignore`'s bytes are identical in
+   * every checkout of a commit, so reading them is environment-independent BY
+   * CONSTRUCTION. `> 0` is an empirical claim about how installs behave — very
+   * robust, not a tautology, and still the same KIND of claim as the
+   * `node_modules/**` one just deleted, only far harder to falsify. It is kept
+   * because the no-op direction genuinely needs a pin and nothing stronger is
+   * available at this layer; if it ever goes red on correct code, delete it rather
+   * than weaken it, because the fixture below is what actually guards the
+   * behaviour. Everything about the derivation's BEHAVIOUR —
+   * nested `.gitignore` files, anchoring, escaping, empty directories, and the
    * deliberate silence about ignored paths that are not on disk — is asserted
    * against the authored fixture repo below, where it holds on every machine.
    */
   it("contributes real globs in any checkout", () => {
     expect(contributedGlobs.length).toBeGreaterThan(0);
-    expect(contributedGlobs).toContain("node_modules/**");
+  });
+
+  /**
+   * THE ONE GUARD LEFT ON THE REAL REPO'S OWN IGNORE RULES, and it is not a
+   * relapse into the live-state class the rest of this file retires. `.gitignore`
+   * is a TRACKED file: at a given commit its bytes are identical in the main
+   * checkout, in every linked worktree, and on every CI runner. Reading its TEXT
+   * is therefore environment-independent by construction — unlike asking git what
+   * it currently ignores ON DISK, which is a fact about one working tree.
+   *
+   * WHAT BREAKS IF THIS LINE GOES AWAY. #362: agent worktrees live under
+   * `.claude/worktrees/`, so without `.claude/` in the root `.gitignore` every
+   * worktree's copy of the tree is collected a second time — the same test files
+   * discovered twice, from two paths. Moving the `.claude/` assertion into the
+   * fixture repo (where it belongs, because that is where it is exercisable in
+   * every checkout) proved the DERIVATION handles the rule, but left nothing at
+   * all watching whether this repo still declares it. Delete the line and the
+   * fixture stays green, the whole suite stays green, and #362 is simply back —
+   * green while running the wrong set of tests, which is the fail-toward-green
+   * mode this entire module exists to prevent.
+   */
+  it("still declares .claude/ in the repo's own tracked .gitignore", () => {
+    // Tracked, not merely present: if `.gitignore` ever stopped being a file git
+    // knows about, reading it off disk would be reading an untracked local file
+    // and the claim would quietly become checkout-specific again.
+    expect(repoFiles()).toContain(".gitignore");
+    const lines = readFileSync(join(REPO_ROOT, ".gitignore"), "utf8")
+      .split("\n")
+      .map((line) => line.trim());
+    expect(lines).toContain(".claude/");
   });
 });
 
@@ -118,17 +169,70 @@ describe("test discovery", () => {
  * (the nested `.gitignore`, in a worktree) or was exercised only by accident of
  * what happened to be sitting on disk.
  *
- * HERMETIC ON PURPOSE. `--exclude-standard` reads `.git/info/exclude` and the
- * user's GLOBAL `core.excludesFile` as well as the repo's `.gitignore` files, so
- * a fixture that did not neutralise both would assert something different on a
- * machine whose owner globally ignores, say, `coverage/`. The repo-local
- * `core.excludesFile` override points at an empty file inside `.git/`, and
- * `.git/info/exclude` is truncated. No commit is made — `--others --ignored`
- * reads the working tree — but the non-ignored files ARE staged, for the reason
- * recorded at the `git add -A` below.
+ * HERMETIC ON PURPOSE, ON TWO AXES. `--exclude-standard` reads `.git/info/exclude`
+ * and the user's GLOBAL `core.excludesFile` as well as the repo's `.gitignore`
+ * files, so a fixture that did not neutralise them would assert something
+ * different on a machine whose owner globally ignores, say, `coverage/`. That is
+ * the CONFIG axis: the repo-local `core.excludesFile` override points at an empty
+ * file inside `.git/`, and `.git/info/exclude` is truncated.
+ *
+ * The second axis is the ENVIRONMENT, and it is the dangerous one. Every git
+ * process here — the fixture's own `git()` helper and the `gitignoredPathGlobs()`
+ * call under test — would otherwise inherit `process.env`. `GIT_DIR`,
+ * `GIT_WORK_TREE` and `GIT_INDEX_FILE` outrank both `-C` and `cwd`, so under
+ * `git bisect run pnpm test`, or inside a hook, or in any wrapper that exports
+ * them, this fixture's `git add -A` would stage the developer's REAL working tree.
+ * `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_n`/`GIT_CONFIG_VALUE_n` sit ABOVE repo-local
+ * config in git's precedence order, so they beat the `core.excludesFile` override
+ * the config axis relies on, and `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` can point
+ * git's config lookup anywhere. So both git surfaces run under `hermeticGitEnv()`,
+ * an ALLOWLIST rather than a denylist of variables to drop — see the note there.
+ *
+ * No commit is made — `--others --ignored` reads the working tree — but the
+ * non-ignored files ARE staged, for the reason recorded at the `git add -A` below.
  */
+/**
+ * The variables a git subprocess in this file is allowed to see, and NOTHING else.
+ *
+ * AN ALLOWLIST, DELIBERATELY — the same argument the module under test makes about
+ * escaping instead of detecting. A denylist of hostile variables (`GIT_DIR`,
+ * `GIT_WORK_TREE`, `GIT_INDEX_FILE`, `GIT_CONFIG_*`, …) would be correct only for
+ * as long as the list is complete, and git's environment surface is large and
+ * growing: `GIT_COMMON_DIR`, `GIT_OBJECT_DIRECTORY`, `GIT_NAMESPACE`,
+ * `GIT_CEILING_DIRECTORIES`, `GIT_ATTR_NOSYSTEM` and more all steer discovery or
+ * config. Naming what may pass THROUGH closes the whole class at once, including
+ * variables git has not shipped yet, and its failure mode is a loud broken
+ * subprocess rather than a quietly redirected one.
+ *
+ * `HOME` is pointed inside the fixture and `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM`
+ * at `/dev/null`, which is belt AND braces with the repo-local `core.excludesFile`
+ * override: the override alone is enough, but only while nothing above it in the
+ * precedence order is set, and this env is exactly what guarantees that.
+ */
+const GIT_ENV_PASSTHROUGH = [
+  "PATH",
+  // Windows needs these three to spawn a process at all; absent elsewhere.
+  "SystemRoot",
+  "SYSTEMROOT",
+  "COMSPEC",
+] as const;
+
+function hermeticGitEnv(home: string): NodeJS.ProcessEnv {
+  const env: NodeJS.ProcessEnv = {
+    HOME: home,
+    GIT_CONFIG_GLOBAL: "/dev/null",
+    GIT_CONFIG_SYSTEM: "/dev/null",
+  };
+  for (const name of GIT_ENV_PASSTHROUGH) {
+    const value = process.env[name];
+    if (value !== undefined) env[name] = value;
+  }
+  return env;
+}
+
 describe("gitignoredPathGlobs against an authored fixture repo", () => {
   let fixtureRoot: string;
+  let fixtureEnv: NodeJS.ProcessEnv;
   let globs: string[];
 
   /** Write one fixture file, creating its parent directories. */
@@ -140,18 +244,30 @@ describe("gitignoredPathGlobs against an authored fixture repo", () => {
 
   /** Run one git command inside the fixture, output discarded. */
   function git(...args: string[]): void {
-    execFileSync("git", ["-C", fixtureRoot, ...args], { stdio: "ignore" });
+    execFileSync("git", ["-C", fixtureRoot, ...args], {
+      stdio: "ignore",
+      env: fixtureEnv,
+    });
   }
 
   beforeAll(() => {
     fixtureRoot = mkdtempSync(join(tmpdir(), "gitignored-path-globs-"));
+    const home = join(fixtureRoot, ".git-home");
+    mkdirSync(home, { recursive: true });
+    fixtureEnv = hermeticGitEnv(home);
     git("init");
-    // Neutralise the ambient environment: an empty global excludes file, and an
-    // empty per-repo exclude. Both live inside `.git/`, so neither is itself a
-    // path `ls-files` can report.
+    // Neutralise the ambient CONFIG: an empty global excludes file, and an empty
+    // per-repo exclude. Both live inside `.git/`, so neither is itself a path
+    // `ls-files` can report. (The ambient ENVIRONMENT is neutralised by
+    // `fixtureEnv` above, which every git process here runs under.)
     writeFileSync(join(fixtureRoot, ".git", "empty-excludes"), "", "utf8");
     git("config", "core.excludesFile", join(fixtureRoot, ".git", "empty-excludes"));
-    writeFileSync(join(fixtureRoot, ".git", "info", "exclude"), "", "utf8");
+    // `init.templateDir` can produce a `.git/` with no `info/` directory at all,
+    // in which case writing straight to `info/exclude` dies ENOENT and errors
+    // every test in this describe.
+    const excludeFile = join(fixtureRoot, ".git", "info", "exclude");
+    mkdirSync(dirname(excludeFile), { recursive: true });
+    writeFileSync(excludeFile, "", "utf8");
 
     // ---- The authored layout. ------------------------------------------
     // Root ignores: two directories that exist and are wholly ignored, one that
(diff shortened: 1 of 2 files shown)


ops/testkit/gitignored-path-globs.test.ts
```diff
@@ -73,20 +94,180 @@ describe("test discovery", () => {
...
+  /** Write one fixture file, creating its parent directories. */
+  function write(relative: string, contents: string): void {
+    const target = join(fixtureRoot, relative);
+    mkdirSync(dirname(target), { recursive: true });
+    writeFileSync(target, contents, "utf8");
+  }
+
+  /** Run one git command inside the fixture, output discarded. */
+  function git(...args: string[]): void {
+    execFileSync("git", ["-C", fixtureRoot, ...args], { stdio: "ignore" });
```
AmetAlvirde: **medium — the fixture is hermetic against git *config*, but not against the git *environment*.**

`git()` and `gitignoredPathGlobs(fixtureRoot)` both shell out with the inherited `process.env`. `--exclude-standard` config leaks are neutralised, but `GIT_DIR` / `GIT_WORK_TREE` / `GIT_INDEX_FILE` are not — and those override `-C` and `cwd` entirely.

Concrete scenario: the suite run from inside a git invocation that exports them — `git bisect run pnpm test`, `git rebase --exec "pnpm test"`, or any hook that runs tests. Verified locally with git 2.50.1:

```
$ GIT_DIR=$outer/.git GIT_WORK_TREE=$outer git -C $fix init -q
$ ls -a $fix          # empty — no .git was created in the fixture
$ GIT_DIR=$outer/.git GIT_WORK_TREE=$outer git -C $fix add -A
$ git -C $outer diff --cached --name-only
a.txt
b.txt                 # the REAL working tree got staged
```

So `git("init")` silently re-inits the ambient repo, `git("add", "-A")` stages the developer's entire working tree into their real index, and `gitignoredPathGlobs(fixtureRoot)` reads the ambient repo — the exact-set assertion then fails on correct code, with a side effect on the user's index.

The repo already has the pattern for this: `gitEnv()` in `apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts:156` builds a *replacement* env (`GIT_CONFIG_GLOBAL=/dev/null`, `GIT_CONFIG_SYSTEM=/dev/null`, fixed identity) rather than inheriting one. Passing an explicit `env` here — at minimum with `GIT_DIR`, `GIT_WORK_TREE` and `GIT_INDEX_FILE` stripped — would close it, and `gitignoredPathGlobs` needs the same treatment since the test cannot scrub the env of a call it does not make.

ops/testkit/gitignored-path-globs.test.ts
```diff
@@ -73,20 +94,180 @@ describe("test discovery", () => {
...
+   * and in every linked worktree alike, so it is the only named path this layer
+   * may mention. Everything about the derivation's behaviour — nested
+   * `.gitignore` files, anchoring, escaping, empty directories, and the
+   * deliberate silence about ignored paths that are not on disk — is asserted
+   * against the authored fixture repo below, where it holds on every machine.
    */
-  it("still excludes agent worktrees under .claude/ — the bug this closed", () => {
+  it("contributes real globs in any checkout", () => {
     expect(contributedGlobs.length).toBeGreaterThan(0);
+    expect(contributedGlobs).toContain("node_modules/**");
```
AmetAlvirde: **medium — this is still a live-state assertion, of the same class the PR retires.**

The comment above argues `node_modules/` is "the one ignored directory that exists in the main checkout and in every linked worktree alike". That is a claim about on-disk shape, and it has a failure mode: git only emits `node_modules/` (with the trailing slash that `globForIgnoredEntry` turns into `/**`) when `node_modules` is a real directory.

If `node_modules` is a **symlink** — a common way to share an install across linked worktrees, and what some CI caches restore — git does not treat it as a directory, so `node_modules/` in `.gitignore` does not match it and nothing is contributed at all. Verified with git 2.50.1:

```
$ ln -s /elsewhere/node_modules node_modules   # .gitignore has `node_modules/`
$ git add -A
$ git ls-files --others --ignored --exclude-standard --directory --no-empty-directory
                                               # (empty)
```

`toContain("node_modules/**")` then goes red on correct code in that checkout, which is the defect this PR exists to remove. The non-emptiness half (`length > 0`) is the checkout-independent claim and carries the "mechanism is not a no-op" job on its own; if a named path is wanted here, `git check-ignore` is the environment-independent way to ask (see the separate note on `.claude/`).

ops/testkit/gitignored-path-globs.test.ts
```diff
@@ -73,20 +94,180 @@ describe("test discovery", () => {
...
+
+  beforeAll(() => {
+    fixtureRoot = mkdtempSync(join(tmpdir(), "gitignored-path-globs-"));
+    git("init");
+    // Neutralise the ambient environment: an empty global excludes file, and an
+    // empty per-repo exclude. Both live inside `.git/`, so neither is itself a
+    // path `ls-files` can report.
+    writeFileSync(join(fixtureRoot, ".git", "empty-excludes"), "", "utf8");
+    git("config", "core.excludesFile", join(fixtureRoot, ".git", "empty-excludes"));
+    writeFileSync(join(fixtureRoot, ".git", "info", "exclude"), "", "utf8");
```
AmetAlvirde: **low — assumes `.git/info/` exists, which the default template guarantees and a custom one does not.**

`git init` only creates `.git/info/exclude` because the default template ships it. A developer (or a CI image) with `init.templateDir` set globally, or `GIT_TEMPLATE_DIR` exported, to a template without an `info/` directory gets `ENOENT` from this `writeFileSync`, `beforeAll` throws, and all seven fixture tests error out with a message that says nothing about excludes. One line fixes it:

```suggestion
    mkdirSync(join(fixtureRoot, ".git", "info"), { recursive: true });
    writeFileSync(join(fixtureRoot, ".git", "info", "exclude"), "", "utf8");
```

ops/testkit/gitignored-path-globs.test.ts
```diff
@@ -73,20 +94,180 @@ describe("test discovery", () => {
...
+    // an empty index the nested `pkg/web/.gitignore` would never be reached and
+    // the fixture would quietly under-report. `git add -A` honours `.gitignore`,
+    // so everything ignored above stays unstaged and therefore still "other".
+    git("add", "-A");
+
+    globs = gitignoredPathGlobs(fixtureRoot);
+  });
+
+  afterAll(() => {
+    rmSync(fixtureRoot, { recursive: true, force: true });
```
AmetAlvirde: **low — masks the real failure when the fixture never got created.**

If `mkdtempSync` throws (full `TMPDIR`, read-only temp on a locked-down runner), `fixtureRoot` is still `undefined` when `afterAll` runs, and `rmSync(undefined, …)` throws `ERR_INVALID_ARG_TYPE`. Vitest reports the teardown error, and the actual cause — the failed mkdtemp — is the one buried. Guard it:

```suggestion
  afterAll(() => {
    if (fixtureRoot !== undefined) {
      rmSync(fixtureRoot, { recursive: true, force: true });
    }
  });
```

ops/testkit/gitignored-path-globs.ts
```diff
@@ -90,11 +90,27 @@ export function globsForIgnoredEntries(
...
+ * `.gitignore` files and anchors them to their own directory — could never run
+ * where this board actually does its work. Point `root` at an AUTHORED fixture
+ * repo instead and the behaviour under test (nested ignores, anchoring, escaping
+ * through a real git round-trip, `--no-empty-directory`, and the deliberate
+ * silence about ignored paths that do not exist on disk) is pinned identically on
+ * every machine. The default is unchanged, so `vitest.config.ts` calls this
+ * exactly as it did before.
  */
-export function gitignoredPathGlobs(): string[] {
+export function gitignoredPathGlobs(root: string = REPO_ROOT): string[] {
```
AmetAlvirde: **low — `root` is now public API, and a non-toplevel `root` fails silently in the wrong direction.**

`git ls-files` reports paths **relative to the cwd** and only descends from it. So `gitignoredPathGlobs("/repo/apps/web")` returns `.tanstack/**`, not `apps/web/.tanstack/**` — globs that look plausible, are anchored to the wrong base, and (as `vitest.config.ts` matches against repo-root-relative paths) silently exclude nothing, or exclude the wrong subtree if a sibling name collides. That is the fail-toward-green direction this module's header warns about, and the header's promise that entries "are repo-root-relative" is now only true for the default.

The fixture happens to pass its own toplevel, so nothing catches a future caller that does not. Worth either documenting the precondition on the parameter, or enforcing it — e.g. deriving `git rev-parse --show-toplevel` under `root` and throwing when it disagrees.

ops/testkit/gitignored-path-globs.test.ts
```diff
@@ -73,20 +94,180 @@ describe("test discovery", () => {
...
+    expect(globs).toContain("pkg/web/.tanstack/**");
+    expect(globs).not.toContain("**/.tanstack/**");
+    expect(matchesAny(globs, "pkg/web/.tanstack/cache.json")).toBe(true);
+    // An identically named directory elsewhere is NOT ignored here, and the
+    // anchored glob must leave it collectable.
+    expect(matchesAny(globs, "pkg/api/.tanstack/x.test.ts")).toBe(false);
+  });
+
+  /** The bug this module closed, now exercisable in every checkout. */
+  it("still excludes agent worktrees under .claude/", () => {
```
AmetAlvirde: **low — moving this to the fixture drops the only guard on the repo's own `.gitignore`.**

Under the old version, deleting `.claude/` from the repo-root `.gitignore` turned this test red in the main checkout. Now the assertion is about the fixture's hand-written `.gitignore`, so it stays green no matter what the real one says — and the live layer only asserts non-emptiness plus `node_modules/**`. The #362 incident (a repo-root `vitest run` collecting 312 files across other lanes' worktrees) would come back silently, and the failure mode is a *passing* suite that ran the wrong tests.

The fixture correctly proves the mechanism; what is now unguarded is the repo's configuration. That claim can be made without depending on on-disk state, because `git check-ignore` consults the ignore rules rather than the working tree:

```
git check-ignore -q --no-index .claude/worktrees/x/a.test.ts   # exit 0 in any checkout
```

Unlike `ls-files --others --ignored`, it answers for paths that do not exist, so it holds identically in a bare worktree, the main checkout and CI.

AmetAlvirde:
## Review addendum — the things I checked that turned out to be non-issues

Six inline comments are on the diff already (three at `gitignored-path-globs.test.ts:109/143/154`, `:189`, `:219`, and one at `gitignored-path-globs.ts:113`). This comment records the rest of the hermeticity surface I probed, so nobody re-litigates it, plus one detail that sharpens the env-inheritance finding.

**Verified green.** I ran the suite in a *linked worktree* (`.claude/worktrees/lane-guard`) — 14/14 pass. That is the checkout where the old file failed 2-of-8, so the PR's central claim is demonstrated, not just argued.

### Cleared after checking — no change needed

- **`GIT_CONFIG_GLOBAL` / a system `/etc/gitconfig` setting `core.excludesFile`.** Neutralised. The fixture's `git config core.excludesFile …` writes to `.git/config`, and repo-local config beats both global and system. Probed directly: with `GIT_CONFIG_GLOBAL` pointing at a config that ignores `extra/`, `ls-files` still reports nothing. The comment at the top of the `describe` is accurate on this point.
- **`core.autocrlf`.** Affects blob content on checkout/stage, never the path list `--others --ignored` emits. Cannot move the asserted set.
- **`init.templateDir` installing hooks.** `git init` + `git add -A` runs no hook (`pre-commit` needs a commit). Only the *missing `info/` directory* case bites, which is the `:154` comment.
- **A `$TMPDIR` that is itself inside a git repository.** Not a hazard: `git init` makes the fixture its own toplevel and discovery stops there. Probed with a parent repo whose `.gitignore` said `pkg/` and `*.test.ts` — the fixture still reported exactly `pkg/web/.tanstack/`, parent rules ignored.
- **Ordering, on any platform.** Both sides of `expect([...globs].sort()).toEqual([…].sort())` are sorted, so git's emission order is irrelevant. Git emits `/` separators on every platform, and the fixture builds its paths with `join()` only for the *root*; the relative globs never see a backslash.
- **macOS `mkdtemp` and `/var` → `/private/var`.** Irrelevant. `git ls-files` reports paths *relative to cwd*, so the absolute prefix — symlinked or resolved — never reaches an assertion.
- **Case-insensitive APFS.** `git init` sets `core.ignorecase=true` there; confirmed the emitted set is byte-identical. No two fixture paths differ only by case, so there is nothing for it to fold.
- **`git add -A` — reasoning, comment, and side effects.** All three check out. I rebuilt the layout by hand: before the `add`, git emits `.claude/`, `node_modules/`, `{tmp,scratch}/` — `pkg/web/.tanstack/` is **missing**, exactly as the comment says, because `--directory` collapses the wholly-unknown `pkg/` and never descends to read the nested `.gitignore`. After the `add`, all four appear. So the staging is load-bearing and the comment states the right reason. It does not distort `--others --ignored` either: `add -A` honours `.gitignore`, so no ignored path becomes tracked and drops out of the listing.
- **Production change.** It is exactly `cwd: REPO_ROOT` → `cwd: root` with `root: string = REPO_ROOT`; the git flags are byte-identical. `gitignoredPathGlobs` has exactly one non-test caller, `vitest.config.ts:46`, still `gitignoredPathGlobs()` with no argument. Real-config behaviour is untouched.

### One addition to the env-inheritance finding (`:143`)

Beyond `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE`, the repo-local `core.excludesFile` override is **also** beatable by `GIT_CONFIG_COUNT` + `GIT_CONFIG_KEY_n`/`GIT_CONFIG_VALUE_n`, which sit *above* repo config in git's precedence order. Probed on git 2.50.1: injecting `core.excludesFile` that way made `extra/` appear in a fixture whose `.git/config` pointed at an empty excludes file. Exotic in a human shell, less so in a CI runner or a wrapper that templates git config through the environment. It does not need its own fix — the scrubbed-env approach already suggested at `:143` (mirroring `gitEnv()` in `apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts:156`) closes this too, provided the allowlist drops `GIT_CONFIG_*` along with `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE`.

### Verdict

**Non-blocking.** The design is right and the central claim is demonstrated in the checkout that used to fail. Nothing here is a correctness bug in the derivation itself; the findings are about the fixture's blast radius (env inheritance) and two remaining live-state assertions (`node_modules/**`, and the now-unguarded repo `.gitignore`). Worth addressing before merge, but none of it makes the guard wrong today.

ops/testkit/gitignored-path-globs.test.ts
AmetAlvirde: Fixed, and this was the finding I would have blocked on regardless of its severity label — it is not only a hermeticity bug, it is a test that could in principle write outside itself.

`gitignoredPathGlobs` now takes an `options.env` seam, spread in conditionally so the no-argument call from `vitest.config.ts` passes a byte-identical options object to `execFileSync`. The fixture runs both its own `git()` helper and the derivation under that environment.

One deliberate deviation from the suggestion: the scrub is an **allowlist**, not a deletion of named `GIT_*` variables. A denylist is correct only while it is complete, and git's environment surface includes `GIT_COMMON_DIR`, `GIT_OBJECT_DIRECTORY`, `GIT_NAMESPACE`, `GIT_CEILING_DIRECTORIES` and more, with new ones shipping — that is the same shape as the metacharacter denylist this module was written to retire. `gitEnv()` in the price-feed harness is itself an allowlist; it was read for the approach and re-derived locally rather than imported, since a cross-package test-helper import would be worse than a local one.

**Verified against a real decoy repository**, with `GIT_DIR`/`GIT_WORK_TREE` pointed at it: the fixture tests pass and the decoy's index is still empty afterwards — nothing was staged outside the fixture.

**Two corrections to the record, in the interest of not overclaiming:**

1. The "`git add -A` stages the developer's real tree" scenario is currently **unreachable**, not merely unlikely. Under a hostile `GIT_DIR`, `git init` never creates `fixtureRoot/.git`, so the first `writeFileSync` into it ENOENTs and `beforeAll` dies long before the `add`. I removed the scrub and confirmed this: 8 fixture tests skipped, decoy index untouched. The hazard was real in principle and prevented by luck of write-ordering. The scrub makes it structurally impossible instead of accidentally impossible, which is the right reason to keep it.
2. This does **not** make the file hostile-env-proof end to end, and it cannot from inside its own scope. `REPO_ROOT` in `repo-sources.testkit.ts` is itself `git rev-parse --show-toplevel` with no `env`, so an ambient `GIT_DIR` relocates the whole live-repo layer; under the decoy, `eats no test file…` and the `.gitignore` guard both fail. That is arguably correct behaviour — lie to git about which checkout you are in and it describes the one you named — and it fails loudly rather than toward green. It is a shared file behind roughly ten guards, so widening a priority fix lane into it is the wrong trade tonight. Flagged to the board owner as a possible follow-up rather than silently fixed.

ops/testkit/gitignored-path-globs.test.ts
AmetAlvirde: Fixed, and taken further than suggested: the assertion is **deleted** rather than replaced with a different named path. Naming any path at this layer is the live-state class the PR retires wearing a different hat, so `contributedGlobs.length > 0` is all that remains as the not-a-no-op pin.

Your implied point about strength was right and the first version of my replacement comment papered over it, so the comment now says it outright: `.gitignore`'s bytes are identical in every checkout of a commit, so the tracked-text guard is environment-independent **by construction**; `> 0` is an **empirical** claim about how installs behave — very robust, not a tautology, and still the same *kind* of claim as the one deleted, only far harder to falsify. The comment now also says what to do if it ever goes red on correct code: delete it rather than weaken it, because the fixture is what actually guards the behaviour.

ops/testkit/gitignored-path-globs.test.ts
AmetAlvirde: Fixed — `mkdirSync(dirname(excludeFile), { recursive: true })` before the write, so a non-default `init.templateDir` without an `info/` directory no longer errors every fixture test at `beforeAll`.

ops/testkit/gitignored-path-globs.test.ts
AmetAlvirde: Fixed — the cleanup is guarded, so a failed `mkdtempSync` surfaces its own error instead of being masked by `ERR_INVALID_ARG_TYPE` from `rmSync`. Confirming your parenthetical too: the temp directory does not leak when `beforeAll` throws partway, since `afterAll` still runs.

ops/testkit/gitignored-path-globs.ts
AmetAlvirde: Fixed, and deliberately past the "document the precondition" suggestion. Prose is the wrong instrument for a fail-toward-green mode in the one module written to close fail-toward-green — a non-toplevel `root` produces globs that look right and exclude nothing, which is exactly the failure that hides itself.

`root` is now **refused loudly**: `git rev-parse --show-toplevel`, with the error naming the actual toplevel. Two details worth noting:

- It runs **only when `root !== REPO_ROOT`**. The default is itself the output of `rev-parse --show-toplevel`, so validating it would add a redundant subprocess to the config-load path that precedes every single test invocation.
- Both sides are compared through `realpathSync`, because macOS `mkdtemp` yields `/var/folders/…` while git answers `/private/var/folders/…`; a raw string comparison would reject every legitimate temp fixture.

A test pins the refusal against `pkg/web` inside the fixture. The header prose covers the precondition as well.

ops/testkit/gitignored-path-globs.test.ts
AmetAlvirde: Fixed, and this was the sharpest finding in the review — thank you. It is the one that would have let the PR make things quietly worse: moving the `.claude/` assertion into the fixture proved the *derivation* handles the rule, while leaving nothing at all watching whether **this repo still declares it**. Delete that line from the root `.gitignore` and the fixture stays green, the whole suite stays green, and #362's duplicate-collection bug is simply back — green while running the wrong set of tests.

The new guard reads `.gitignore` as **tracked text** (`repoFiles()` must contain it, then its lines must contain `.claude/`), not as disk state. That is the distinction that keeps it from being a relapse into the class this PR retires: committed bytes are identical in the main checkout, in every linked worktree and on every CI runner, whereas asking git what it currently ignores on disk is a fact about one working tree.

Verified to bite: removing `.claude/` from the root `.gitignore` turns it red with `expected [ '.agents/', '.codex/', …(7) ] to include '.claude/'`; restored, and green again.

PR #373, opened 2026-08-18
feat(price-feed): recover a lost day with prices:fetch --as-of

### LINKED ISSUE #356: Price feed: a fully-missed 18:00–23:00 window is unrecoverable, and reports success at every surface
## The class

A day whose **entire 18:00–23:00 scheduled window is missed** — laptop asleep,
packed, or powered off across all six fires — produces **no anchor for that
date, and cannot be recovered by the schedule**. It is not a job that failed.
It is a job that never ran, and every surface a human checks reports success.

This is the class that `#309` (closed), `#311` and `#312` all miss: each of them
assumes a job that **starts**.

## It has already happened, twice, and is dated to happen again

- **2026-08-14 (Fri)** and **2026-08-15 (Sat)** carry no anchor. Three
  independent confirmations:
  - `data/gap-report.json` lists both with `"reason": "no-anchor"`;
  - the 08-17 08:01 backfill jumps `45/46 asOf=2026-08-13` → `46/46
    asOf=2026-08-16`, straight over both;
  - `~/Library/Logs/numisma/` shows **zero in-window runs** on either date —
    only out-of-window `RunAtLoad`-on-wake catch-ups at `08-14 10:34` and
    `08-15 10:45`.
- 08-16 (Sun) ran its full six-fire window and **is** anchored.
- The exposure is **not random**. It is structurally Fri/Sat/Sun and travel
  days. 08-14/08-15 were a Tampico→Monterrey roadtrip.
- **The next one is already dated: Friday 2026-08-21** — flight out of Monterrey
  at 20:50, so the machine is packed across the whole window. Then Sat 08-22
  and Sun 08-23.

## Why the schedule cannot recover it

The authority is the plist header at
`~/Library/LaunchAgents/com.numisma.pricefeed.daily.plist`, and the obvious
story is wrong in both directions.

Per `man 5 launchd.plist`, launchd does **not** drop a slept-through interval —
it fires **one coalesced catch-up** on wake. What it cannot do is make that
catch-up useful: the catch-up lands under a **new `asOf`** (the CDMX calendar
date), and `isFreshBar` (`apps/price-feed/src/fetch-prices.ts:393-394`,
`success.observationDate === asOf`) correctly refuses it.

**A `Wake` key would not have saved 08-14.** The real split is:

| case | what launchd does | recoverable? |
|---|---|---|
| asleep | one coalesced catch-up on wake | **no** — lands under the new `asOf` |
| powered off / logged out | LaunchAgent never loaded; `RunAtLoad` covers it | **no** — same reason |

Neither recovers a day the machine missed entirely.

## The part that makes it a defect rather than a limitation

**The failure is invisible at every surface a human checks.**

`data/job-heartbeat.json` reads `"exitCode": 0, "lastStep": "complete"` while
two days are missing, and every catch-up run prints *"price-feed daily run
complete"*. The gap report names the hole **only after the fact**, and only if
someone opens it.

**A green heartbeat is not evidence of an anchored day**, and nothing currently
says so. This is the same false-clean shape the price-feed work has been
closing elsewhere.

## Also: there is no supported backfill path

`apps/price-feed/src/cli.ts` calls `runPriceFetch()` **with no arguments**. The
injectable clock that anchors `asOf` and the mark gate exists as a seam
(`apps/price-feed/src/fetch-prices.ts:129`, *"Injectable clock; defaults to `new
Date()`"*) but is **never exposed as a CLI flag**. The CLI's own error text
names the only current route: *"hand-author the mark through the inbox (the
permanent manual fallback)"* — 2 days × 9 instruments = **18 marks**.

Worth recording for whoever builds the flag: **a clock override alone is not
sufficient.** All three providers hardcode "newest bar" and none accepts a date:

- `apps/price-feed/src/binance-provider.ts:30` — `limit=2`, no `startTime`
- `apps/price-feed/src/twelvedata-provider.ts:95` — `outputsize=1`, no `date`
- `apps/price-feed/src/banxico-provider.ts:18` — the literal `/datos/oportuno`
  ("latest") endpoint

Moving `asOf` to 2026-08-14 while the providers still return the 08-16/08-17
bars would make `isFreshBar` reject **every instrument** — backfilling nothing
while exiting green, i.e. reproducing the very bug this issue describes. All
three APIs do support historical queries (Binance verified 2026-08-17: it still
serves both lost daily bars), so the recovery is feasible — it is just a
CLI + orchestrator + three-provider change, not a one-flag change.

## What this issue asks for

Not a fix for the laptop's availability — that is a separate, larger question.
This issue is for the **detection and recoverability** gap:

1. A **missed-window is surfaced as a failure**, not reported as success — the
   heartbeat must be able to say "no anchor for date D", and a green heartbeat
   must stop meaning "the day is anchored".
2. A **supported backfill path** for a named past date, so a known-missed day is
   a one-command recovery rather than 9 hand-authored marks per day.

## Context

Diagnosed in `2026-08-17-price-feed-lost-day-handoff.md`. The 2026-08-15 standup
diagnosed this exact mechanism and put "file the lost-day defect" at the top of
that day's plan; nothing was filed, and **08-15 was lost that same night, by the
same mechanism, hours after it was understood.**

Adjacent: `#309` (closed — a hung backfill ate the window; a job that *started*),
`#311`, `#312`.

### LINKED ISSUE #358: Adds a target-date window to all three price providers
**Slice 1 of 4 — the lost-day recovery increment (#356).** Spec: `2026-08-17-lost-day-recovery-spec` §4. Converged design: `2026-08-17-lost-day-recovery-seam-grill` D6, D9.

**Blocked by:** nothing. This is the foundation slice — slices 2–4 chain off it.

---

## Why

2026-08-14 and 2026-08-15 have no marks in the durable log, and there is no way to ask for them: every provider fetches *the newest bar* and nothing else. This slice gives all three the ability to fetch **one named past day**.

Two corrections to #356's framing that this work is built on:

- **The registry holds 13 instruments, not 9** — 4 crypto (Binance, *daily* cadence), 3 US equities and 6 MXN-derived rows (both Twelve Data, *weekdays* cadence). 2026-08-14 (Friday) owes all 13; 2026-08-15 (Saturday) owes only the 4 crypto. The hole is **~17 marks**.
- **A lost day is not an unrecoverable day.** The full provider ladder was measured first-hand: Twelve Data serves daily bars back to **1985** with **no plan gate at any depth**, Banxico SF43718 back to **1991**, Binance to pair inception. There is **no recovery horizon on any provider**, and nothing in this increment is a race.

## Naming

The provider parameter is **`targetDate`**, not `asOf`. `ProviderObservation` documents `observationDate` as *"the provider's own bar date … it is NOT the trading-day `asOf` (which is timezone-anchored in the engine)"*. The provider layer must not learn the word `asOf`; that the two coincide on the recovery path is the orchestrator's conclusion, not the provider's premise.

### 1. Twelve Data

`fetchTwelveDataDailyCloses(entries, options)` gains an optional `targetDate` on its options. When present, the request selects the bar by **date window** instead of by `outputsize`:

- `start_date=<targetDate>`, `end_date=<targetDate + 1 day>`.
- ⚠️ **The end is EXCLUSIVE.** `start_date === end_date` returns the no-data 400 **even when the bar exists**. This is the single most likely thing in the whole increment to get wrong, and it is a *silent* wrongness: that 400 looks exactly like "that day had no trading." Use the engine's exported `addDays(targetDate, 1)`; do not hand-roll date arithmetic.
- Everything else holds: one comma-separated batched request per chunk, the same per-minute credit pacing, per-symbol attribution preserved, and the batch-level `status:"error"` fan-out unchanged. A windowed response still carries the row under `values`; keep reading `values[0]`.
- With no `targetDate`, the request is **byte-identical to today's** (`outputsize=1`, no date params).

### 2. Banxico

`fetchBanxicoFix(options)` gains an optional `targetDate`. When present, the series path becomes `/datos/<targetDate>/<targetDate>` — **inclusive both ends** — instead of the literal `/datos/oportuno`.

⚠️ **There must be no multi-day Banxico path, ever.** `extractLatestDatum` takes the **last** element of `datos` — the *newest* entry — and the range endpoint returns **ascending**. A window wider than one day would silently return the wrong day's FIX for every day but the last, and the failure would surface as a plausible-looking number rather than an error. Pinned to `/<target>/<target>` the existing parser works verbatim and must not be touched.

On a date with no published FIX, Banxico omits `datos` entirely and the existing "no FIX observation in payload" refusal fires. That is correct: under slice 2's owed-set filter it is reachable only on a weekday holiday, where a loud refusal is what is wanted.

### 3. Binance

`fetchBinanceDailyClose(entry, options)` gains an optional `targetDate`. When present, the request uses `startTime`/`endTime` in **epoch milliseconds** bounding that UTC day, instead of `limit=2`.

⚠️ **The `>= 2 klines` guard must not fire on the date-pinned path, and the reason is not cosmetic.** That guard is not a payload sanity check — it is a **proxy for settlement**. `limit=2` returns `[completed D, running D+1]` and the code takes the first row precisely to avoid marking a still-running candle. A date-pinned window for one past UTC day returns **exactly one row**, and under slice 2's guard (`asOf` strictly earlier than the current trading day, in a UTC-6 timezone) that day's UTC candle is **complete by construction**. Left unchanged, this guard rejects every single crypto recovery.

So on the date-pinned path: expect one row, take it, and treat **zero** rows as the symbol-attributable failure, naming the date. With no `targetDate`, the `limit=2` request and its settled-candle rule are unchanged.

### 4. `fetchJson` preserves the provider's own error text

Today a non-ok HTTP status discards the response body and yields `reason: "HTTP 400 Bad Request"`. That reads as a *malformed request* — a code bug — when on this path it almost always means *that day had no trading*. The API's own sentence sits in the body one line before it is thrown away, e.g.:

```json
{"code":400,"message":"No data is available on the specified dates. Try setting different start/end dates.","status":"error","meta":{"symbol":"AAPL","interval":"1day"}}
```

The change is **additive** and lives in the shared `fetchJson`, so the daily path gains a better sentence and no new control flow:

- On a non-ok status, attempt to read the body. Prefer a string `message`, then a string `msg`, then a string `error` — the three providers disagree on the key. Failing that, a trimmed snippet of the raw text body.
- Compose `reason` as `HTTP <status> <statusText> — <provider text>` when text was recovered; keep the bare `HTTP <status> <statusText>` as the fallback.
- **Cap the appended text** (a couple of hundred characters is ample) so an HTML interstitial cannot flood a scheduled run's log.
- Reading the body must be guarded: a body read that throws, or a non-JSON body, must still produce the bare fallback. **A non-ok status must never become a throw** — the Result-not-throw contract is what keeps one bad Twelve Data symbol from aborting the whole run.
- The ok path, the timeout reason and the transport-error reason are **unchanged**.

**Without this, slice 3's exit contract is unimplementable** — "name the suspected cause in the message" cannot be done with `reason` fixed at `"HTTP 400 Bad Request"`.

## Acceptance criteria

- [ ] Each provider, given a `targetDate`, returns an observation whose `observationDate` equals that date, proven against an **authored** fixture response.
- [ ] Twelve Data's request carries `end_date` one day **after** `start_date`, with a test that fails if the two are equal.
- [ ] Banxico's request path is `/<target>/<target>`, with a test asserting the existing parser still reads the single datum unchanged. No multi-day range is accepted anywhere in this slice.
- [ ] Binance returns a valid observation from a **single-row** windowed payload, and a zero-row payload is a symbol-attributable failure naming the date.
- [ ] With no `targetDate`, every provider's request URL and behaviour are unchanged — pinned by a test, because this is the live 18:00 path.
- [ ] `fetchJson` appends the provider message on a non-ok status, falls back to `HTTP <status> <statusText>` when there is none, caps the appended length, and never throws on a non-ok status.
- [ ] `pnpm verify` green.

⚠️ **Every committed fixture must be authored, never seeded from a real run** — the live ledger and its outputs are private.

## Out of scope

Any change to how a bar becomes a mark, to the owed set, to the CLI, to the mark gate, or to `docs/price-feed-ops.md`. No release mechanics of any kind.

### LINKED ISSUE #359: Adds a validated `asOf` override to the price fetch run
**Slice 2 of 4 — the lost-day recovery increment (#356).** Spec: `2026-08-17-lost-day-recovery-spec` §5. Converged design: `2026-08-17-lost-day-recovery-seam-grill` D4, D5, D7.

**Blocked by #358** — this slice cannot honor an `asOf` the providers cannot accept.

---

## Why

One call should be able to produce the honest record of a recovered day — `observationDate: 2026-08-14, fetchedAt: 2026-08-17` — with the mark gate answered by *settlement* rather than by the wall clock, and with instruments that owed nothing never fetched at all.

### 1. An explicit `asOf`, decoupled from `now`

`RunOptions` gains an optional `asOf: string`. When absent, the run is **byte-identical to today's**: `asOf = tradingDayAsOf(now(), config.timeZone)`. When present, it replaces that derivation and **nothing else** — `now` stays the real clock, so `fetchedAt` records the instant the recovery actually happened.

⚠️ **Do not fabricate `now`.** It was weighed and rejected twice over:

- It moves `fetchedAt` as well as `asOf`, **forging provenance** — the durable log would assert the recovery never happened.
- It **steers no provider anyway.** Both providers use `now` only for the `fetchedAt` stamp and as a fallback for a malformed bar timestamp, so a forged clock would still return the newest bar and the freshness gate would reject all 13. It pays a real cost for nothing.

A recovered mark dated for the day it measures and stamped with the day it was fetched *is* what "measurement recovered late" means.

⚠️ **Rejected alternative, recorded so it is not re-proposed:** a separate `runPriceBackfill` entry point. Two paths would have to agree forever on the mark contract, the freshness rule and the inbox id scheme with nothing forcing them to — the exact divergence the venue calendar's `satisfies Record<PriceSource, VenueCadence>` latch exists to prevent.

### 2. The override is all-or-nothing and validated, never coerced

The module's stated invariant is *"one clock, so the stored quote and the queued mark cannot diverge on `asOf`"*. Today that divergence is impossible **by construction**; with an override it becomes merely incorrect. So the override carries a validation obligation, and **the validation lives in `runPriceFetch`, not in the CLI**, so a programmatic caller cannot bypass it:

1. **Well-formed:** the value must satisfy `isIsoCalendarDate` (exported from `@numisma/engine`) — shape **and round-trip**. A shape-only check accepts `"2026-02-30"`, which then silently becomes March 2 in every date computation downstream. **Never silently coerced.**
2. **Strictly in the past:** `asOf` must be strictly earlier than `tradingDayAsOf(now(), config.timeZone)`. This also disposes of a future date.
3. Both refusals throw a plain `Error` carrying an operator-readable sentence. Slice 3 renders it; this slice owns it.

### 3. An explicit `asOf` forces the mark gate open, guarded

When a valid `asOf` override is present, `markEmitted` is **`true`**, unconditionally — `isAtOrAfterMarkTime` is not consulted.

`isAtOrAfterMarkTime` exists to answer *"has today's bar settled yet?"*; 18:00 CDMX is when the close becomes trustworthy. For a past date that question is **vacuous** — the bar settled days ago. On the live path 18:00 *infers* settlement; on the recovery path settlement is in hand, so the proxy has nothing left to do.

⚠️ **Rule 2 above is what makes this safe, and it is load-bearing.** Without `asOf < today`, `--as-of=<today>` before 18:00 would bypass the gate and mark an **unsettled** bar — the one thing the gate legitimately prevents. It is also what guarantees Binance's target UTC day is complete (see #358). Refuse loudly, naming the alternative: *"recover today" is just "run the daily job."*

⚠️ **Rejected alternative, recorded so it is not re-proposed:** keep the gate and require the operator to run after 18:00. Its failure mode is *correct fetch, zero marks, exit 0* — **#356's exact shape, re-created inside the tool built to fix #356.**

### 4. Compute the owed set before fetching

When an `asOf` override is present, the run filters the registry through `owesMarkOn(source, asOf)` — already exported from `@numisma/engine`, derived from `lastExpectedMarkDate` rather than re-deciding weekends — **before constructing any request**, and reports what it filtered.

This **dissolves** the ambiguity rather than handling it:

- If you never ask Twelve Data for a Saturday, you never receive the ambiguous 400 for the Saturday reason.
- It protects batch attribution. A non-ok status is a **request-level** failure for Twelve Data, so **one dateless symbol collapses all nine into `failAll(...)`**. Filtering by cadence means the doomed request is never built.
- It makes the output honest: *"08-15: 4 owed, 4 marked"* rather than *"13 attempted, 9 absent."*

⚠️ **This filter applies ONLY when an `asOf` override is present. Ruled explicitly.** Applied unconditionally it would change the **live daily path**: today a Saturday run fetches all 9 Twelve Data symbols, stores 9 quotes carrying Friday's close under Saturday's `asOf`, and records 9 stale-mark skips. All of that stays exactly as it is on `main`. Every reason the filter exists is a recovery-path reason — the ambiguous 400 it dissolves is *created* by #358's date-pinned request; on the live path Twelve Data does not 400 on a Saturday, it returns Friday's bar, which is precisely why the stale-skip channel exists.

### 5. The widened result

`FetchRunResult` gains:

- `asOf: string` — the trading day this run marked against, whether derived or overridden. The caller currently has no way to name the run's date except by reading a quote.
- `notOwed: readonly { instrumentId: string; symbol: string; source: PriceSource }[]` — instruments filtered out by the owed-set filter. **Empty on the live path.**

`totalCount` becomes the count of instruments the run **attempted**: the owed set under an override, all 13 without one. On 2026-08-15 that makes the existing `storedCount/totalCount` line read `4/4`, which is the honesty the owed-set filter is for.

## Measured facts, so they are not re-derived

- **The price store cannot be clobbered by a back-dated run.** `upsertQuote` is upsert-**by-`asOf`**, latest-wins per date, re-sorted on write. Recovering 2026-08-14 on 2026-08-17 inserts one row in date order and leaves every other row untouched.
- **The inbox is already idempotent.** `emitMarksToInbox` merges by the deterministic id `pm-<instrumentId>-<asOf>`; a mark already queued is skipped and hand-authored records are preserved verbatim.
- **The FIX for 2026-08-14 exists and agrees with the equity leg** (17.0218), so both legs of the MXN derivation agree on that date — no provider split.

## Acceptance criteria

- [ ] With no `asOf`, every observable of a run is unchanged: same `asOf` derivation, same mark gate, same 13 attempted instruments, empty `notOwed`. Pinned by a test — this is the live 18:00 path.
- [ ] A valid past `asOf` produces marks whose `asOf` is the target date and whose `fetchedAt` is the real current instant. **Assert both in one test** — the decoupling is the decision, and a test that checks only `asOf` would pass on a forged clock.
- [ ] `asOf` equal to the current trading day is refused with a message naming the daily job as the alternative. A future `asOf` is refused.
- [ ] `"2026-02-30"` is refused, not coerced into March 2.
- [ ] A valid past `asOf` emits marks regardless of the wall-clock hour — proven with an injected clock set before the mark time.
- [ ] Under a Saturday `asOf`, no Twelve Data or Banxico request is constructed at all, `notOwed` names the 9, and `totalCount` is 4.
- [ ] The recovered quotes appear in the store in date order without disturbing existing rows, and re-running the same date adds no new inbox marks.
- [ ] `pnpm verify` green.

⚠️ **Every committed fixture must be authored, never seeded from a real run** — the live ledger and its outputs are private.

## Out of scope

Argument parsing, exit codes, console output, `docs/price-feed-ops.md`, and any change to mark construction, the MXN derivation or the freshness rule. **The owed-set filter must not reach the no-flag daily path.** No release mechanics of any kind.

### LINKED ISSUE #360: Adds the `--as-of` recovery flag and its exit contract
**Slice 3 of 4 — the lost-day recovery increment (#356).** Spec: `2026-08-17-lost-day-recovery-spec` §6. Converged design: `2026-08-17-lost-day-recovery-seam-grill` D6, D8, D10.

**Blocked by #359** — this slice cannot expose a flag #359 does not implement.

---

## Why

An operator should be able to type one date and get back a report they can act on, and **a recovery that recovered nothing must not exit 0.**

### 1. The first argument-parsing seam this command has ever had

The `prices:fetch` entry point parses **zero** arguments today — there is no `process.argv` anywhere in it. `--as-of` is its first flag, so this slice brings a parsing seam with it. Author it as an exported, unit-tested **pure function over an argv array**, inside the price-feed app.

- Accept **`--as-of=<date>`** and **`--as-of <date>`**.
- **Single date. No range.** Multi-day recovery is a shell loop — `for d in …; do pnpm prices:fetch --as-of=$d; done` — and the inbox's idempotency already makes repetition safe. The unit of correctness is the day everywhere: the freshness rule compares one string, the inbox id is `pm-<instrumentId>-<asOf>`, the venue calendar answers per day. A range is a loop wearing a flag, and it walks into the Banxico ordering trap recorded in #358.
- ⚠️ **Refuse unknown arguments loudly.** This is the command's first parser, so there is no precedent to inherit, and one specific typo is dangerous: `--asof=2026-08-14` silently ignored would run the **daily** job, report success, and look exactly like a completed recovery. That is a #356-shaped failure introduced by the fix for #356.
- **Do not import the TUI spine's `parseAsOfArg`.** An app may not import an app in this repo, and the two are not the same contract anyway — the spine's `--as-of` is a *fold window*, this one is a *fetch target*, and this one's validation is strictly stronger (round-trip, not shape-only). What is genuinely shared **is** shared: the validation comes from the engine's exported `isIsoCalendarDate`, which is a pure declaration rather than flag parsing. Only the argv walk is duplicated, and it is the cheap half.

Shape validation may live in the parser; **semantic validation stays in `runPriceFetch`** (#359), which is the boundary a programmatic caller also crosses.

### 2. Three states per instrument, and exit non-zero on the third

Every registry instrument lands in exactly one state for the run's date:

| state | meaning | source |
| --- | --- | --- |
| **not owed** | the venue owed no mark on that date | the run's `notOwed` — never attempted |
| **owed and marked** | a mark was constructed and queued | present in the run's marks |
| **owed and absent** | owed, but no mark exists | a fetch failure, a derivation failure, or a bar whose date did not match |

**Exit 1 if the third set is non-empty.** Print each absent instrument with the provider's own words (which #358 makes available), and **name the suspected cause in the message — never pretend the exit code can separate a market holiday from a provider failure.** Say so in the output, in as many words.

⚠️ **Under an explicit `asOf`, a stale-bar skip on an *owed* instrument is a failure.** On the live daily path that skip is INFO by design — an equity's newest bar on a Saturday is Friday's, and treating that as failure would fail the nightly automation. On the recovery path the request was **pinned to the target date**, so a bar that is not that date means the day did not come back. **This boundary is the whole slice: the live path's INFO classification must not change, and the recovery path's must.**

### 3. Why the gap report's literal-`0` exit contract does not bind here

⚠️ **State this in the shipped code's own words, not only in the tracker**, or the literal `0` will later be read as house style and someone will "fix" this exit code back to green.

#266 D8 is *warn only, no exit-code change*, with `exitCode` typed as the literal `0` so the rule is testable, and #266 D7 promotes it from preferable to mandatory. Its entire argument rests on one premise, stated verbatim in three modules: *"a lost day is reported; a broken job is failed, **because a permanently red job is one nobody reads**."*

That is an argument about **recurring unattended automation.** `pnpm gap-report` runs nightly from the wrapper, and having accepted ~10 holiday false positives a year, a non-zero exit there would fail the daily job ten times a year. **This command is operator-invoked, one date at a time, by someone who just typed the date.** It is not in that channel and cannot erode it.

The positive argument is the stronger one: **#356 exists because a total failure exited green.** A recovery tool that recovers nothing and exits 0 is that defect with a new name.

**Recorded trigger that would flip this:** if recovery is ever put *into* an unattended loop — a catch-up pass at the top of the daily job — it inherits the nightly channel and #266 D7's argument reaches it. Then, and only then, revisit.

### 4. The run writes stored quotes and inbox marks only

Three refusals, each with its own reason:

1. **No heartbeat.** `job-heartbeat.json` records whether *the daily job ran*. Stamping it for 2026-08-14 would assert an evening run that never happened — recovering the marks while erasing the evidence the evening was missed, which is the manufactured-clean surface #356 is about. Its **only** writer in the codebase is the daily wrapper's EXIT trap, so this reduces to one operational rule: **never route recovery through the wrapper.** That is right on independent grounds too — the wrapper derives its mark window from the live wall clock, which is meaningless for a past date.
2. **No spine.** `--as-of` is a flag on an existing command, not a new pipeline, and a flag that silently extends post-conditions is a trap. The spine's ±50% guard is the authoritative gate, and a mark recovered days late deserves the same review.
3. **No accumulus commit, no projection refresh.** Both are existing wrapper steps and both stay the operator's, in the runbook (slice 4).

**The existing fetch-time spine pre-check stays exactly as it is.** It runs, its rejections still force exit 1, and its baseline is unchanged.

⚠️ **Noted deliberately so nobody "fixes" it:** the pre-check builds its reference from genesis plus the **whole** durable log, so a back-dated mark is judged against the newest known close rather than the one immediately preceding its own date. That is **correct by construction** — the pre-check's contract is fidelity to what `pnpm spine` will actually do, and the spine folds the log then walks the inbox. Making the pre-check "smarter" here would make it wrong.

## Acceptance criteria

- [ ] `--as-of=<date>` and `--as-of <date>` both parse; a missing or malformed value fails loud; an unknown argument fails loud rather than being ignored.
- [ ] With no arguments, the command's output and exit code are **unchanged** — pinned by a test.
- [ ] A recovery in which every owed instrument marked exits **0** and reports owed, marked and not-owed counts with the date named.
- [ ] A recovery in which at least one owed instrument produced no mark exits **1**, names each absent instrument, and carries the provider's own message where one exists.
- [ ] A recovery in which an owed instrument returned a bar for the wrong date exits **1** — and the equivalent skip on a no-flag run still exits **0**.
- [ ] `--as-of=<today>` and `--as-of=<future date>` are refused as a single readable message and exit 1, **without a stack trace**.
- [ ] The run touches the price store and the inbox and nothing else: no heartbeat write, no spine invocation, no commit, no projection refresh.
- [ ] `pnpm verify` green.

⚠️ **Every committed fixture must be authored, never seeded from a real run** — the live ledger and its outputs are private.

## Out of scope

Any range flag, any dry-run flag, any change to the spine pre-check's baseline, any change to `pnpm gap-report`'s exit contract, and `docs/price-feed-ops.md` (slice 4). Tightening the TUI spine's own weaker `--as-of` validator is a separate command with its own pinned messages and is **not** in this increment. No release mechanics of any kind.

### LINKED ISSUE #361: Documents the lost-day recovery procedure
**Slice 4 of 4 — the lost-day recovery increment (#356).** Spec: `2026-08-17-lost-day-recovery-spec` §7. Converged design: `2026-08-17-lost-day-recovery-seam-grill` D11.

**Blocked by #360** — the runbook documents a flag #360 introduces.

---

## Why

A future operator should be able to recover a lost day from the docs alone, and the word **"backfill" must stop meaning two things.**

## What to build

A **lost-day recovery** section in `docs/price-feed-ops.md`.

- [ ] The procedure is four steps, and **every one of them is a step already run today**: `pnpm prices:fetch --as-of=<date>` → inspect the output → `pnpm spine` → commit to accumulus.
- [ ] Show the multi-day form as the shell loop it is — `for d in …; do pnpm prices:fetch --as-of=$d; done` — and say why there is no range flag: the unit of correctness is the day everywhere, and the inbox's `pm-<instrumentId>-<asOf>` idempotency already makes repetition safe.
- [ ] **Show what a correct 4-of-4 Saturday recovery looks like beside a correct 13-of-13 weekday one**, so a future operator does not read the Saturday as a partial failure. The registry holds 13 instruments — 4 crypto on a *daily* venue, 9 on a *weekdays* venue — so a Saturday genuinely owes 4 and a 4/4 report there is **success, not a partial**.
- [ ] State the exit contract, and in one sentence that **the exit code cannot distinguish a market holiday from a provider failure — read the message.**
- [ ] State the containment refusals in the doc, not only in code: recovery writes **stored quotes and inbox marks only** — no heartbeat, no spine invocation, no accumulus commit, no projection refresh — and it is **never** run through the daily wrapper (`ops/price-feed/run-daily-fetch.sh`), which is the sole writer of `job-heartbeat.json` and which derives its mark window from the live wall clock.

### Vocabulary

- [ ] **`pnpm backfill` is already taken.** It is bound to the hosted Neon projection refresh and is an existing step of the daily wrapper. **This increment is "lost-day recovery."**
- [ ] Note in the doc that where #356 and the older handoffs say *"backfill,"* the term is **inherited and is being retired for this meaning**. `--as-of` is a good flag name precisely because it names the *parameter*, not the operation.

### Coordination

- [ ] ⚠️ **#357's runbook slice also writes `docs/price-feed-ops.md`** (the shell-profile snippet and the notice procedure). **This PR merges first; #357 inherits the merge.** The two additions are different sections and a conflict here is mechanical.

## Acceptance criteria

- [ ] `docs/price-feed-ops.md` carries a lost-day recovery section whose four steps can be followed without reading any tracker issue, grill or spec.
- [ ] The doc shows both a full weekday recovery and a crypto-only weekend recovery, and states explicitly that the weekend result is correct absence rather than a shortfall.
- [ ] The doc states the exit contract and the holiday-vs-failure ambiguity in the operator's own terms.
- [ ] The doc states the three containment refusals and the "never through the wrapper" rule.
- [ ] The doc records that `pnpm backfill` means the projection refresh and that "backfill" is retired for the recovery sense.
- [ ] `pnpm verify` green.

## Out of scope

The shell-profile snippet and the notice runbook (#357's). Any code change. No release mechanics of any kind — no manifest bump, no tag, no CHANGELOG.

## After this PR merges — operational, not a slice, and a human's to run

Not part of this issue and not part of the PR, recorded here because it is the point of the whole increment: those ~17 marks are missing from the durable log right now.

1. Recover **2026-08-14** — all **13** instruments, no provider split expected (Twelve Data served all 9 of its symbols and Banxico served the 14/08 FIX at 17.0218, so both legs of the MXN derivation agree on the date).
2. Recover **2026-08-15** — **4** instruments. A run reporting 4/4 there is success.
3. `pnpm spine`, inspect, commit to accumulus.
4. The gap report is a pure function of the log, so it stops naming those days on its own once the marks are appended. No separate cleanup.

AmetAlvirde:
Closes #356. It lands as four slices: closes #358, closes #359, closes #360 and closes #361.

**The defect.** 2026-08-14 and 2026-08-15 have no marks in the durable log —
the daily job did not run and every health surface stayed green, because a
lost day is only visible to someone who goes looking. The command that would
fix it did not exist: `pnpm prices:fetch` had exactly one behaviour, fetch
today.

**Two of #356's own claims are stale and this PR does not inherit them.** The
registry holds 13 instruments, not 9 — 4 crypto on Binance (daily cadence)
and 9 Twelve Data symbols (weekdays) — so the recovery is asymmetric: Friday
08-14 owes all 13 and Saturday 08-15 owes only the 4 crypto, about 17 marks
rather than 18. And a lost day is not unrecoverable: Twelve Data serves daily
bars to 1985, Banxico to 1991, Binance to pair inception. There is no
recovery horizon on any provider, so nothing here was a race.

**The four slices**, one commit each: providers accept a target date
(#358); a validated `asOf` override on the run (#359); the `--as-of` flag
and its exit contract (#360); the runbook (#361).

**Four silent traps, handled deliberately** — each is a decision, not an
accident:

1. Twelve Data's `end_date` is exclusive, so `start_date === end_date`
   returns the no-data 400 even when the bar exists. The window uses the
   engine's `addDays`, and a test fails if the two are ever equal.
2. Banxico is pinned to `/<target>/<target>` and there is deliberately no
   multi-day path: the range endpoint returns ascending and the parser
   takes the newest datum, so a window would silently return the wrong
   day's FIX for every day but the last.
3. Binance's `>= 2 klines` check was a settlement proxy tied to the
   `limit=2` request shape, not a payload sanity check. A date-pinned
   window returns exactly one row, so left unchanged it would have
   rejected every single recovery. It is now conditional, and the
   past-date guard supplies the settlement guarantee it was standing in
   for.
4. A fourth trap neither grill caught, called out as deliberate: Binance's
   `endTime` is treated as inclusive, so the window ends at
   `start + DAY_MS - 1`. A bare `start + DAY_MS` would have reached into
   the next day's candle.

**One deliberate deviation from the spec, and why.** §4 R1.1 says of a
windowed Twelve Data response: *"keep reading `values[0]`."* On the
date-pinned path this PR does not — it selects the row whose `datetime`
**is** the target date, and a response carrying no such row becomes a
symbol-attributable failure naming the date (`Twelve Data AAPL -> no bar
dated 2026-08-14 in the response`).

The reason is **§6 R3.2**, which was worked out after R1.1 was written. R3.2
makes *"a bar whose date did not match"* a first-class source of the **owed
and absent** state — the state that must exit 1. Positional selection asserts
nothing about which day was returned, so it cannot raise that state at the
boundary where the mismatch is attributable. It is the same dependency the
spec already records for R1.4 (*"without R1.4, R3's exit contract is
unimplementable"*), showing up a second time.

Concretely, `values[0]` fails in the wrong direction. If `end_date`
exclusivity ever slipped, the window would admit `target + 1` too, and Twelve
Data's newest-first default ordering would put the **neighbour day's bar**
first. The downstream stale-bar gate would still withhold the *mark*, so the
exit contract survives — but **the quote store is written before that gate**,
so `<dataDir>/prices/<id>.jsonl` would take the wrong day's close under the
target `asOf`, and the operator would see a generic "no fresh close" skip
rather than a failure naming the date. A silently wrong stored price is the
exact defect class this lane exists to close, so the check moved to the
provider boundary, ahead of any persistence. This deviation was escalated and
approved rather than taken by the implementing lane.

**Provenance is not forged.** An explicit `asOf` replaces the derived
trading day and nothing else — `now` stays the real clock, so `fetchedAt`
records the instant the recovery actually happened. Fabricating `now` was
rejected twice over: it would move `fetchedAt` too, making the durable log
assert the recovery never happened, and it steers no provider anyway. A
test asserts `asOf` and `fetchedAt` together, because a test checking only
`asOf` would pass on a forged clock — that mutation was applied and the
single-assertion version does pass under it, which is why the pair is
asserted.

**Validation lives in `runPriceFetch`, not the CLI**, so a programmatic
caller cannot bypass it: the engine's round-trip `isIsoCalendarDate` (so
`2026-02-30` is refused, never coerced to March 2) plus a strictly-earlier-
than-today rule that is load-bearing twice — it makes forcing the mark
gate open safe, and it guarantees Binance's target UTC day is complete.

**The live daily path is deliberately unchanged, and the no-flag pins are the
safety argument for this whole PR.** With no flag, every observable is
byte-identical: the same `asOf` derivation, the same mark gate, the same 13
attempted instruments, the same requests. That claim is not left to
inspection — it is pinned by dedicated no-`targetDate` suites in all three
providers (`twelvedata-provider.test.ts`, `banxico-provider.test.ts`,
`binance-provider.test.ts`), by `fetch-prices.test.ts`'s "with no asOf the
live path is unchanged (R2 pin)", and by `cli-main.test.ts`'s "with no
arguments the live daily path is unchanged". The provider pins assert exact
request strings rather than `toContain`, so a stray parameter fails them. The
pins are run as their own targeted invocation, separately from the full
suite, so their verdict is never mixed into a larger run.

The owed-set filter is **recovery-path only** — a Saturday daily run still
fetches all 9 Twelve Data symbols, still stores Friday's close under
Saturday's `asOf`, and still records 9 stale-mark skips as INFO with exit 0.
Under an explicit `asOf` that same stale skip is a failure, because the
request was pinned to the target date. Both halves are pinned, in both
directions: computing the absent set unconditionally turns the live Saturday
test red, and dropping the stale map turns the recovery test red.

**The exit contract, and why #266 D7/D8's literal-`0` does not bind here.**
#356 exists because a total failure exited green, so a recovery that
recovers nothing exits 1, naming each absent instrument with the provider's
own words and saying plainly that the exit code cannot distinguish a
market holiday from a provider failure. #266 D8's warn-only contract
governs recurring unattended automation — a permanently red nightly job is
one nobody reads — and this command is operator-invoked, one date at a
time. The recorded trigger that would flip it: if recovery is ever put
into an unattended catch-up loop.

**Scope.** The run writes stored quotes and inbox marks only — no
heartbeat, no spine invocation, no accumulus commit, no projection
refresh — and it is never routed through the daily wrapper. The delivery
half (making a lost day reach a human who is not looking) is #357 and is
out of scope. `docs/price-feed-ops.md` is also written by #357, which
merges after this one and inherits the merge.

**The review round.** Six findings, one MEDIUM and five LOW, all addressed,
all replied to inline and all resolved:

1. **MEDIUM** — the parser refused `pnpm prices:fetch -- --as-of=<date>`.
   pnpm 11 forwards the separator into argv, and `--` is this repo's own
   documented flag-forwarding idiom (`pnpm gap-report -- --write`), so an
   operator will type it. One leading bare `--` is now skipped. Only the
   leading position is forgiven — a `--` anywhere else is still an unknown
   argument, because a silently dropped flag would run the daily job and
   report success.
2. Refusals became a **type** rather than a path. `PriceFetchRefusal` is
   thrown by the parser and by the `asOf` validator, and the CLI renders a
   bare sentence for that type and nothing else. The previous gate — "an
   `--as-of` was given" — would have stripped the stack from an `EACCES`, an
   atomic-write failure or a defect in `buildMarks`, and only on the recovery
   path, which is the newer and less-exercised half of the command. Both
   directions are pinned now: a refusal renders bare, an unexpected fault
   keeps its stack on the recovery path *and* on the live path.
3. The Twelve Data row selection above.
4. `fetchJson`'s JSDoc had been orphaned by a block inserted beneath it; it
   sits on the function it documents again.
5. **Fixture authorship** — `17.0218` was the real published Banxico FIX for
   08-14 and had no business in a committed fixture. It is now an invented
   `18.4407`, and the rest of the PR's fixtures were swept for captured
   values; none remained.
6. The runbook's wrapper-step mapping was corrected against the Components
   table, and `toEqual({})` tightened to `toStrictEqual({})` — which `toEqual`
   would have passed even if the CLI began sending `{ asOf: undefined }`.

**Verification**, after merging `origin/main` into the branch (a merge, not a
rebase — nothing on this branch was rewritten):

- **Full suite: 2386 passed, 48 skipped, 0 failed** (156 files passed, 4
  skipped). The two failures this PR carried earlier are gone: they were
  `ops/testkit/gitignored-path-globs.test.ts`'s *"reaches nested .gitignore
  files, anchored to their own directory"* and *"still excludes agent worktrees
  under .claude/ — the bug this closed"*, which fail on any fresh clone and are
  fixed by #368. Merging `origin/main` cleared them; nothing in `ops/testkit/`
  was touched by this branch.
- **The no-flag live-path pins: 17 passed, run as their own targeted
  invocation** so their verdict is never averaged into a full-suite run. They
  span all three providers, `runPriceFetch`, the CLI and the argument parser —
  including the new pin that feeds a two-row live response and asserts
  `values[0]` still wins, which is what stops the date-selecting branch from
  leaking onto the nightly path.
- **`pnpm typecheck`: clean** across all six packages plus `tsconfig.ops.json`.

The merge brought in `packages/engine`, `apps/tui` and `ops/testkit` changes
and touched **no** `apps/price-feed` file, so there was no collision with this
branch's work.

Every new guard was mutation-tested: the mutation named, applied, the red
observed, then restored. Every fixture is hand-authored — no captured
provider response and no line from the private ledger.

**Not done here, deliberately:** the operational recovery of 2026-08-14
and 2026-08-15 against the real ledger is the operator's to run after this
merges.


apps/price-feed/src/cli-args.ts
```diff
@@ -0,0 +1,114 @@
...
+      claim(take(argv[i + 1], arg));
+      i++;
+      continue;
+    }
+    if (arg.startsWith(`${AS_OF_FLAG}=`)) {
+      claim(take(arg.slice(AS_OF_FLAG.length + 1), arg));
+      continue;
+    }
+    throw new Error(
+      `unknown argument "${arg}". \`prices:fetch\` takes ${AS_OF_FLAG} and nothing ` +
```
AmetAlvirde: **MEDIUM — the parser refuses `pnpm prices:fetch -- --as-of=<date>`, which is the repo's own documented flag-forwarding form.**

Verified on this branch:

```
["--","--as-of=2026-08-14"] -> THROWS: unknown argument "--". `prices:fetch` takes --as-of and nothing else...
["--as-of=2026-08-14"]      -> {"asOf":"2026-08-14"}
["--as-of","2026-08-14"]    -> {"asOf":"2026-08-14"}
```

And verified against pnpm 11.1.2 itself, with a throwaway script outside the repo:

```
$ pnpm run echoargs --as-of=2026-08-14      -> ["--as-of=2026-08-14"]
$ pnpm run echoargs -- --as-of=2026-08-14   -> ["--","--as-of=2026-08-14"]
```

So both documented spellings work without `--` (good), but the `--` form — which this very doc teaches elsewhere, `pnpm gap-report -- --write` in `docs/price-feed-ops.md`, and which is the habit most people carry into pnpm — lands the literal `--` in argv[0] and gets refused.

It **fails safe** (exit 1, no daily run, no silent success), so this is friction rather than danger. But the runbook exists so a stressed operator gets it right the first time, and "unknown argument \"--\"" reads like the command is broken rather than like a spelling correction.

Actionable, one line: skip a single leading `--` before the walk (`const args = argv[0] === "--" ? argv.slice(1) : argv;`), or state in the runbook that `--` must not be used here. Either is fine; silently refusing the repo's other convention is the part worth fixing.

apps/price-feed/src/provider.ts
```diff
@@ -66,7 +66,61 @@ interface FetchJsonOptions {
...
+ * bare `HTTP 400 Bad Request` reads as a malformed request when on the recovery path
+ * it almost always means that day had no trading. The `HTTP <status> <statusText>`
+ * prefix stays first and byte-identical — callers match it by prefix.
+ */
+/**
+ * How much of a provider's own error text may ride along in a `reason`. Generous
+ * enough for any real API sentence, small enough that an HTML interstitial served
+ * in place of JSON cannot flood a scheduled run's log.
+ */
+const MAX_PROVIDER_TEXT = 200;
```
AmetAlvirde: **LOW — `fetchJson`'s JSDoc is now orphaned; this block took it over.**

The long doc comment above (ending `…callers match it by prefix.`) used to sit immediately above `export async function fetchJson`. The new `MAX_PROVIDER_TEXT` / `clip` / `readErrorText` block was inserted between them, so TypeScript now attaches that whole comment — including the new R1.4 paragraph that explains the appended provider sentence — to `const MAX_PROVIDER_TEXT`, and `fetchJson` at line 124 has no doc comment at all. Hover, `.d.ts` emit and any doc tooling all lose it.

Fix: move the three new declarations above the `fetchJson` doc block, or move the `fetchJson` doc back down so it sits directly on the function.

apps/price-feed/src/cli-main.ts
```diff
@@ -0,0 +1,295 @@
...
+
+  let result: FetchRunResult;
+  try {
+    result = await run(options);
+  } catch (error) {
+    // On the recovery path the only thing that throws before any IO is R2.2's
+    // validation refusal (not a real calendar day, or not strictly in the past), and
+    // an operator who just typed a date needs its sentence, not a stack. The live
+    // path keeps the original stack-preserving behaviour — there, a throw IS a bug.
+    if (recovering) {
```
AmetAlvirde: **LOW — this catch is gated on `recovering`, not on "this was a validation refusal", so a genuine bug during a recovery loses its stack.**

The comment says "the only thing that throws before any IO is R2.2's validation refusal" — true as written, but the `try` wraps the *whole* run, not just the pre-IO part. An `EACCES` from `mkdir`, an atomic-write failure in the price store or the inbox, or a real defect inside `buildMarks` all reach here on the `--as-of` path and get printed as one sentence with no stack, while the identical fault on the live path keeps its stack via `cli.ts`'s `.catch`.

That's the wrong asymmetry: the recovery path is the *newer* code, and it is the one where a bug is harder to see.

Actionable: have `validateAsOfOverride` throw a tagged error (a small `AsOfRefusal extends Error`, or a sentinel property) and render only that as a bare sentence; rethrow everything else so it keeps its stack. The two acceptance criteria that care about this (`--as-of=<today>` / `2026-02-30` refused "without a stack trace") are unaffected.

apps/price-feed/src/twelvedata-provider.ts
```diff
@@ -90,9 +100,19 @@ export async function fetchTwelveDataDailyCloses(
...
+    // ⚠️ Twelve Data's `end_date` is EXCLUSIVE. `start_date === end_date` returns
+    // the no-data 400 EVEN WHEN THE BAR EXISTS, and that 400 is indistinguishable
+    // from "that day did not trade" — a silent wrongness. The window is therefore
+    // target..target+1, and the +1 comes from the engine's UTC `addDays` rather
+    // than local-time `Date` arithmetic or string surgery.
+    const selector =
+      options.targetDate === undefined
+        ? `&outputsize=1`
+        : `&start_date=${encodeURIComponent(options.targetDate)}` +
+          `&end_date=${encodeURIComponent(addDays(options.targetDate, 1))}`;
```
AmetAlvirde: **LOW / uncertain-by-design — the windowed selection depends on `end_date` exclusivity *and* on the default result ordering, and only one of those is pinned.**

The window is correct given the measured exclusivity, and the test that fails if `start_date === end_date` is a good guard. But the row is still picked positionally (`values[0]`, unchanged below). If the exclusivity assumption ever slips — a provider change, or a symbol where the semantics differ — the window admits `target+1` too, and `values[0]` under Twelve Data's newest-first default ordering is then the **wrong day's bar**.

What happens then is mostly good: `observationDate !== asOf`, the mark is withheld, `classifyAbsent` reports it ABSENT and the run exits 1. But the *quote store* is written before that gate (two-plane rule), so `<dataDir>/prices/<id>.jsonl` gets the neighbouring day's price under the target `asOf`. That store is the disposable cache and the operator is told to look, so this is not a correctness hole in the durable log — it is just a free hardening you can take or leave.

If you want it: on the pinned path, select the row whose `datetime` equals `targetDate` rather than `values[0]`. Then neither exclusivity nor ordering is load-bearing. Flagging as a judgment call, not a defect.

apps/price-feed/src/cli-main.test.ts
```diff
@@ -0,0 +1,461 @@
...
+    if (href.includes("banxico.org.mx")) {
+      const window = /\/datos\/(\d{4})-(\d{2})-(\d{2})\//.exec(href)!;
+      return Promise.resolve(
+        new Response(
+          JSON.stringify({
+            bmx: {
+              series: [
+                {
+                  idSerie: "SF43718",
+                  datos: [{ fecha: `${window[3]}/${window[2]}/${window[1]}`, dato: "17.0218" }],
```
AmetAlvirde: **LOW — fixture authorship: `17.0218` is the real Banxico FIX for the real recovered date.**

The spec records it as measured first-hand ("Banxico served **17.0218** for 14/08"), and here it is a fixture value for `--as-of=2026-08-14`. Every other suite on this branch uses an invented rate — `18.5` in `fetch-prices.test.ts`, `18.9012` in `banxico-provider.test.ts` — so this is the one number in the new fixtures that came off a live provider rather than out of the author's head.

It is not a captured response body and not a line from the private ledger, so it is nowhere near the serious end of the rule. But the standing constraint is *authorship*, and the surrounding block is titled hand-authored. Swap it for an invented rate and the whole set is consistent.

(Same file, trivial: the comment on this test says "Driven through the REAL `runPriceFetch` (no `run` injected)" and then injects a `run` that wraps the real one. The second comment gets it right; the first sentence contradicts it.)

docs/price-feed-ops.md
```diff
@@ -592,6 +592,173 @@ private `<fund>` repo, `~/Dev/<fund>/data` by default. The `inbox/` and
...
+
+- **Writes the heartbeat.** `job-heartbeat.json` records whether the daily job
+  ran that evening; its only writer is the wrapper's `EXIT` trap. Stamping it
+  for a recovered past date would assert an evening run that never happened —
+  manufacturing exactly the clean-looking surface a lost day hides behind.
+- **Runs the spine.** `--as-of` is a flag on the fetch step, not a new
+  pipeline; folding the recovered marks into the event log is still step 3 of
+  the procedure above, run by hand.
+- **Commits to the `<fund>` data repo, or refreshes the hosted projection.**
+  Both stay the operator's, as steps 3–4 above and steps 5–6 of the daily
```
AmetAlvirde: **LOW — the wrapper-step mapping in this bullet doesn't line up with the Components table above.**

The bullet names two things (commit to the `<fund>` repo, refresh the hosted projection) and maps them "respectively" to "steps 3–4 above and steps 5–6 of the daily wrapper (`pnpm gap-report -- --write`, `pnpm backfill`)". Per the Components table:

- the wrapper's **commit** is step **3** (step 4 is the post-check, which is not the operator's and is not in the recovery procedure);
- only step **6** (`pnpm backfill`) refreshes the projection — step 5 (`pnpm gap-report -- --write`) is neither of the two things this bullet names, and the recovery procedure above never asks the operator to run it.

Suggest: "…as step 4 of the procedure above (the wrapper's step 3) and `pnpm backfill` (the wrapper's step 6) respectively." Everything else in the section checks out — I walked the two example outputs column by column against the `padEnd(7)` / `padEnd(11)` format strings in `cli-main.ts` and the line order matches exactly, and `<dataDir>/prices` + `<dataDir>/inbox/transactions.json` are the real resolved paths.

AmetAlvirde (commented):
## Review — no blocking findings

**Verdict: nothing at HIGH severity.** The safety argument holds: I could not find a way for this branch to change what the nightly 18:00 job does. Six findings below, one MEDIUM and five LOW, all posted inline. None of them blocks a merge.

I read the approved spec first and did not re-open any of the four §8 decisions.

### 1. The live daily path — the thing that mattered most

I checked this four ways and it survives all four.

- **Request byte-identity.** With no `targetDate`: Twelve Data reassembles to exactly `?symbol=…&interval=1day&outputsize=1&apikey=…` (the selector concatenates in the original position); Banxico's `seriesUrl(undefined)` reproduces the old constant character for character; Binance's `base + "&limit=2"` reproduces the old template. Pinned by exact-string tests in all three suites, not by `toContain`.
- **`runPriceFetch` with no `asOf`.** `owed()` short-circuits to `return entries` before allocating, `targetDate` is `undefined` at every call site, `markEmitted` falls straight through to `isAtOrAfterMarkTime`, `notOwed` stays `[]`, `totalCount` stays 13. Passing `targetDate: undefined` positionally to Binance is fine — the provider branches on `=== undefined`, and Twelve Data/Banxico get a conditional spread instead because their option types are not `| undefined`.
- **The `cli.ts` → `cli-main.ts` split.** I diffed the new function against `main:apps/price-feed/src/cli.ts` line by line: every printed string, every `padEnd` width and the emission order are unchanged, and the exit expression gains only `absent.length > 0`, which is `0` without the flag. `process.exitCode = 0` where the old code left it unset is equivalent.
- **The scheduler.** `ops/price-feed/run-daily-fetch.sh:625` invokes `pnpm prices:fetch` with **zero** arguments and only reads `$FETCH_STATUS`, so the new parser is never handed anything and the exit contract it depends on is untouched. Worth noting the wrapper harness skipped (27 cases) because the arming path set does not include `apps/price-feed/src/cli*.ts` — but the harness stubs `pnpm` via `NUMISMA_PATH_PREPEND`, so arming it would not have exercised the real CLI anyway. Reading the wrapper is the right check here, and it passes.

One deliberate live-path change, and it is the spec'd one: `fetchJson`'s non-ok `reason` now appends the provider's own sentence (§4 R1.4). The `HTTP <status> <statusText>` prefix stays first, every caller matches by prefix, and the body read is inside the abort-guarded region with the timer still live, so a hanging error body still resolves to the bare fallback rather than stalling the run.

### 2. The three date windows

All three are right, each in its own convention.

- **Twelve Data** — `end_date = addDays(target, 1)`, exclusive end handled, `addDays` is the engine's UTC one (`new Date(\`${asOf}T00:00:00Z\`)` + `setUTCDate`), so no local-timezone slippage and no string surgery. Month boundary covered by a test (`2026-02-28 → 2026-03-01`).
- **Banxico** — pinned to `/datos/<target>/<target>`, both ends the same day, `extractLatestDatum` untouched. There is no code path that can produce a multi-day range: `seriesUrl` takes one string and interpolates it twice. The ascending-order trap is structurally unreachable, not merely avoided.
- **Binance** — `startTime = Date.parse(\`${target}T00:00:00.000Z\`)`, `endTime = start + DAY_MS - 1`, inclusive-end handled. The test computes the expected bound independently as `Date.UTC(2026, 7, 12, 23, 59, 59, 999)` rather than reusing the provider's arithmetic, so it cannot agree with a bug in it. `limit=` is asserted absent on the pinned path, and the `>= 2` settlement proxy is pinned on *both* sides — still firing on the live path, not firing on the pinned one.

The one soft spot is Twelve Data's positional `values[0]` read, which leans on exclusivity *and* on default ordering. Detail inline; it fails safe, so it is a hardening suggestion rather than a defect.

### 3. Decorative tests — none found

I walked every new guard and named the mutation that would turn it red. I did **not** apply the mutations: the worktree was read-only for this review, so this is static analysis rather than an executed mutation run — worth knowing when weighing it against the PR body's own mutation-testing claim.

The ones I checked: the end-date-inequality pin, both live-path URL pins, the Binance `-1` bound, both halves of the `>= 2` guard, the zero-row date-naming throw (deleting that branch produces "unexpected payload shape", which does not contain the date, so the regex goes red), `isIsoCalendarDate` vs a shape regex, `>=` vs `>` on the past-date rule, the forced mark gate *with its live-clock control*, the owed-filter's recovery-only scoping in both directions, and the exit-code term.

**The stale-bar boundary is genuinely pinned both ways**, which is the pair that matters: computing `absent` unconditionally turns the live Saturday test red (9 unmarked equities, none in `notOwed`, exit 1), and dropping the stale map turns the recovery test red (the reason string degrades to "no reason was recorded"). Neither half can regress under cover of the other.

**One shape that is not reachable through the real pipeline**, since you asked: `cli-main.test.ts`'s "vanished with no reason recorded at all". Today an unmarked instrument is always either a `failure` or a `staleMarkSkip`, so that third `reason` branch cannot fire. It is deliberate defence — the whole point of deriving `absent` from the registry rather than the failure list — and I would keep it. Noting it only because it is the class of thing you asked me to look for.

### 4. Fixtures

Authored, with one exception worth a swap: `cli-main.test.ts:435` hardcodes `17.0218`, the real Banxico FIX for the real recovered date. Inline. Everything else is invented — `18.5`, `18.9012`, the `kline()` filler, the Twelve Data bodies. The Binance `{ code: -1121, msg: … }` and Twelve Data's "No data is available on the specified dates" reproduce real *formats* from knowledge, which is authorship, not capture.

### 5. The parser

Solid, and it closes the typo it was built to close. `--asof=`, bare positionals, missing values, `--as-of --other`, empty `--as-of=`, and a duplicate flag are all refused with the usage block attached; shape-only here with the semantic half in `runPriceFetch`, so a programmatic caller crosses the same boundary. One gap — the leading `--` form — is inline and is the single MEDIUM.

### 6. The runbook

Accurate. I checked both example outputs column by column against the real format strings (`padEnd(7)` / `padEnd(11)`, the two-space gap before the price), confirmed the line order matches the emission order in `runPriceFetchCli`, and confirmed `<dataDir>/prices` and `<dataDir>/inbox/transactions.json` are what `resolvePriceFeedPaths` actually produces. The exit contract, the no-range argument, the D10 refusals and the D11 vocabulary note all match the shipped code. One wrapper-step mis-mapping, inline.

### 7. Docs outside this PR's scope, as follow-up

Spec §7 scoped R4 to `docs/price-feed-ops.md`, so this is a note rather than a finding, but the CLI split leaves four maintained docs describing the old shape:

- `apps/price-feed/README.md:56` — "The CLI (`src/cli.ts`) — thin console/exit-code wiring … No domain logic." The exit contract and the three-state classification now live in `cli-main.ts`, which the README does not mention.
- `docs/coverage-rationale.md:36` — justifies excluding `cli.ts` because it holds "top-level console reporting and exit-code wiring". That is now `cli-main.ts`, which is measured.
- `docs/coverage-rationale.md:517` — the uncovered-line list for `fetch-prices.ts` (`:161,162,164,260,286,356,385`) is stale: the file grew ~158 lines, and `:161-164` now points inside the `marks` JSDoc.
- `docs/codebase-map.md` does not list `cli-main.ts` or `cli-args.ts`.

Also small: the `index.ts` barrel was not widened. `FetchRunResult.notOwed` is typed `readonly NotOwed[]` and `fetchBinanceDailyClose` now takes `BinanceFetchOptions`; both types are exported from their modules but neither is re-exported from the barrel that `index.test.ts` locks, so a barrel consumer can read `notOwed` without being able to name its element type. One line each, if you want the barrel to stay honest.

Nit, take it or leave it: `cli-main.test.ts:115`'s `expect(captured.options).toEqual({})` would still pass if the CLI started sending `{ asOf: undefined }`, because `toEqual` ignores undefined-valued keys. Harmless today; `toStrictEqual` is the assertion that means what the comment says.

### Verification I ran

- `npx vitest run apps/price-feed` — **189 passed, 27 skipped, 0 failed**. The 27 skips are the wrapper-harness cases the arming trigger muted.
- `pnpm typecheck` — clean across all six packages plus `tsconfig.ops.json`.
- pnpm 11.1.2 argument forwarding, probed with a throwaway package outside the repo.
- `parsePriceFetchArgs` driven directly for the three argv forms.

I did **not** run the full-repo suite, `pnpm smoke:startup`, or a coverage run, and I did not execute any mutation. The two known `ops/testkit/gitignored-path-globs.test.ts` failures are out of scope per the brief and I did not touch them.

apps/price-feed/src/cli-args.ts
AmetAlvirde: Fixed. `parsePriceFetchArgs` now skips ONE leading bare `--` (`const args = argv[0] === "--" ? argv.slice(1) : argv;`) before the walk, so `pnpm prices:fetch -- --as-of=2026-08-14` parses. Confirmed on the real binary: that form now reaches R2.2's validator (`-- --as-of=2026-02-30` refuses with the calendar-date sentence, no stack), and the plain form is unchanged.

The unknown-argument guard is deliberately NOT weakened — only the leading position is forgiven. New cases in `cli-args.test.ts`: `["--","--as-of=<d>"]` and `["--","--as-of","<d>"]` parse, `["--"]` alone is the live path, and `["--as-of=<d>","--"]`, `["--as-of","<d>","--"]`, `["--","--"]`, `["--","--asof=<d>"]`, `["--","2026-08-14"]` all still throw.

Mutations run: removing the skip → 4 red across `cli-args.test.ts` + `cli-main.test.ts`; replacing it with a global `argv.filter(a => a !== "--")` → 3 red (the misplaced-`--` cases), which is the guard that stops the fix from becoming a hole.

apps/price-feed/src/provider.ts
AmetAlvirde: Fixed. The `MAX_PROVIDER_TEXT` / `clip` / `readErrorText` block moved ABOVE the `fetchJson` doc comment, so the comment (R1.4 paragraph included) sits directly on the function again. The constant kept its own short doc.

apps/price-feed/src/cli-main.ts
AmetAlvirde: Fixed, and gated on the error TYPE rather than the path, as suggested.

New `apps/price-feed/src/refusal.ts` exports `PriceFetchRefusal extends Error`. Every throw in `parsePriceFetchArgs` and both throws in `validateAsOfOverride` now use it; `runPriceFetchCli`'s catch is `if (error instanceof PriceFetchRefusal) { … return 1 } throw error`. So an EACCES from `mkdir`, an atomic-write failure or a defect in `buildMarks` keeps its stack on BOTH paths, and the two acceptance criteria (`--as-of=<today>` / `2026-02-30` refused without a stack) still hold. The parse catch rethrows a non-refusal too, so a defect in the parser itself is not laundered either.

The two refusal-rendering tests now drive the REAL `runPriceFetch` (the refusal fires before any IO, so nothing is written) instead of a hand-thrown `new Error` — otherwise they would assert the fixture, not the type gate. Added two tests for the other half: an unexpected fault rethrows with its stack on the recovery path AND on the live path.

Mutations run: gating on `recovering` again → 1 red (the recovery-path fault test); dropping the rethrow entirely → 2 red; throwing plain `Error` from `validateAsOfOverride` → 2 red; throwing plain `Error` from the unknown-argument branch → 2 red.

apps/price-feed/src/twelvedata-provider.ts
AmetAlvirde: Taken — with the reviewer's framing, and flagged as a deliberate narrow deviation from §4 R1.1's "keep reading `values[0]`", accepted for this PR.

On the date-pinned path only, `observationFromBody` now selects the row whose `datetime` equals `targetDate`; no matching row becomes the symbol-attributable failure naming the date (`Twelve Data AAPL -> no bar dated 2026-06-30 in the response`), the same channel a zero-row payload already used. The quote store is written before the stale-bar gate, so `values[0]` on a slipped window would have put the neighbour day's price under the target `asOf` — silent wrong data, which is what this lane exists to close.

**The live path is untouched**: with no `targetDate` it still reads `values[0]`, and that is now pinned by a test with a two-row live response.

Mutations run: pinned path back to `values[0]` → 2 red; date-selection leaking onto the live path → 5 red; the no-matching-row failure no longer naming the date → 1 red.

apps/price-feed/src/cli-main.test.ts
AmetAlvirde: Both fixed. `17.0218` → an invented `18.4407`, so every fixture in the lane is authored. Swept the rest of the PR's new fixture numbers for captured values and found none — the remaining rates and closes are invented.

The contradictory first sentence is corrected too: that comment now reads "wrapped by the injected `run`" rather than "no `run` injected".

Also from the follow-up notes in the review body: `expect(captured.options).toEqual({})` → `toStrictEqual({})` (mutation: making the CLI send `{ asOf: undefined }` → 1 red, which `toEqual` would have passed).

docs/price-feed-ops.md
AmetAlvirde: Fixed against the Components table and the wrapper. The bullet now reads: "Both stay the operator's: the commit is step 4 of the procedure above (the wrapper's step 3), and refreshing the projection is `pnpm backfill` (the wrapper's step 6)." The `gap-report -- --write` reference is gone — it is neither of the two things the bullet names and the recovery procedure never asks for it. Diff kept to those two lines, since Lane S (#357) also writes this file.

apps/price-feed/src/twelvedata-provider.ts
AmetAlvirde: Correcting the record on one phrase in the reply above, before this thread is resolved.

"Accepted for this PR" was not the implementer's to write. A deviation from a spec line is not something the lane doing the work can wave through on its own say-so, and it was escalated rather than inherited. **It has now been approved on the board** — and the approved reasoning is stronger and more specific than the "optional hardening" this thread framed it as, so it is worth stating properly.

**The reason is §6 R3.2, not general robustness.** R3.2 makes *"a bar whose date did not match"* a first-class source of the **owed and absent** state — the state that must exit 1. R1.1's `values[0]` line was written before R3's exit contract was worked out. This is the same dependency the spec already records for R1.4 (*"without R1.4, R3's exit contract is unimplementable"*), showing up a second time.

**One precision on the failure mode**, since the reply above slightly overstates it: with `values[0]` the date mismatch is *not* wholly undetected. The downstream stale-bar gate compares `observationDate` to `asOf`, so the **mark** is still correctly withheld and the exit contract still fires. What positional selection loses is narrower and still serious:

- **The quote store is written before that gate**, so `<dataDir>/prices/<id>.jsonl` takes the neighbour day's close under the target `asOf`. A silently wrong stored price is precisely the defect class this lane exists to close.
- The operator sees a generic "no fresh close" stale skip instead of a failure that **names the date** and attributes it to the symbol.

Date selection moves the check to the provider boundary, ahead of any persistence, which is where R3.2's date-mismatch state is actually attributable.

The conditions attached to the approval are both met and both pinned: **the live path still reads `values[0]`** (a test feeds a two-row live response and asserts position 0 wins), and the deviation is stated explicitly in the PR body with the R3.2 justification rather than being left to read as an incidental improvement.

PR #375, opened 2026-08-18
The delivery seam: a pull-only channel becomes an unmissable one

### LINKED ISSUE #357: Price feed: the lost-day signal is computed and warned, but reaches no one outside the TUI

### LINKED ISSUE #374: The wrapper harness does not arm on an uncommitted wrapper edit — a trim eats one character of the first dirty path
## The wrapper harness does not arm on an uncommitted edit to the wrapper

`apps/price-feed/src/wrapper-harness/arming-trigger.testkit.ts` fails to detect a modified-but-uncommitted `ops/price-feed/run-daily-fetch.sh` — the exact case the harness advertises as *"sees an UNCOMMITTED wrapper edit — the moment you most want the suite."*

### Mechanism — two lines, one off-by-one

1. `git()` (~line 205) returns `execFileSync(...).trim()`, trimming the **whole** `git status --porcelain` output. That eats the leading space of the **first line only**.
2. `parsePorcelain` (~line 216) then does `line.slice(3)` — correct for `" M path"` and `"?? path"`, but one character off for that first, de-spaced line.

Net effect: the first modified-not-staged path silently loses one character. Every other path in the same listing is parsed correctly, which is why this survived.

### Measured, not inferred

Probing `resolveTriggerFacts` directly in a worktree with a modified wrapper returned:

```
"ps/price-feed/run-daily-fetch.sh"
```

The leading `o` is gone, so the path misses the `ops/price-feed/**` glob, and the harness printed:

```
[wrapper harness] harness skipped: no changes under ops/price-feed/**, ... since 7a5690c
```

…with `run-daily-fetch.sh` visibly modified in `git status`. Every other entry in the same `changedPaths` array (`package.json`, `packages/event-store/src/index.ts`, five `??` untracked files) was correct.

This bites hardest for this repo's layout specifically: among dirty files, `ops/…` sorts first, so the wrapper is usually the mangled entry.

### Why the existing test did not catch it

The test at `run-daily-fetch.test.ts:440` writes the wrapper into a fresh fixture dir, so git reports it as `?? path`. The `??` status field has **no leading space**, so it survives the whole-output trim intact and parses correctly.

**The test passes because it pins a shape the real producer never emits.** A tracked, modified wrapper (` M`) — the shape the advertised behaviour actually depends on — is never exercised.

This is the same failure family the board has now hit five times: a guard that cannot fail on the thing it names.

### Blast radius — bounded

**CI was never blind.** The committed half of the changed-set (`git diff --name-only base...HEAD`) is unaffected, so once a wrapper edit is committed the harness arms normally. The gap is **local iteration only**: an author editing the wrapper and running the suite gets a silent skip and may believe their change is covered when nothing ran.

Workaround while this is open: `NUMISMA_WRAPPER_TEST=always`.

### Fix, bounded to three things

1. `git()` — stop mangling the first line. Trim trailing only, or do not trim before parsing. Other `git()` callers (`rev-parse`, `merge-base`, `diff`) rely on the current trim for clean single-line values and must keep working.
2. The test at `run-daily-fetch.test.ts:440` — pin the **reachable** shape: a tracked, modified wrapper (` M`), not an untracked one. An untracked case may stay if it earns its place.
3. Nothing else in `arming-trigger.testkit.ts`. No refactor of `decideArming`, `matchesPathSet` or `TRIGGER_PATH_SET`.

Acceptance: with the fix in place and an uncommitted wrapper edit in the tree, the harness **arms on its own**, with no `NUMISMA_WRAPPER_TEST` override. Mutation proof: restoring the `.trim()` must turn the corrected test red.

### Note

Found while building the lost-day delivery seam (#357), whose slice 3 adds wrapper-harness cases and therefore depends on this instrument being sound. Fixed there rather than deferred, because validating new harness cases only through the `always` override would validate them through a path the next reader does not get.

AmetAlvirde:
Every derivation in this repo was already correct. The defect was that **every surface is pull-only**.

On 2026-08-14 and 08-15 the price feed lost two days. The TUI banner named both, live and correctly, the whole time. The dashboard's freshness verdict was right. `gap-report.json` was right after the next run. Nobody looked for three days — while sitting at the machine. A standup found it on 08-17.

So this is a **transport, not a detector**. It adds no derivation of any kind. It composes signals that already existed into a file the shell profile `cat`s on every new terminal, so the first moment the operator is back is unmissable.

This closes #357. It also closes #374, an instrument defect found while building it — see "The arming trigger" below.

## What it does

A well-known `operator-notice.txt` lands beside the durable log, next to `gap-report.json` and `job-heartbeat.json`. It is rewritten every run, with **no rotation, no history, and no dismissal state**. **Empty means healthy** — a clean machine prints nothing on a new shell, and there is nothing to clear by hand.

When it does print, it carries the same two signals the TUI banner does, in the same cause-then-effect order — the job first, the data second — because a failed run is *why* the days went missing.

**Lost days are enumerated, each with its own recovery command. Venue-dark days are counted on one line.** That split is the decision this channel lives or dies on. A lost day is remediable and self-extinguishing: name the date, run the command, the row is gone. A venue-dark day is permanent and accumulates — about ten a year are US market holidays and no command will ever clear one. Enumerating those on a channel that prints on every new terminal is cry-wolf death arriving on a schedule, inside the fix.

### What it would have caught, measured against the real store

Run read-only against the real store, the notice's steady state today is **5 lines / 444 bytes**: two lost days — **2026-08-14 and 2026-08-15**, exactly the pair the 08-17 standup surfaced after three days of nobody looking — each with its own `pnpm prices:fetch --as-of=<date>` line, plus one venue-dark count line.

**Had this shipped before 08-14, the first new terminal opened on 08-16 would have printed both dates and the command for each.** That is #357 demonstrated rather than argued.

## Where the composition lives, and why

`packages/event-store/src/operator-notice.ts`, beside the primitives it composes — `loadHeartbeatLines`, `loadGapReport`, `formatLostDays` were all already exported from that package.

The grill said to call `loadLivenessLines`. That lives in `apps/tui`, and **an app may not import an app here**: `gap-lines.ts:12-19` records that a prototype did exactly that through a computed specifier `tsc` could not follow, and that both the bridge and its hand-written interface were deleted for it. Verified: nothing outside `apps/tui` imports that module today.

The standing recommendation went one step further — lift the composition so the banner and the notice share **one** composer. **That half is deliberately not done, and the reason is a real constraint rather than a shortcut.** The banner *enumerates* venue-dark days under `MAX_GAP_LINES`/`RESERVED_LOST_LINES`; the notice *counts* them. One composer serving both would have to change TUI behaviour to satisfy a different channel's noise budget. What the two genuinely share is the **primitives**, and those already live in the package. So the composer moved next to the vocabulary rather than the vocabulary moving next to a renderer.

## The cap — found by running the deliverable, not by inspecting it

Every unit test passed an explicit narrow window. The real CLI passes none, so production gets the derivation's ~90-day default. Smoke-tested against a throwaway store with no log, the notice wrote **90 lines / 7425 bytes** — 45 lost days, each with its own recovery command. Correct, well-formed, and unusable.

**A notice that prints 90 lines on first contact trains its reader to skip it on day one, which reproduces #357 inside the fix for #357.** The runbook in this PR tells operators to wire this into their profile, so that first impression is not hypothetical.

**This was found by executing the deliverable, not by inspecting it.** That is the sixth time on this board that a green suite has been pinning a shape production never emits — and the first one caught by running the thing rather than reading it. The arming-trigger defect below is the seventh, found the same way in the same session.

The bound was not removed by anyone. The argument for reusing `loadLivenessLines` was, verbatim, that it *"already bounds its own line count"* — and when that reuse was correctly forbidden, the bound was the one property that did not survive the reroute. Everyone assumed it came along.

`MAX_NOTICE_LOST_DAYS = 10` now enumerates the most recent lost days and collapses earlier ones into one `N earlier lost day(s) withheld — enumerate them with pnpm gap-report.` line. **Ten was chosen against a real outage shape, not an aesthetic one:** two lost days are unrecovered right now and three more are expected from a known away window, so the realistic near-term is about five — and a cap that withheld those would hide exactly the debt the notice exists to surface. Measured against the real store, steady state is **5 lines / 444 bytes**, so the cap costs nothing in the common case and only engages where the channel would otherwise destroy itself.

**The accepted trade, recorded rather than silent:** a withheld day is no longer individually named with its command, so the self-extinguishing property applies only to what is shown. The tail is what keeps the rest visible.

## Step 0 — speaking when the toolchain cannot

Step 5b only speaks on a run that *reaches* step 5b. The two failures this channel exists for are exactly the runs that do not: an unresolvable `pnpm`/`node`, and a run that died partway. So a pure-bash step 0 runs **before `resolve-tools`**, using only `sed`, `head` and `printf` — nothing that could itself be the missing thing.

It is a **pure equality read** of the previous heartbeat's `exitCode` and `lastStep`. **No date math, no threshold, no clock, no comparison of instants.** That is a decision, not an unfinished implementation: a staleness rule in bash would be a second copy of `dueThrough`'s contract with nothing forcing the two to agree, on the one channel where a disagreement reads as an all-clear. A rule that needs a pin is a rule worth not creating.

Two behaviours are load-bearing and easy to get wrong:

- **A clean previous run leaves the file completely untouched — not truncated.** The notice standing there was written by the previous run's step 5b and may name real lost days. Clearing it at the *start* of a run, before that run has proved it can do anything, and then dying at step 3, would leave an empty notice standing over a real loss: `cat` prints nothing and the operator reads silence as health. Self-clearing belongs to step 5b alone, which runs only once the day's work is done.
- **An absent or unparseable heartbeat says nothing and writes nothing.** A machine that has never run this job must not be told on its first morning that the last run failed.

## The arming trigger (closes #374)

The wrapper harness did not arm on an uncommitted edit to the wrapper — the exact case it advertises as *"the moment you most want the suite."* `git()` trimmed the whole `git status --porcelain` output, eating the leading space of the **first line only**; `parsePorcelain`'s `line.slice(3)` then cut one character off that first path. `ops/price-feed/run-daily-fetch.sh` became `ps/price-feed/run-daily-fetch.sh`, missed the `ops/price-feed/**` glob, and the suite skipped.

The old test passed because it wrote the wrapper into a fresh directory, so git reported `??` — **a status with no leading space, which survives the trim intact.** It pinned a shape the real producer never emits.

This was fixed here rather than deferred because it is a defect in the instrument that validates this PR's own slice-3 cases. Validating them only through `NUMISMA_WRAPPER_TEST=always` would have validated them through a path the next reader does not get. CI was never blind — the committed-diff half is unaffected — so the gap was local iteration only.

## Known limits, bounded and deliberate

- **A run that never fires at all is not covered.** The check rides the existing wrapper, so it speaks only when a run runs. Catching a silently-absent job needs a clock and a second LaunchAgent, which was weighed and rejected on cost. This is a known limit, not a bug and not a TODO.
- **A misconfigured `NUMISMA_DATA_DIR` is a case step 0 cannot see.** The refusal for a blank or whitespace data dir fires before `HEARTBEAT_FILE` is derived, so such a run writes no heartbeat at all and leaves the previous run's in place — which says "clean." Step 0 will therefore report healthy. That is the same silently-absent-job shape above, it is out of scope for the same reason, and it is recorded here so the next reader does not think step 0 missed it by accident.

- **The notice's job half is one run stale — filed as #376, not fixed here.** `job-heartbeat.json` has one writer, the `EXIT` trap, so step 5b composes the job half against the *previous* run. A recovery run can therefore print "the daily price job FAILED on \<yesterday\>" after succeeding, and a run that dies at `backfill` can leave an all-clear until the next fire's step 0. The **lost-day half — the actual point of #357 — is unaffected**, because it derives live from the durable log. Step 0 already delivers the job half strictly fresher, which is why #376 exists rather than a patch here: every remedy is a design decision on a converged grill. Note this is the *same shape* as the bullet above — both are cases where the job half can read healthy while something is wrong — and the two are deliberately recorded together rather than one being mistaken for the other.

## Verification

- **`shellcheck` 0.11.0 — exit 0** on the wrapper, and `bash -n` clean under `/bin/bash` 3.2.57, which is what launchd actually runs.
- **The wrapper harness armed on its own**, with no override, on an uncommitted wrapper edit: **56 passed (56)**.
- **Every test added here was mutation-proved.** Each mutation was applied to real source, observed red, and reverted. Across the composer, the cap, step 0 and the arming fix that is 21 mutations and 21 reds. The ones worth naming:
  - Making the writer skip the empty write turns the self-clearing tests red — the notice's healthy state is an actively truncated file, not an absent one.
  - Lowering the cap below the realistic outage shape turns the realistic-shape guard red, so the cap cannot be tightened into hiding the debt without the suite objecting.
  - Restoring the `.trim()` bug turns the new tracked-wrapper test red **while the untracked test stays green** — direct confirmation the old test could not catch it.
  - Relocating step 0 to after the `command -v pnpm` check turns the broken-toolchain case red, which is the case that gives step 0 its reason to exist.
- **The notice cannot dirty the ledger repo.** `operator-notice.txt` lands in the data dir, which in the deployed configuration is the ledger repo's `data/`. That repo ignores `/data/*` and re-admits a named allowlist; the notice is not on it, so it falls through exactly as `gap-report.json` and `job-heartbeat.json` already do — confirmed with `git check-ignore`, all three ignored while `events.jsonl` stays tracked. The daily run therefore cannot dirty the repo or trip the wrapper's step 4 post-check.
- **Step 5b's wiring is demonstrated, not asserted — and the residual gap is stated exactly.** Six harness cases drive the REAL wrapper through step 5b, twelve consecutive runs each — case 1 (healthy), case 7 (hang), both halves of case 6 (cry-wolf at `backfill`), and both halves of case 8 — each asserting via `pnpmReached` that `pnpm operator-notice` was invoked. Case 6 goes further and pins the exact order with `invocationOrder(...).toEqual([...WRAPPER_PNPM_COMMANDS])`, so the notice step is fixed BETWEEN `gap-report` and `backfill`: a notice step that drifted to the far side of the networked `backfill` goes red rather than shipping. **What no harness case asserts is the CONTENT of `operator-notice.txt` as written at 5b**, because the harness fakes `pnpm` wholesale and the real CLI never executes inside it — the writer's output is covered by the unit tests and by the read-only measurement against the real store instead. That split is a property of the harness's design, and it is named here so no reader assumes end-to-end content coverage that does not exist.
- Tests **drive** the constants rather than restating them; no test hard-codes the cap.
- **The two new `@numisma/event-store` files measure 100%** — statements, branches, functions and lines, for both `operator-notice.ts` and `operator-notice-io.ts` (`vitest run --coverage`, scoped to those files). `docs/coverage-rationale.md` §7's completeness contract therefore gains no sub-100% unlisted file from this merge. Measured here so #377's owner does not have to.
- Every fixture is authored — dates and counts only, no prices, positions or balances, nothing seeded from a real run.


AmetAlvirde:
**HIGH — the notice's job half is always one run stale, and it can write an all-clear over a run that then fails**

`ops/price-feed/run-daily-fetch.sh:902-903` (step 5b) · `packages/event-store/src/operator-notice-io.ts:52-62`

`loadOperatorNoticeLines` reads `job-heartbeat.json` (`loadHeartbeatLines`, line 58). That file is written by **one** writer only — `write_heartbeat`, installed as the `EXIT` trap at `run-daily-fetch.sh:441`. So at step 5b the breadcrumb on disk describes the **previous** run, never the one composing the notice.

This is the one place the notice is *not* the banner's twin. `apps/tui/src/liveness-lines.ts` evaluates the identical composition on demand, so its heartbeat half always describes the last *finished* run. Step 5b evaluates it mid-run, one outcome too early.

**The bad direction — a written all-clear over a failed run.** Steps 1-5 succeed, so step 5b runs. The previous run was clean → `loadHeartbeatLines` returns `[]`. The window is clean (today is below `dueThrough`'s ceiling; yesterday just got marked) → no gap lines. `writeOperatorNoticeFile` (`operator-notice-io.ts:99-106`) therefore **truncates the notice to empty** — the channel's healthy state. Then step 6 `pnpm backfill` fails: a Neon outage, or the missing `PROJECTION_WRITE_DATABASE_URL` this file's own comment at line 916 says throws immediately. The run exits non-zero at `backfill`.

The heartbeat now says `exit 1 at step 'backfill'`. The TUI banner would print it. **The notice is empty, and empty means healthy.**

Step 0 is not the backstop here, because step 0 only runs when the *next* run starts. The plist fires 18:00-23:00, so a failure on the **23:00 fire** has no successor for ~19 hours — the notice stays silently empty across the entire next morning, which is exactly the window `cat`-on-new-terminal exists to cover. This is the failure direction the PR body names as the worst one, reached from inside the fix.

**The mirror direction — cry-wolf.** Previous run failed at `backfill`, the 23:00 fire succeeds end to end. Its step 5b reads the *previous* (failed) heartbeat and writes `Numisma: the daily price job FAILED on … — exit 1 at step 'backfill'`. Its own `EXIT` trap then records `exit 0 / complete`. The notice standing overnight reports a failure the job has already recovered from, while the TUI, reading live, is silent. Same channel, two surfaces, contradictory verdicts — and the credibility spend the whole PR is written to avoid.

Not the out-of-scope "silently-absent job": in both cases the run **ran**, and the notice actively wrote the wrong thing.

Fix directions (all inside the existing design, no second LaunchAgent): compose the job half from *this* run's in-flight state (`LAST_STEP` / current status) rather than from the on-disk breadcrumb; or move the notice write to after the heartbeat is on disk (the `EXIT` trap path, which already runs pure bash after step 6); or have step 5b write only the gap half and let a step-0-shaped bash write own the job half on the next start. Whichever is chosen, the invariant worth pinning in a test is: **a run that fails after step 5b must not leave an empty notice.** No case in `run-daily-fetch.test.ts` covers a run that reaches 5b at all — all four case-9 tests deliberately end before it — so nothing currently fails on this.

AmetAlvirde:
**MEDIUM — the narrow-window test defect the PR narrates is still open: no test drives the zero-argument shape production actually runs**

`packages/event-store/src/operator-notice-io.test.ts:74` (and every use of it: 95, 104, 117, 126, 134, 144) vs `apps/price-feed/src/operator-notice-cli.ts:37`

The PR body's own account of how the 90-line notice shipped:

> Every unit test passed an explicit narrow window. The real CLI passes none, so production gets the derivation's ~90-day default.

That is still true of this PR. `const WINDOW = { since: "2026-08-15", until: "2026-08-16" }` is passed to **every** call in `operator-notice-io.test.ts`. The production entry point calls `writeOperatorNotice(resolveEventStorePaths(resolveDataDirDefault()), { now: new Date() })` — no `window` at all, so `operator-notice-io.ts:113` falls through to `{}` and the derivation's default floor (`boundedEraFloor`) applies. **Nothing in the suite exercises that call shape.**

The cap tests in `operator-notice.test.ts` do not close it either: they hand-build `GapReport` objects (`report()`, line 18) and call the pure `formatOperatorNotice` directly. So `MAX_NOTICE_LOST_DAYS` is proven against synthetic reports and never against the path that produced the bug — a broken `?? {}` default, a CLI regression that passed a wide explicit window, or a future change that applied the cap upstream of the composer would all stay green.

Concretely, `it("treats an ABSENT log as a window with no anchors, not as a broken check")` (line 140) asserts `expect(lines.filter(l => l.includes("NO DATA"))).toHaveLength(2)`. Two is a property of `WINDOW` being two days wide. Run the *same* fixture the way the CLI runs it and the correct answer is 10 enumerated days plus a `N earlier lost day(s) withheld` tail — the exact shape the cap exists to produce, and the exact shape the PR says was found by executing the deliverable rather than by inspecting it. That assertion is the closest thing here to the smoke run, and it is pinned to the one window production never passes.

Suggested closure: one test in `operator-notice-io.test.ts` that calls `writeOperatorNotice(paths, { now: NOW })` against a store with no log, and asserts the written body has `MAX_NOTICE_LOST_DAYS` recovery lines plus one withheld tail — driven off the exported constant, not restated. That is a handful of lines and it turns the smoke run into a regression.

AmetAlvirde:
**MEDIUM — "the two MUST agree on the path" is not true: `$ENV_FILE` can split step 0 and step 5b into different directories**

`ops/price-feed/run-daily-fetch.sh:91-99` (the new comment), `:59` (`DATA_DIR`), `:673-685` (the `set -a` source)

The new comment on `OPERATOR_NOTICE_FILE` states the invariant this channel depends on:

> Step 0 below writes it in bash and step 5b rewrites it through `pnpm operator-notice`, and the two **MUST** agree on the path — which is why there is no `NUMISMA_*` override for it and must never be one.

There *is* an effective override, and it is the one the wrapper itself installs. `DATA_DIR` is resolved at line 59, at the top of the file. The private token file is sourced at lines 675-682 under `set -a`, i.e. **every assignment in it is exported**. So a `NUMISMA_DATA_DIR=/Volumes/ledger/data` line in `~/.config/numisma/price-feed.env` produces:

- bash `$DATA_DIR` = `$HOME/Dev/accumulus/data` (resolved before the source) → step 0 writes `operator-notice.txt` and the `EXIT` trap writes `job-heartbeat.json` **there**;
- `pnpm operator-notice` at step 5b → `resolveDataDirDefault()` sees the exported `NUMISMA_DATA_DIR` → `operatorNoticePath()` resolves to `/Volumes/ledger/data/operator-notice.txt`.

Two writers, two directories, one `cat` in the profile — and per the comment's own last sentence, "for a channel whose healthy state is an empty file, that failure is not a missing message: it is silence that reads as health."

This predates the PR for `job-heartbeat.json` and `gap-report.json` (both are also written by post-source node steps against a pre-source bash `$DATA_DIR`), but the PR is what makes it load-bearing and what asserts in a comment that it cannot happen. Also note step 3/4 already run `git -C "$DATA_DIR"` against the pre-source value, so the split is wider than the notice.

Cheapest honest fixes: re-resolve `DATA_DIR` (and the two derived paths) immediately after the `source`, or refuse the run if `NUMISMA_DATA_DIR` changed across the source, or soften the comment to say what actually holds. Silently disagreeing is the one option the comment rules out.

AmetAlvirde:
**MEDIUM (docs) — the "third reader" section misses the divergence that will actually happen: launchd does not see your shell's `NUMISMA_DATA_DIR`**

`docs/price-feed-ops.md:622` (the snippet), `:644` (the export bullet), `:666-680` ("Two ways the bash expansion and the engine's rule actually differ today")

The section is careful and the framing is right, but the enumerated list is of *value-format* divergences (a `~/` prefix, a relative path). The divergence that fires in practice is an *environment-scope* one, and it is not mentioned:

The profile snippet expands `${NUMISMA_DATA_DIR:-…}` **in an interactive login shell**. The wrapper runs under **launchd**, which starts the job with a bare, non-login environment — this file's own install section and `run-daily-fetch.sh:44-54` are built entirely around that fact for `PATH`. `NUMISMA_DATA_DIR` is no different: a value exported from `~/.zshrc` — the very file this section tells the operator to edit, and the natural place to put it — is **invisible to the scheduled run**.

Result: the wrapper writes `operator-notice.txt` into `$HOME/Dev/accumulus/data`, the profile `cat`s `/Volumes/ledger/data/operator-notice.txt`, which never exists. `[ -s … ]` is false on every new terminal, forever. That is precisely the section's own stated failure — "the notice is written to one directory and `cat`-ed from another … It looks like a clean machine" — arrived at through the configuration the page most encourages.

The bullet at :644 ("If you set `NUMISMA_DATA_DIR`, **export it before this block runs**") reads as if exporting in the profile is sufficient, which makes it worse than silent: it points the operator at the half of the problem that does not matter here.

The stated mitigation at :682-684 ("set `NUMISMA_DATA_DIR` to an absolute, already-expanded path … or leave it unset") does not cover it either. The complete rule is: a non-default `NUMISMA_DATA_DIR` must **also** be set in the LaunchAgent plist's `EnvironmentVariables` (and, given the `set -a` source, see my separate note on `$ENV_FILE`), or left unset everywhere.

Worth a fourth row in the table too — the plist is a real reader of this default and is not listed.

AmetAlvirde:
**LOW — `process.exit(0)` can truncate the one diagnostic the CLI header calls load-bearing**

`apps/price-feed/src/operator-notice-cli.ts:40` and `:48`

The header says of a disk failure:

> That one case goes to stderr, into the run log the wrapper tees, and the run continues. This is the only channel left at that point, and saying nothing would be the failure this whole increment exists to remove.

But line 48 calls `process.exit(0)` synchronously after `console.error`. Under the wrapper, stdout and stderr are both a pipe into the `tee` process substitution (`run-daily-fetch.sh:469`), and Node's writes to a **pipe** are asynchronous — `process.exit()` terminates without flushing pending writes. So the message that is described as "the only channel left" is exactly the one that can be dropped, and it is dropped only when it matters (a real `writeFile` failure), never in an interactive test run against a TTY, where writes are synchronous.

Both `process.exit(0)` calls are also unnecessary. The rejection is already handled by the second argument to `.then`, so the promise never goes unhandled and the process exits 0 on its own once the event loop drains. Deleting both lines gets the flush *and* keeps the exit-0-on-purpose contract, which the handled rejection already guarantees.

If an explicit exit is wanted for clarity, `process.exitCode = 0` (rather than `process.exit(0)`) is the flush-safe form.

AmetAlvirde:
**LOW (judgment call) — step 0 discards a still-accurate lost-day enumeration, and the trade is not recorded anywhere**

`ops/price-feed/run-daily-fetch.sh:299-303`

Step 0 writes its two lines with `>`, replacing the file wholesale. The comment at :279-286 argues carefully — and correctly — that step 0 must not *truncate* on a clean previous run, because the standing notice "may name real lost days". But on the failing branch it does something with the same consequence: it deletes those lines.

Concretely: run N-1 completes and its step 5b enumerates `2026-08-14` and `2026-08-15`, each with its `pnpm prices:fetch --as-of=…` command. Run N dies at step 3. Run N+1's step 0 sees the failure and overwrites the file with the two bounded lines — the dated recovery commands are gone. If run N+1 also dies before 5b (a provider outage takes all six evening fires the same way), the operator's terminal now shows only "the previous daily price job run FAILED", with **no lost day named and no command to run**, for the whole outage. Lost days are permanent, so the discarded lines were still accurate the entire time.

The scope line is honest — it says the file "says nothing about lost days yet" — so this is not a false claim, and I can see the argument that one bounded, current message beats a mixed file. But it inverts the channel's stated purpose (the standing, self-extinguishing debt) in exactly the multi-day-outage case that purpose was written for, and unlike every other trade in this PR it is not recorded in the code, the runbook or the PR body.

If the current behaviour is intended, a sentence at :299 saying so would match the standard the rest of this file sets. If not, prepending the two lines to the existing body (read it before the `>` — still pure bash, still guarded) preserves both, at the cost of the file no longer being a single `printf`.

Flagging as a judgment call rather than a defect; not re-litigating the settled "step 0 is a pure equality read" decision, which this does not touch.

AmetAlvirde:
**Checked and clean — recorded so the next reader does not re-do it**

**`git()` / `parsePorcelain` (`apps/price-feed/src/wrapper-harness/arming-trigger.testkit.ts:230-251`) — the fix is correct, including the shapes the comment does not name.** I built a fixture repo and checked the real producer's output rather than reasoning about it:

```
·M·"a·file·with·spaces.txt"
·M·ops/price-feed/run-daily-fetch.sh
R··zz.txt·->·"renamed·with·space.txt"
??·"trailing·space·"
```

- `.replace(/\s+$/, "")` has no `m` flag, so `$` anchors at end-of-input only: it can never touch the leading space of *any* line, not just the first. The reported bug is fully closed, not narrowed.
- ` M`, `M `, `??`, `R  old -> new` and `UU` all put the path at `slice(3)`; `parsePorcelain` is right for each.
- Renames: git leaves the old path bare and quotes the new one when needed; `indexOf(" -> ")` + the quote strip lands on the new path in both forms.
- **The one case I expected to break it does not**: a path with trailing whitespace would be eaten by `\s+$` on the last line — but git *quotes* such a path (`?? "trailing space "`), so the final character is `"`, never whitespace. Same for embedded spaces. The trailing-whitespace strip is therefore safe against every path git can emit here.
- No regression for the other callers: `rev-parse`, `merge-base` and `diff --name-only` never emit leading whitespace, so `.replace` and `.trim` are identical for them.

Two pre-existing, non-blocking notes, mentioned only so they are not mistaken for new: `parsePorcelain` does not decode C-style escapes inside quoted paths (`\303\251`, `\t`), and `resolveTriggerFacts`' committed half (`:304`) does not strip quotes at all — both irrelevant to an all-ASCII `TRIGGER_PATH_SET` and neither introduced here.

**The wrapper's bash.** Independently reproduced the PR's verification claims: `shellcheck` 0.11.0 exits 0, and `/bin/bash -n` under 3.2.57(1)-release is clean. Step 0's two `sed` extractions were run against both the pretty-printed heartbeat the wrapper's `printf` emits *and* the single-line `JSON.stringify` shape the io test seeds, plus a `timeout:backfill` decorated step name — correct in all three, and failing closed (empty → silent) on a non-numeric or absent field, as documented. The `|| VAR=""` guards, the `-f` gate, the single-`printf` overwrite and the placement ahead of `resolve-tools` and ahead of the `EXIT` trap install are all right; `set -euo pipefail` cannot be tripped by the block.

**`MARKS_LANDED_STEPS`** (`:164`): adding `operator-notice` is correct — the step runs after `spine`, so a run that dies there has landed the day's marks and must still stamp.

**The cap** (`packages/event-store/src/operator-notice.ts:182-191`): the index zip is sound. `formatLostDays` maps `report.lost` 1:1 (`gap-report.ts:398-404`) and `computeGapReport` builds `lost` ascending by construction (`gap-report.ts:325`), so `index < withheld` really does withhold the *oldest* and keep the most recent, and `withheld < report.lost.length` always holds — the notice can never go empty while lost days exist. `venueDark.length` is a venue-day count, matching `formatGapSummary`'s wording exactly.

`lines[index] as string` on `:188` is the one thing I would tighten: the cast is what would let a future change to `formatLostDays` inject `undefined` into the notice silently instead of failing to compile. Cosmetic, not filed separately.

AmetAlvirde:
## Review disposition — every finding answered

Thanks — this was a good review. Five findings fixed, one escalated to the repo owner rather than patched, and the clean-record comment acknowledged. Nothing is being left silent.

### 1 — HIGH: the notice's job half is always one run stale — **ESCALATED, NOT FIXED**

Confirmed independently before acting on it. `job-heartbeat.json` has exactly one writer, `trap write_heartbeat EXIT` at `run-daily-fetch.sh:441`; step 5b runs at `:902`; and `formatHeartbeatWarning` (`heartbeat.ts:251`) emits the FAILED line on `exitCode !== 0` with **no staleness gate**. So both mirrors you describe are real:

- **Cry-wolf**: a recovery run that succeeds still composes against the *previous* failed run and writes "the daily price job FAILED on \<yesterday\>", which then stands for ~24h after the run that actually fixed things.
- **False all-clear**: a run that reaches 5b with clean data and then dies at `backfill` leaves an empty (= healthy) notice until the next run's step 0, ~19h later on the 23:00 fire.

You are also right that no test covers a run reaching 5b at all — cases 9a–9d all end before it.

This is **not** being patched in this PR, deliberately. Every candidate remedy is a design decision on a converged grill: writing an in-progress heartbeat before 5b reaches into #191's mark-window logic; dropping the heartbeat lines from 5b changes what the channel says; recording it and shipping accepts a day of cry-wolf on the recovery path. That is the repo owner's call, not a reviewer-fix cycle's, so it has been escalated with the mechanism above rather than resolved here.

### 2 — MEDIUM: the narrow-window gap was still open — **FIXED**

You were right, and it was the sharpest finding here: the PR narrated this defect class and then left an instance of it in its own tests. `operator-notice-io.test.ts` now has `writeOperatorNotice, called the way the CLI calls it`, which passes **no `window` key** — the CLI's exact argument shape — against an authored store, takes the real default window, and produces 43 lost days through real files and the real composer. It asserts the cap engages, the withheld tail appears with a count, the newest lost day is enumerated with its command, and the era floor is not named. The cap constant is imported and driven, never restated. Mutation-proved three ways (cap removed → red at 43 vs 10; tail dropped → red; slice direction inverted → red).

### 3 — MEDIUM: the comment asserted something false — **FIXED (comment only)**

Correct catch. `DATA_DIR` resolves at `:59`, before `$ENV_FILE` is sourced under `set -a`, so a `NUMISMA_DATA_DIR=` in the token file really would split step 0 from step 5b. The false clause is gone; `:91-116` now states the ordering plainly, names the resulting split, and names the one supported configuration.

**The resolution order was deliberately not changed** — line 59 belongs to another lane's change and reordering it is a behaviour change this PR should not smuggle in. Comment only, on purpose.

### 4 — MEDIUM (docs): the divergence that actually fires was missing — **FIXED**

Agreed, and this was the one most likely to bite in practice. `docs/price-feed-ops.md` now leads with the environment-scope divergence — launchd runs a bare non-login environment and never sees an export from `~/.zshrc`, the very file the section tells you to edit — marks it as the one that fires, adds the LaunchAgent `EnvironmentVariables` row to the table, fixes the `:644` bullet that pointed at the wrong half, and rewrites the mitigation to cover both halves.

### 5 — LOW: `process.exit(0)` can truncate the diagnostic — **FIXED**

Both explicit exits removed. Exit-0-on-failure behaviour is unchanged — it already came from the handled rejection — and the header now records why neither arm may call `process.exit`: the write is an async pipe into the wrapper's `tee`, so forcing the exit would drop precisely the diagnostic the header calls the only channel left, and only in the deployed shape.

### 6 — LOW (judgment): step 0 discards a still-accurate enumeration — **TRADE RECORDED, behaviour unchanged**

Taking this the way you filed it: the behaviour stays, the trade gets written down. The comment beside that branch now walks the multi-day outage explicitly, notes that lost days are permanent so nothing discarded had gone stale, and gives the reason it is still the right call — prepending would carry undatable text forward with nothing able to retire it, and only step 5b knows whether a lost-day line is still true. Named as an accepted cost rather than left implicit.

### 7 — the verified-clean record — **acknowledged, no action**

Particularly useful: building a real fixture repo to check the `.replace(/\s+$/, "")` fix against ` M`, `M `, `??`, `R  old -> new`, quoted paths and `--untracked-files=all` is exactly the kind of check that would have caught the original defect. The note that git quotes paths with trailing whitespace, so the last character is `"`, is the detail that makes the fix provably complete rather than narrowed.

Not actioned, and agreed as out of scope: the C-style escape decoding in `parsePorcelain`, the quote-stripping asymmetry at `resolveTriggerFacts:304`, the `lines[index] as string` cast, and `REALISTIC_LOST_DAYS` being literal.

AmetAlvirde:
## Correction to my own disposition, and the escalation issue

**Escalation filed: #376** — the HIGH finding, recorded with mechanism, the narrowed blast radius, and three candidate remedies. Not built here, deliberately.

### I repeated a claim that turns out to be wrong

In my disposition above I wrote *"You are also right that no test covers a run reaching 5b at all."* **That is incorrect, and I should not have accepted it without checking.**

Six harness cases drive the **real wrapper through step 5b**, twelve consecutive runs each, each asserting via `pnpmReached` that `pnpm operator-notice` was invoked:

- `case 1 — healthy run`
- `case 7 — a HANG under a watchdog disabled for not leading its group`
- `case 6 — cry-wolf: a timeout at backfill IN the mark window`
- `case 6 — the same cry-wolf run OUT of the window`
- `case 8 — a schemaVersion-1 file's finishedAt is CARRIED`
- `case 8 — an in-window v2 file that died before spine`

Case 6 goes further and pins the exact order — `invocationOrder(...).toEqual([...WRAPPER_PNPM_COMMANDS])`, under the comment *"The sentinels are a set; the wedge is a position"* — so the notice step is fixed **between `gap-report` and `backfill`**. A notice step that drifted to the far side of the networked `backfill` goes red rather than shipping. The wiring is demonstrated, not asserted.

What is true is narrower, and it is now stated in the PR body: **no harness case asserts the CONTENT of `operator-notice.txt` as written at 5b**, because the harness fakes `pnpm` wholesale, so the real CLI never executes inside it. The writer's output is covered by unit tests and by the read-only measurement against the real store. That gap is a property of the harness's design, and it is carried into #376 rather than left for a reader to assume.

Cases 9a–9d ending before 5b is true and remains true — but those are the step 0 cases, and they were never the ones that would have covered this.

Issue #379, opened 2026-08-18
docs(ADR-017): status says the implementation "does NOT land" — it landed, and the ADR now describes the opposite of the code


AmetAlvirde:
`context/adr/ADR-017-a-close-is-the-positions-last-dated-event.md:10-13`:

> _Status: accepted — decision taken by the fund's operator 2026-08-11. **The
> implementation does NOT land with this ADR.** The code today does the opposite,
> deliberately and under test; this ADR authorizes the change and names what it costs._

It landed. `packages/engine/src/position-seal.test.ts:766` reads
`describe("case C is admitted, the backdated trim ADR-017 relaxed", …)` and `:773` is
`it("admits the backdated trim, reporting neither the already-closed nor the seal message", …)`.

A reader consulting ADR-017 today is told the code does the opposite of what it does.
That is the most consequential single instance found by #349's grep pass, and it is why
it is filed rather than patched: the status field, the pin narrative, the verb table and
`context/adr/INDEX.md:21`'s summary column all need rewriting to a post-landing tense.

Stale citations inside the same document, verified:

- `:6-7`, `:220` cite `position-seal.test.ts:753` for the case-C pin — `:753` is now
  mid-assertion inside a different block; the describe is `:766`, the it is `:773`.
- `:121`, `:228` and two test files (`position-backdated-trim.test.ts:290`,
  `position-backdated-add-to.test.ts:199`) cite **`crossref.ts:819`** as "the seal rule's
  own strictness". `crossref.ts:817-820` contains no comparison at all — it is prose about
  ADR-015's third delta class. The `>=` is specified at `crossref.ts:163` (*"reject when
  `event.asOf >= closedAsOf`"*) and enforced at `:1197` and `:1335`. See the section below
  for why these three were pulled out of #349 rather than fixed there.
- `:124` cites `crossref.ts:608-615` for "breach is derived per OPEN position"; that
  claim lives at `:143-144`/`:766`, while `:608-615` is where `positionLots` is built —
  the thing `:173` cites `crossref.ts:459` for.
- `:175` cites `crossref.ts:1054-1063` as "the position-lot-sufficiency gate"; it is
  `checkDebit`, a **reserve-balance** check.
- `:122` cites `crossref.ts:1117` for the `PositionAddedTo` row; it is the seal-check
  ordering comment inside the `PositionClosed` path.
- `:124` (`crossref.ts:625`, `InvalidationMarked`) and `:123`/`:78` (`crossref.ts:939`,
[...]

Issue #378, opened 2026-08-18
docs(ADR-004): the "three stale claims" passage asserts a comment that no longer exists


AmetAlvirde:
`context/adr/ADR-004-preferences-sidecar.md:204-212` says `preferences.ts:~39` still
reads *"latent today (no runtime caller), but becomes silent split-brain the moment the
sidecar is wired into the read path"*, states *"That moment has already passed"*, and
records that **"the comment itself is left untouched"**.

The comment was rewritten. `packages/preferences/src/preferences.ts:44-53` now reads:

> The read path is ALREADY WIRED — `apps/web/src/push/push-core.ts` calls
> `loadPreferences(resolvePreferencesPath())` live, via `loadReserveFloorAsOf` →
> `buildGlanceForAnchor` — so the split-brain hazard this resolver prevents is live,
> not hypothetical.

The quoted "latent today" string does not exist anywhere in the repo. `context/adr/INDEX.md:8`
already carries the corrected account (*"spec #320 was that change and the debt is PAID"*),
so the index and the ADR body now disagree.

All four line numbers in the same passage are also stale:

| cited | actually |
| --- | --- |
| `push-core.ts:130` (`loadPreferences(resolvePreferencesPath())` in `loadReserveFloorAsOf`) | `:202`, in the function declared at `:199` |
| `push-core.ts:142` (`buildGlanceForAnchor`) | `:250` |
| `push.ts:73` | `:95` |
| `backfill-core.ts:175` | `:271` |

**Why this is not a renumbering.** The passage's premise — "the comment still says the
latent thing, and we deliberately left it" — is false, so the paragraph needs rewriting
to a post-#320 tense, not new numbers. Follow the practice this ADR already records
elsewhere: *"correction 2's cited line numbers have all moved, so the live chain is now
named by its functions rather than pinned."*

Found by #349's bounded grep pass and split out under that issue's ruling.

Issue #377, opened 2026-08-18
docs(coverage-rationale): §7's table omits three `packages/preferences/src` files that measure below 100%


AmetAlvirde:
`docs/coverage-rationale.md` §7 states its own completeness contract: it accounts for
"every other currently measured file below 100% that the earlier sections do not
individually name."

Measured with `pnpm coverage` at `7269a9f`, three files in `packages/preferences/src`
are below 100% and have **no row anywhere in the document**:

| File | Lines | Uncovered |
| --- | --- | --- |
| `sidecar-io.ts` | 92.22% | `:123-124,171-173,175-176` |
| `plans.ts` | 96.39% | `:157-158,200-201,203-204,298-299,322-323,388-389` |
| `reconciliations.ts` | 86.66% | `:183-187,197-198,200-204,212-215,219-221,223-224,236-238,240,274-275,326-327,347-352,521-523` |

`sidecar-io.ts` is the reason the gap opened: the shared lock + temp + rename mechanics
were extracted out of `orders.ts` into it, taking their uncovered branches along. #349
fixed the `orders.ts` row that was left citing line numbers past the end of a 145-line
file, but deliberately did **not** add these rows — `reconciliations.ts` alone has twelve
uncovered ranges, and accounting for each honestly is not a one-line fix.

Not a coverage-raising task. The ask is the accounting §7 already promises: one row per
file, each uncovered range named and categorized (defensive guard / unreached fixture /
real branch), in the voice the existing rows use.

PR #380, opened 2026-08-18
docs: reconcile docs and comments with the source of truth they cite

### LINKED ISSUE #349: docs and comments that assert a source of truth the code does not read
Collected while running the open-issue board. Each item was found by a lane whose scope did not cover fixing it, and reported rather than smoothed over. Filed together because they are one family: **a doc or comment that asserts something the code does not do.**

## The class, and why it is worth a ticket rather than a tidy-up

Two of the five lanes' single worst findings this run were the same shape — not a wrong behavior, but a **comment that named a source of truth the code did not actually read**, which let a broken guard look authoritative:

- The pg-free module-graph walker's comment said its workspace groups came from `pnpm-workspace.yaml`. They were hardcoded to `packages`/`apps`. Adding `services/*` to that yaml with a pg-importing package in it, the guard **passed on exactly the case it had just been widened to catch.** Fixed in #347.
- A sibling case in test helpers: an `?? ""` data-dir default that reads as "not configured" and silently resolves to the real ledger. Filed separately as #348.

The common failure mode is **silent green**: the guard runs, reports success, and is trusted precisely because it looks like it consulted an authority. A grep for other "follows X" / "derived from X" / "comes from X" comments that do not in fact read X is worth one pass.

## Verified instances

**1. `packages/preferences/README.md` describes two sidecars where there are three.**
Line 3 calls the package *"the access-surface half of two file-IO sidecars"* and line 32 says *"Append-only, **both** sidecars."* The package carries at least three — `preferences.jsonl`, `plans.jsonl`, `orders.jsonl` — and the surface table omits the whole `plans.jsonl` tenant. Pre-existing; predates #320. (A fourth, `reconciliations.jsonl`, arrives with #321, which will make the count wrong again if this is fixed by hardcoding a new number rather than by describing the class.)

**2. `docs/coverage-rationale.md`'s `orders.ts` row is stale.**
Line 556 cites **89.74%** for `orders.ts` (preferences package); measured actual is **94.64%**. Reported by lane D; the percentage itself is not independently re-measured in this issue, so re-run coverage before editing rather than trusting the number here.

## Explicitly NOT an instance — checked and false

A fourth item was reported in the same batch: that `push-core.ts:161` cites `preferences-import-guard.test.ts`, which *"does not exist in the tree."* **This is wrong.** `apps/web/src/preferences-import-guard.test.ts` exists on `main` (94 lines, tracked) and on every branch that reported it missing. The citation is correct and needs no change. Recorded here so the claim is not re-derived by the next reader.

## Also noted, deliberately not folded in

ADR-020 leaves one boundary in prose rather than in a test — discarding an input you were handed, versus refusing at a gate. The first component that sits ambiguously on that line will have to argue it rather than look it up. That is a known and accepted limit of the ADR, not drift, and is listed here only so it is findable.

AmetAlvirde:
Closes #349.

Docs and comments asserting a source of truth the code does not read. U's ruling bounded
this to **two verified instances plus one bounded grep pass**, with anything larger filed
rather than folded in. That held: three issues were filed instead of fixed (#377, #378,
#379), and three tempting one-line fixes were declined on purpose.

## The evidence this PR did not expect to have

The pass found `packages/engine/src/orders/coverage.ts` citing `appendOrders` as
`packages/preferences/src/orders.ts:244` — a file 145 lines long. The citation directly
above it in the same docblock reads:

> `OrderRecord` in `./records.ts`, **cited BY NAME** for the reason the paragraph above
> gives; **that line number had already drifted once**

So the file already knew line pins rot, said so, and then drifted again in the next clause.

It got better. `appendOrders` was measured at **`:143` in a 145-line file**. By the time
the fix was applied it was at **`:145` in a 147-line file** — because this branch's own
earlier commit added two lines to that header. **A line-number citation went stale inside
one lane, within one session, between being measured and being used — by the change that
was fixing stale citations.**

That is the issue's thesis under controlled observation, with a measured before and after,
rather than an argument that citations rot. It is why the house style here is **cite by
name**, and it is the answer to "why drop the numbers instead of correcting them": a
corrected number is a fix with a known expiry.

**Then it happened a second time, to this PR's own fix.** The `orders.ts` coverage row was
corrected to cite `defaultWarn` at `:71-73` — measured, correct at the merge base, and
already wrong by the time it was written, because this branch's own header edit had moved
it to `:73-75`. Review caught it. So the count is now two instances of the defect occurring
inside the single PR that exists to fix it.

**The second one is not fixed by a name, and that distinction is the point.** A coverage
table's uncovered ranges are **measurements, not citations**:

- A *citation* claims "the construct I describe lives at line N." When code moves, the
  number still resolves — to something else, which usually reads plausibly. It rots
  **silently**, and a reader cannot tell.
- A *measurement* claims "at commit C, lines N-M did not execute." Scoped to C it cannot
  become false. It becomes **outdated**, which is visible: the stamp no longer matches the
  head you are reading at.

No name can express "these lines were not executed," so the defense for a measurement is
the as-of stamp, not a symbol. Every re-measured row and paragraph here now carries the
commit it was measured at. That does not stop the numbers drifting; it stops the drift
being invisible, which is the actual defect #349 names.

## Instance 1 — `packages/preferences/README.md`

The issue says the README is wrong by two. It is wrong by four: `reconciliations.ts`
landed after filing — spec #321, shipped by PR #344 in commit `3b3e3a2`, so `packages/preferences/src/` now holds `preferences.ts`,
`plans.ts`, `orders.ts` and `reconciliations.ts` over a shared `sidecar-io.ts`.

**The README does not now say "four."** A README that says four is the same defect with a
fresher number — this class has already falsified one hardcoded count. The load-bearing
sentence:

> **Which members those are is not a number this README states.** `src/index.ts` is the
> enumeration, and it is the only one worth trusting: a member arrives by being exported
> there, and this file goes stale the moment it tries to keep a tally instead of pointing
> at one.

**Why a fifth sidecar cannot falsify it:** no sentence states a total, and no sentence
states an arrival ordinal. The current members appear only inside "what that enumeration
carries today" — scoped by *today* and by *per `src/index.ts`* — so a fifth export makes
that clause incomplete as of then, not false, and the sentence above already points at
where truth lives. Invariants read "class-wide" and "every loader"; the on-disk table is
keyed on `<sidecar>.jsonl.<pid>.<n>.tmp` rather than on `orders`; the tests section states
the colocation rule and hands the reader an `ls` instead of a list. Every surviving numeral
in the file was audited to confirm it is a genuine two-of-something (two lost records, two
processes, a three-way outcome) and not a class count.

Where an ordinal genuinely earns its place it is pinned to a named, closed axis —
"counting DURABLE ARTIFACTS", "the third BY CLASS MEMBERSHIP" — which is ADR-004's own
instruction, *"say which count you mean."* That note (ADR-004:142-155) records that three
counts legitimately coexist and disagree. It is also why `plans.ts` called itself both the
"THIRD member" and a "FOURTH TENANT" nine lines apart: two axes, one contradiction.

The same file also named `src/index.ts` as its source of truth while omitting every
`plans.*` and `reconciliations.*` export from the table that cites it — the issue's defect
in its purest form. Those rows are added, written from each symbol's own docblock.

## Instance 2 — the `orders.ts` coverage row

**The percentage was the smaller half.** Two of that row's three citations — `:181-182`
and `:222-230` — point **past the end of a 145-line file**. The non-`ENOENT` rethrow and
the `open(lockPath, "wx")` lock-contention retry moved into the shared `sidecar-io.ts` and
took their uncovered branches with them. A number that is merely stale is a smaller problem
than a citation that cannot resolve at all.

Re-measured with `pnpm coverage`: **94.64% lines, 90.9% branch, uncovered `:73-75` only**
(`defaultWarn`'s `console.warn` fallback — the one citation of the three that survived).

⚠️ **Measured twice, against two different heads, and it did not move.** First at `7269a9f`,
then again at `8393bf6` after `origin/main` @ `aecab2c` (Lane D / #348) was merged in — D
being the change to this very package that made a re-measurement necessary. **94.64% both
times.** That agrees with the figure quoted in the issue, which the issue itself records was
never independently re-measured. Stated plainly so the agreement is not mistaken for
inheritance: this is a number that was measured, twice, and happens to match.

The neighbouring `preferences.ts` row **did** move across that merge — 98.68% → **98.72%**
lines, 95.16% → **95.45%** branch, uncovered `:225-226` → **`:244-245`** — which is what a
figure looks like when D's change reaches it, and the reason waiting for D was not
ceremony.

The neighbouring `preferences.ts` row was re-measured too (**98.68% / 95.16%, uncovered
`:225-226`**) — and its surviving narrative did not hold. `:225-226` is `readFailureCode`'s
non-`Error` fallback, not the `String(error)` arm the row described; that construct is gone
from the file entirely, and the `?? ""` arm it also cited now measures covered. Rewritten,
not renumbered.

## The `events/fold.ts` paragraph — a citation invalidated by a sibling lane

#329 landed and rewrote the tier allocator, which invalidated
`docs/coverage-rationale.md`'s claim that `events/fold.ts:47-48` is *"the degenerate
zero-cost-lot arm … which attributes the whole delta to `lots[0]?.tier ?? "c1"`"*. Lines
47-48 are now `SKIP_DETAIL` prose strings, and both arms the paragraph described now
measure covered: the zero-cost arm sends the delta to `present[0]`, the canonically first
tier holding a lot, and the no-lots arm returns `[]` as a Discard-Channel discard.

Re-measured: **98.02% lines** (the bullet said 96.91%), **three** uncovered ranges rather
than the four claimed — `:691-692` (the `Withdraw` arm's `reserve-absent` skip, which the
paragraph never named), `:737-740,742-743` (the `_never` latch, formerly `:492-498`) and
`:1045-1046` (`weightedAverageCost`'s zero-quantity guard, formerly `:688-689`). The
`:649-650` branch is no longer uncovered. `parse.ts` and `crossref.ts` figures in the same
bullet were deliberately not touched — outside the ruled scope.

## The bounded grep pass

Fourteen stale citations found; every one opened and confirmed against the tree before any
edit. Eleven were genuinely one-line and are here. Three were not, and are filed.

**`ADR-012:110` is a better finding than the one it was sent to fix.** It pinned a single
number for two different things — `applyReserveDelta` (`fold.ts:175`) and a quoted
*"untiered: amount is the whole truth"* comment (`fold.ts:711-713`, inside `foldEvents`'
`ReserveOpened` arm). So *"documented there"* was **false regardless of the number**: the
citation was not stale, it was conflating two locations, and no line-number fix would have
caught it. Both are now named. That is the class this issue is about, one level deeper than
drift.

Also retargeted or de-pinned: `ADR-012:119` (`fold.ts:367` → the `ReserveOpened` arm),
`ADR-012:120` (`crossref.ts:290-298` → `:551`), `ADR-012:83` (`parse.ts:515-516` → the lot
*construction*, not the arity check; now `parseLots`, `parse.ts:528`), `INDEX.md:8`
(`isObservedAtStamp` at `records.ts:293`, which is `CURRENCIES`; de-pinned),
`INDEX.md:16` (the same `fold.ts:85` claim as ADR-012), `contract.ts:213` and
`ADR-007:319` (both citing `plans.ts:254-258`, the `SkippedPlanLine` prose-only docblock,
for a claim that lives at `:289-292` — two independent citers, one wrong target),
`position-backdated-trim.test.ts:322-323` (`fold.ts:250` → `:200`, `ingest-walk.ts:191` →
`:231`; comments only, no test logic), `event-store/README.md:32` and the preferences
README (both pointing at `orders.ts` for a private `readOptional` that now lives at
`sidecar-io.ts:116`), and `apps/tui/README.md` (an unnamed count axis).

### Three one-line fixes declined on purpose

`position-backdated-trim.test.ts:290`, `position-backdated-add-to.test.ts:199` and
`ADR-017:228` all cite **`crossref.ts:819`** as "the seal rule's own strictness".
`:817-820` contains no comparison — it is prose about ADR-015's third delta class — and the
`>=` is specified at `crossref.ts:163` and enforced at `:1197` and `:1335`. A three-site
retarget, and the easiest change in the sweep.

Declined, because those same comments also assert *"the seal rule ACCEPTS a verb dated
EQUAL to a close"*, and with ADR-017's relaxation now landed that sentence cannot be shown
to still hold. **A citation swap there would launder an unverified claim into looking
freshly checked** — a comment more trustworthy-looking and no more true. That is this
issue's own defect one level up, and it would pass review, because a correct line number is
exactly what a reviewer checks. Filed as part of #379 instead, blocked on the ADR rewrite
rather than on itself.

## Filed, not fixed

- **#377** — `docs/coverage-rationale.md` §7 omits `sidecar-io.ts` (92.22%), `plans.ts`
  (96.39%) and `reconciliations.ts` (86.66%), violating the completeness contract §7 states
  for itself. `reconciliations.ts` alone has twelve uncovered ranges; not a one-line fix,
  and deliberately not half-fixed here.
- **#378** — `ADR-004:204-212` quotes a `preferences.ts` comment that no longer exists and
  records that it was "left untouched". `INDEX.md:8` already carries the corrected account,
  so the index and the ADR body disagree. The premise is false, so it needs rewriting, not
  renumbering.
- **#379** — `ADR-017`'s status says *"The implementation does NOT land with this ADR. The
  code today does the opposite, deliberately and under test."* It landed:
  `position-seal.test.ts:766` is `describe("case C is admitted, the backdated trim ADR-017
  relaxed", …)`. **A reader consulting that ADR today is told the code does the reverse of
  what it does.**

## Out of scope, deliberately

**No release.** This is the docs sweep that preceded v0.8.0 by precedent, and v0.9.0 is out
of scope by ruling: no manifest bump, no tag, no CHANGELOG, no changesets. The board carried
this as "129+ commits overdue"; measured against `origin/main`, it is **182 commits since
`v0.8.0`**. They stay overdue.

Also untouched by ruling: ADR-020's prose-only boundary (a known and accepted limit of the
ADR, not drift), and the citation of `preferences-import-guard.test.ts` from
`apps/web/src/push/push-core.ts` — **that file exists**, tracked, at
`apps/web/src/preferences-import-guard.test.ts`; the issue's claim that it does not is false
and the citation needs no change.

⚠️ One note on that entry, found while auditing this PR's own body for claims taken on
report rather than checked: **the issue records that citation as living at
`push-core.ts:161`. It is at `:175`.** The line number in the issue's own
"checked and false" note has itself drifted — #349's defect inside the record of #349. Left
unfixed here because the citation it describes is correct and needs no edit; recorded so the
next reader is not sent to the wrong line.

## Review round

A fresh reviewer found four defects, all real, all confirmed against the tree before fixing.
Two are recorded above (the `:73-75` drift, and `orders.ts` being 147 lines rather than 145).
The third is the sharpest thing in this PR after the drift itself:

**The added `resolvePlansPath` row said "The three cases".** True at this branch's base, and
false the moment #348 merged — that PR split the fused `unset/blank` arm so a blank override
now **throws** rather than defaulting. The fix is not "four": **#348 deleted the numeral from
`sidecar-io.ts` outright**, which is this PR's own move, arrived at independently, in the file
that owns those arms. So the row drops the count and points at `resolveSidecarPath` as the
enumeration — and `plans.ts`, which picked up a fresh `"The four cases"` in the same merge,
loses it too. A PR arguing that hardcoded counts rot must not ship one.

The fourth (`fold.ts:993-1008` naming neither `splitTierRemoval` nor its zero-`tierTotal`
branch) is folded in, along with the reviewer's consistency observations: `canonical.ts:386`,
`crossref.ts:551` and both `plans.ts:289-292` pins now lead with
`buildReserveTierContributions`, `buildEventReference` and `LoadedPlans`, keeping the number
as a secondary hint.

## Verification

Measured and run at **`8393bf6`**, which merges `origin/main` @ **`aecab2c`** (Lane D / #348)
— the one landed change that touches `packages/preferences`, and therefore the one that
could have moved the figure this PR was told to re-measure.

**Lane S (#375) was still open at measurement time**, and its file list was checked rather
than assumed: `apps/price-feed/**`, `ops/price-feed/run-daily-fetch.sh`,
`docs/price-feed-ops.md`, `package.json`, and four files under `packages/event-store/src`.
**None of them is a file whose coverage figure this PR states**, and none is
`docs/coverage-rationale.md`. So the measurement is **scoped, not stale** — taken against
everything that could affect it, rather than merely against the most recent merge. Those
are different claims and only the second is worth making.

S has since merged (`80f78ca`, in this branch as of `e12f134`) and the suite below was run
against that head. The measurement was **not** retaken: the figures this PR states are for
`packages/preferences` and `packages/engine/src/events`, and S touches neither, so re-running
an instrumented pass would reproduce numbers already measured twice rather than test
anything. That is a scoping judgement, stated so it can be challenged.

(One consequence for #377, noted rather than acted on: S adds `operator-notice.ts` and
`operator-notice-io.ts` under `packages/event-store/src`. S's own lane reports both at 100%,
which would mean §7's completeness contract gains nothing from the merge — reported here
rather than asserted, because this PR did not measure them. Either way it is #377's subject,
not this one's.)

No test was added or changed; the single test-file edit is comment-only — two `//` lines, no
assertion, fixture, name or import moved.

- `pnpm typecheck` — passed, all seven projects.
- `pnpm test` at **`e12f134`** (the head after merging S) — **159 test files passed, 4
  skipped (163); 2433 tests passed, 52 skipped (2485). Zero failures.**
- `pnpm test` at `8393bf6` (before S) — 157 passed / 4 skipped (161); 2408 passed / 48
  skipped (2456). Zero failures. The growth is S's own suite arriving.
- The four skipped files are integration suites that require a real Postgres via
  `NUMISMA_TEST_DATABASE_URL`, unset in this worktree:
  `apps/web/src/client-bundle.integration.test.ts` (2),
  `apps/web/src/projection/provision.integration.test.ts` (9),
  `apps/web/src/push/backfill.integration.test.ts` (6),
  `apps/web/src/push/push-core.integration.test.ts` (4). The remaining skips are individual
  cases inside otherwise-passing files.

⚠️ **CI is the authoritative gate, not this local run** — the main checkout carries on-disk
artifacts no worktree or fresh clone has, and a run with the database URL set will skip
fewer. **CI is green on `06a32d0`**: `test` ×2 SUCCESS, `Vercel` ×2 SUCCESS,
`mergeStateStatus` CLEAN.


docs/coverage-rationale.md
```diff
@@ -552,8 +571,8 @@ individually itemized with its own paragraph. Grouped by package.
...
+| `orders.ts` (preferences package) | 94.64% lines, 90.9% branch | `:71-73` | Re-measured at `7269a9f`. Two of this row's three original citations pointed **past the end of the file** — `orders.ts` is 145 lines — because the code they named moved out: the non-`ENOENT` rethrow on the sidecar read (cited `:181-182`) and the lock-contention retry path in the advisory file lock (cited `:222-223,225-227,229-230`, `open(lockPath, "wx")` racing another writer) now live in the shared `sidecar-io.ts` and are not in this file at all. What is left is the row's first citation, at its new lines: `:71-73` is `defaultWarn`'s `console.warn` fallback when no `warn` is injected. |
```
AmetAlvirde: **Stale citation, introduced by this PR — `:71-73` is not `defaultWarn` any more.**

At this PR's own head (`16b9169`), `packages/preferences/src/orders.ts` reads:

```
71: }
72:
73: function defaultWarn(message: string): void {
74:   console.warn(message);
75: }
```

`defaultWarn` is `:73-75`. `:71-73` was correct at the merge base `7269a9f` — and this branch's own edit to `orders.ts` (+5/−3 in the header docblock) shifted the file by +2, so the row's citation went stale between the measurement and the merge. The row this replaces said `:73-75`, which is correct again.

This is the same mechanism the PR body documents for `appendOrders` (`:143` → `:145`), landing on a citation the PR is introducing rather than one it is retiring.

**Also in this row: "`orders.ts` is 145 lines" is false at this head — it is 147.** (`git show 16b9169:packages/preferences/src/orders.ts | wc -l` → 147; at `7269a9f` → 145.) The argument the numeral supports still holds — `:181-182` and `:222-230` are still well past the end of a 147-line file — so only the numeral needs correcting. Or drop it: "past the end of the file" is true without a length.

docs/coverage-rationale.md
```diff
@@ -236,19 +237,37 @@ tests moved with the code into `packages/event-store/src/event-store.test.ts`.
...
+  `fold.ts:47-48` is now `SKIP_DETAIL` prose for the `"position-absent"` skip
+  reason. In its place `tierWeightedDeltas` sends a zero-cost-lot delta to
+  `present[0]` — the canonically first Tier holding a lot, in `c1`/`c2`/`c3`
+  order, which is order-independent and invents no Tier — and returns `[]` for
+  no-lots-at-all, a Discard-Channel discard, the file's own comment recording
+  that the old fallback "minted a full-magnitude c1 delta for a cash movement
+  carrying no provenance whatsoever… Not a bad attribution: an invented one."
+  Both arms measure covered. And **`:649-650`, the lot-selection helper's
+  zero-`tierTotal` early return, is no longer uncovered** — the helper still
+  exists (now `fold.ts:993-1008`), it simply measures covered. The
```
AmetAlvirde: **`fold.ts:993-1008` does not name the helper it claims to name.**

At `16b9169` the lot-selection helper is `splitTierRemoval`, `packages/engine/src/events/fold.ts:987-1017` (signature `987-991`, body through `1016`, close `1017`). Its zero-`tierTotal` early return — the branch this sentence is retiring the `:649-650` citation for — is `:996-998`:

```
996:   if (tierTotal === 0) {
997:     return { removed, remaining: lots.map((lot) => ({ ...lot })) };
998:   }
```

`993-1008` matches neither the function nor the branch; it starts inside the body and ends mid-loop. Given the rest of the sweep's house style, the natural fix here is the name: "the helper (`splitTierRemoval`) still exists, it simply measures covered."

docs/coverage-rationale.md
```diff
@@ -206,7 +206,7 @@ archival, magnitude-threshold env plumbing, `parseAsOfArg`); its read-path unit
 tests moved with the code into `packages/event-store/src/event-store.test.ts`.
 
 - **`packages/engine/src/events/*.ts`** (92.56% lines, measured) — the covered
```
AmetAlvirde: **Two vintages in one bullet.** The constituent figure on line 209 is raised to a fresh measurement (`fold.ts` 96.91% → 98.02%), while the directory aggregate on this line (`92.56%`) is left at its old value. If `fold.ts` rose, the `events/*.ts` aggregate rose with it, so `92.56%` is now stale by the same change that fixed its neighbour.

Re-measuring is out of the ruled scope and I am not asking for it. But the bullet now reads as one measurement when it is two, and the aggregate is the number a reader trusts first. A dated qualifier on the aggregate — the way `crossref.ts` carries "re-measured 2026-08-08, after PR #261" a few lines down — would keep it honest without a new run.

Non-blocking.

AmetAlvirde:
## Fresh review — PR #380 (`16b9169` vs base `7269a9f`)

Reviewed with no prior context on the work. Everything below was checked by opening the cited file at the cited line (or finding the named symbol) at this PR's own head, not at the merge base — which is where the one real defect lives.

### Findings

**2 to fix, both in `docs/coverage-rationale.md`, both in the `orders.ts` §7 row** ([inline](https://github.com/AmetAlvirde/numisma/pull/380#discussion_r3805518420)):

1. **`:71-73` is stale at this head.** `defaultWarn` is `packages/preferences/src/orders.ts:73-75` at `16b9169`. `:71-73` was right at `7269a9f`; this branch's own +2 lines to the `orders.ts` header moved it. The row this replaces said `:73-75` — correct then, correct again now. Same mechanism the PR body documents for `appendOrders` (`:143` → `:145`), except this time it landed on a citation the PR is *introducing*.
2. **"`orders.ts` is 145 lines" is false at this head — it is 147.** The argument the numeral carries survives at 147 (`:181-182` and `:222-230` are still past the end), so only the numeral needs fixing, or dropping.

**1 low** ([inline](https://github.com/AmetAlvirde/numisma/pull/380#discussion_r3805519818)): `fold.ts:993-1008` names neither the helper (`splitTierRemoval`, `:987-1017`) nor its zero-`tierTotal` branch (`:996-998`).

**1 non-blocking** ([inline](https://github.com/AmetAlvirde/numisma/pull/380#discussion_r3805521213)): the `events/*.ts` `92.56%` aggregate is left un-re-measured in the same bullet whose `fold.ts` constituent was raised.

### Verdict per check

**1. Factual accuracy of rewritten citations — clean apart from the above.** Every retarget opened and confirmed at `16b9169`:

| Citation | Verified |
| --- | --- |
| `plans.ts:289-292` (`contract.ts:213`, `ADR-007:319`) | ✅ exactly the "empty buckets would assert *this position has no plan*" paragraph |
| `parseLots`, `parse.ts:528` | ✅ `"Position requires at least one lot."` |
| `crossref.ts:551` | ✅ the `lots: undefined` vs `lots: []` reserve comment |
| `applyReserveDelta` early-returns on falsy `lots`; `foldEvents`' `ReserveOpened` arm carries *"untiered: amount is the whole truth"* | ✅ `fold.ts:175` and `fold.ts:712-713` — the ADR-012 split of one number into two named locations is correct |
| `isObservedAtStamp` de-pinned from `records.ts:293` | ✅ `:293` is `CURRENCIES`; the function is `:326` |
| `readOptional` → `preferences/src/sidecar-io.ts` | ✅ `:116`, under the "A DELIBERATE PRIVATE COPY" docblock at `:104` |
| `appendOrders` cited by name | ✅ `orders.ts:145`, and nothing else in the tree pins it by line |
| `fold.ts:200` / `ingest-walk.ts:231` (test comment) | ✅ "(asOf, then log order)" and `[...priorEvents, ...accepted]` |

**2. Count-proofing of `packages/preferences/README.md` — the claim holds.** I hunted specifically for what a fifth sidecar would falsify. No total, no arrival ordinal, no "both"/"neither"/"either", no table implying completeness. Every surviving numeral is a genuine two/three-of-something: two lost records (`:67`), two concurrency axes (`:117`), the loader's three-way outcome (`:43`), `resolveSidecarPath`'s three cases (`:46`), `LOCK_TIMEOUT_MS` 10s (`:80`), one preferences writer (`:41`, `:69`). The Public-surface table is complete against `src/index.ts` — all 22 exports present, none invented — and `:33-34` names `src/index.ts` as the tiebreaker.

Two enumerations would be *incomplete* under a fifth member, not false, and each sits under a sentence that is explicitly class-wide: `:61-62` ("`orders.ts`, `plans.ts` and `reconciliations.ts` all append through it", under "**Append-only, class-wide.** No writer here ever destroys prior history") and `:91-93` (the four `loaded`/`load-failed` spellings, under "Every loader also distinguishes…"). Recording them, not raising them.

Every invariant in the file also verifies against source: `sidecar-io.ts` holds the lock/temp/rename/BOM/torn-terminator mechanics, `orders.ts:146`, `plans.ts:583` and `reconciliations.ts:536` all route through `appendSidecarLines`, `preferences.ts:390` is the one inline `appendFile`, the four `*-reliable.test.ts` files and `sidecar-io-append-failure.test.ts` exist, and `package.json` depends on `@numisma/engine` alone.

**3. Coverage prose vs. source — exact, except the `orders.ts` row above.** I did not re-run `pnpm coverage`; I read the cited lines.

- `fold.ts:691-692` → the `Withdraw` arm's `recordSkip(event, order, "reserve-absent")` ✅
- `fold.ts:737-740,742-743` → the `_never` latch ✅
- `fold.ts:1045-1046` → `weightedAverageCost`'s zero-`totalQuantity` `return 0` ✅
- `fold.ts:47-48` → `SKIP_DETAIL` prose for `"position-absent"` ✅, and `tierWeightedDeltas` really does return `[]` at `:118` and `present[0]` at `:129` ✅
- `preferences.ts:225-226` → `readFailureCode`'s `return "unknown-read-error"` non-`Error` fallback ✅; the `?? ""` arm really is at `:251` now ✅

**4. Scope discipline — clean.** 12 files, docs and comments only. No manifest bump, no tag, no CHANGELOG, no changeset. None of the deliberately-filed exclusions were half-taken: no `crossref.ts:819` retarget anywhere in the diff, `ADR-004:204-212` untouched, no `ADR-017` file in the changeset, `INDEX.md` touches exactly two rows (ADR-004 and ADR-012) and leaves the ADR-017 row alone, and §7 gains no `sidecar-io.ts` / `plans.ts` / `reconciliations.ts` rows.

**5. The test file — comment-only, confirmed.** `position-backdated-trim.test.ts` changes exactly 2 lines, both `//` comments, both inside one block. No assertion, fixture, test name, `describe` or import moved. (`position-backdated-add-to.test.ts` is not in this changeset at all, consistent with the declined `crossref.ts:819` retarget.)

**6. Markdown integrity — clean.** Every `INDEX.md` row still carries 6 pipes, including both giant single-line cells this PR edited by substring replacement (ADR-007's 12 is pre-existing escaped `\|` and untouched); the `§7` table rows all carry 5. No broken links, no orphaned list markers, no mangled cell.

### Two consistency observations, no action asked

- **The PR introduces three fresh numeric pins** while arguing that a corrected number is a fix with a known expiry: `plans.ts:289-292` (twice — `contract.ts:213` and `ADR-007:319`) and `crossref.ts:551`. All three are correct today. Finding 1 above is what that risk looks like when it fires inside a single branch, so it seems worth naming that the sweep took the by-name route in some places and the by-number route in others.
- **`canonical.ts:386` is left pinned** in the ADR-012 paragraph this PR rewrote, two sentences after `fold.ts:85` and `crossref.ts:290-298` were de-pinned in the same paragraph. It is *nearly* right: `:386` is `buildReserveTierContributions`'s signature, and the length-0-is-untiered claim is at `:392-393`. Reads like the one the de-pinning pass walked past.

**No merge/approve action taken — comment only.** No tool call was blocked.

packages/preferences/README.md
```diff
@@ -25,75 +41,99 @@ Enumerated explicitly in `src/index.ts`:
...
-| `appendOrders` | function | Genuinely append-only: build the full next image, write to a same-directory unique temp file, then `rename` over the sidecar (crash-atomic), serialized across processes by an exclusive-create lock file. |
+| `appendOrders` | function | Genuinely append-only: build the full next image, write to a same-directory unique temp file, then `rename` over the sidecar (crash-atomic), serialized across processes by an exclusive-create lock file. Those mechanics are shared (`sidecar-io.ts`); what stays here is the only part that is about orders — the canonical serialization of an `OrderRecord`. |
 | `OrderSkip`, `OrdersLoad`, `LoadOrdersOptions` | types | One skipped line `{line, problem, message}`; the loader's three-way total outcome; `{warn?}` injection point for tests. |
+| `resolvePlansPath` | function | Resolve `<dataDir>/plans.jsonl` under the same ADR-006 invariant — absolute and homedir-derived, never CWD-relative. The three cases (unset/blank → the accumulus default, absolute → verbatim, relative → a loud throw) live once in `sidecar-io.ts`'s `resolveSidecarPath`. |
```
AmetAlvirde: **"The three cases" is already false on `main`, and this row is new — it will land wrong.**

This PR's base is `7269a9f`, but `origin/main` is now `aecab2c` (PR #364, *refuse an empty data dir*). At `aecab2c`, `resolveSidecarPath` is:

```ts
if (dataDir === undefined) return join(resolveDataDir(), fileName);
const raw = dataDir.trim();
if (raw === "") throw new Error(`a sidecar data directory must not be empty …`);
if (!isAbsolute(raw)) throw new Error(`… must be an absolute path …`);
return join(resolve(raw), fileName);
```

Four cases, and the split is exactly on the pair this row fuses: **unset → the accumulus default; blank/whitespace → a loud throw**; absolute → verbatim; relative → a loud throw. `packages/preferences/src/plans.ts`'s own docblock was updated in that same PR from "The three cases" to "The four cases" — so this README row is being introduced in the shape the source has already left.

The PR body says the suite runs "once `origin/main` is merged in", so this is catchable before it lands. Suggested: mirror `plans.ts`'s wording, or drop the enumeration and defer to `resolveSidecarPath`'s docstring the way the rest of this README defers to `src/index.ts`.

(The two files do not otherwise collide — `main` touches `plans.ts:65-77`, this PR touches `plans.ts:2-14` — so this is a semantic conflict, not a textual one, which is the kind that merges clean and stays wrong.)

docs/coverage-rationale.md
```diff
@@ -552,8 +571,8 @@ individually itemized with its own paragraph. Grouped by package.
...
+| `preferences.ts` | 98.68% lines, 95.16% branch | `:225-226` | Re-measured at `7269a9f`. Spec #320 made the loader total, so **the non-`ENOENT` rethrow this row originally named no longer exists** — that path is the `load-failed` arm of the returned envelope and is covered, as is the split-denominator guard the original row's other citation pointed at. The row's later `:214`/`:225` pair has gone stale the same way: `String(error)` is no longer how the `load-failed` message is built, and the `?? ""` arm of `lines[index]` (now `:251`) measures covered. What remains is one arm and it is not a behavior: `:225-226` is `readFailureCode`'s `return "unknown-read-error"` — the non-`Error` fallback, reachable only if something other than an `Error` is thrown out of `readFile`. |
```
AmetAlvirde: **`:225-226` is correct at `16b9169` but goes stale the moment `origin/main` is merged.**

Verified at this PR's head: `preferences.ts:225-226` is `readFailureCode`'s `return "unknown-read-error"` and its closing brace, exactly as described, and the `?? ""` arm really is at `:251`. ✅

But `main` has moved to `aecab2c` (PR #364) and that PR added ~19 lines above this point in `preferences.ts`. At `aecab2c`:

- `readFailureCode` is `:239`, and `return "unknown-read-error"` is `:244`
- `:225-226` is docblock prose ("NEVER `error.message`. Node's fs errors interpolate the ABSOLUTE PATH…")
- the `?? ""` arm the row also cites as `:251` is now `:270`

`preferences.ts` is not in this changeset, so the merge will be clean and silent. Worth re-checking both pins against the merge result before the "measured at" note is finalized — and note the "Re-measured at `7269a9f`" stamp will then name a commit that is no longer the tree the numbers describe.

Same caution applies to the `orders.ts` row's `:71-73`, though that one is already wrong at `16b9169` for a different reason — see the other comment on this file.

AmetAlvirde:
## Addendum — the base has moved, and two more claims are affected

After posting the review above I checked this branch against `origin/main` as it stands rather than against its base. **`main` is now `aecab2c`; this PR's merge base is `7269a9f`.** PR #364 (*refuse an empty data dir*) landed in between and touched `packages/preferences/src/{preferences,plans,sidecar-io,reconciliations}.ts`. The PR body says the suite and coverage results get posted "once `origin/main` is merged in", so this is all catchable first.

Two more findings, both posted inline:

**A. [fix] `packages/preferences/README.md:46` — "The three cases (unset/blank → the accumulus default, absolute → verbatim, relative → a loud throw)" is already false.** ([inline](https://github.com/AmetAlvirde/numisma/pull/380#discussion_r3805534921)) At `aecab2c` `resolveSidecarPath` has **four** arms, split exactly on the pair this row fuses: `undefined` → the default, `""`/whitespace → a **loud throw**, absolute → verbatim, relative → throw. `plans.ts`'s own docblock was rewritten to "The four cases" in that same PR. This row is *added* by this PR, so it introduces a wrong count into the file whose headline claim is that it does not state counts.

**B. [check] `docs/coverage-rationale.md:574` — `preferences.ts`'s `:225-226` is correct at `16b9169` and stale at `aecab2c`.** ([inline](https://github.com/AmetAlvirde/numisma/pull/380#discussion_r3805536675)) #364 added ~19 lines above it: `readFailureCode`'s `return "unknown-read-error"` moves `:225` → `:244`, and the `?? ""` arm the row cites as `:251` moves to `:270`. `preferences.ts` is not in this changeset, so the merge is clean and silent.

**No textual conflict** — `main` touches `plans.ts:65-77` and this PR touches `plans.ts:2-14`; `INDEX.md` collides only if git is unlucky, since `main` edits the ADR-006 row and this PR edits ADR-004 and ADR-012. Both of the above are semantic conflicts, which merge clean and stay wrong.

**Running total: 4 to fix, 1 low, 2 non-blocking.** The three-in-`orders.ts`/`preferences.ts` cluster and finding A share one root cause worth naming: every number in this sweep was measured at `7269a9f`, and three of them have since been invalidated — one by this branch's own commit, two by a sibling that landed on `main`. That is the issue's thesis firing a third and fourth time, on the fix itself.

docs/coverage-rationale.md
AmetAlvirde: Fixed in `06a32d0` / `8660785`, and this is the finding that earned its place in the PR body.

Both corrections applied: uncovered range `:71-73` → `:73-75`, and "145 lines" → 147.

You caught the PR committing its own defect. `:71-73` was correct when I measured it at the merge base; this branch's own header edit then added two lines and moved `defaultWarn` to `:73-75`. That is the second time this exact drift happened inside this branch — the body already documents the first, an `appendOrders` citation measured at `:143` and applied at `:145`.

**Deliberately not converted to a by-name citation, and the distinction is now a section of the PR body.** A coverage table's uncovered ranges are *measurements*, not *citations*. A citation claims "the construct lives at line N" — when code moves it still resolves, to something plausible, and rots silently. A measurement claims "at commit C, lines N-M did not execute" — scoped to C it cannot become false, only visibly outdated. No name can express "these lines did not execute," so the defense here is the as-of stamp, which the row now carries (`8393bf6`).

docs/coverage-rationale.md
AmetAlvirde: Fixed in `8660785`. Now reads `` (`splitTierRemoval`, `fold.ts:987`, its `if (tierTotal === 0)` early return at `:996-998`) `` — name first, line as a secondary hint, matching the pattern used elsewhere in this PR. Verified `export function splitTierRemoval(` at `:987` and the guard at `:996-998`.

docs/coverage-rationale.md
AmetAlvirde: Fixed in `06a32d0`. You were right, and the cause was my own scoping instruction.

I told the implementer to re-measure `fold.ts` but explicitly fenced `crossref.ts` and the directory aggregate out as "outside the ruled scope." That fence produced exactly what you describe: one bullet with a fresh constituent and a stale aggregate, and nothing telling a reader which figure came from which run. That is #349's defect, introduced by the scoping of its own fix.

Re-measured at `8393bf6`: aggregate `92.56%` → **93.85%**, `crossref.ts` `96.23%` → **97.61%**. `parse.ts` was already correct at 83.33%, which is useful evidence that the bullet was not uniformly stale — only the figures a landed change had actually moved. The bullet now names the commit it was measured at.

packages/preferences/README.md
AmetAlvirde: Fixed in `8660785`, and your framing changed the fix.

You are right that the row was new and would have landed wrong. But I did not change "three" to "four": **#348 deleted the numeral from `sidecar-io.ts` outright** (`:23` now reads "The cases, and why each is what it is:"), which is this PR's own de-counting move, arrived at independently, in the file that owns those arms.

So the row drops the count and points at `resolveSidecarPath` as the enumeration, mirroring the README's opening ("Which members those are is not a number this README states"). I also de-counted `plans.ts:69`, which picked up a fresh `"The four cases"` in the same merge — a PR arguing that hardcoded counts rot must not ship one.

docs/coverage-rationale.md
AmetAlvirde: Fixed in `8660785`. Your predicted hazard landed exactly as described.

`origin/main` @ `aecab2c` is now merged into this branch (`8393bf6`), and re-measurement moved the row: `98.68%` → **98.72%** lines, `95.16%` → **95.45%** branch, uncovered `:225-226` → **`:244-245`**. Verified `:244` is `return "unknown-read-error";` and `:245` its closing brace, so the row's narrative survived renumbering.

The stale `(now :251)` parenthetical you flagged alongside it was de-pinned rather than renumbered — it now names the construct ("the blank-line guard in `loadPreferences`'s line walk") instead of carrying a number. Row stamped `8393bf6`.

Issue #367, opened 2026-08-18
Ruling needed (premise retracted): no gate asymmetry — the residual question is the durable-log read path


AmetAlvirde:
> **CORRECTED 2026-08-17, after review of #366.** The premise this issue was filed on is **false**, and the correction is kept in the open rather than quietly edited away. There is **no gate asymmetry**. What survives is a narrower and different question, restated below. See the correction comment for what changed and why.

## The retraction

This issue originally claimed: *the ingest gate admits a lot-less funding leg the fold will discard.* It does not, and it cannot.

Both cited call sites are **post-parse**. `crossref.ts:621` describes itself as the *"Second ingest gate (after `parseEvent`)"* and `:627` notes callers chain `parseEvent` → `crossReferenceEvent`; `crossReferenceOpen` at `:1431` takes an already-parsed `PositionOpenedEvent`. And `parseLots` (`packages/engine/src/events/parse.ts:526-528`) rejects an empty `position.lots` outright — *"Position requires at least one lot."*

So:

- **`crossref.ts:1480`** (`PositionOpened`, called with `position.lots`) — cannot receive an empty array, because no parsed `PositionOpened` carries zero lots.
- **`crossref.ts:1354`** (`PositionAddedTo`, called with `[event.lot]`) — cannot receive an empty array, because a one-element array literal is never empty.

`reserveDeltasForOpen` therefore never returns `[]` at either gate site, no sufficiency check is ever skipped, and the "guard that stops guarding" does not exist on this path. **Do not rule on this hole; it isn't there.**

## What is actually true

The only route that drives `tierWeightedDeltas` to an empty list in a well-formed event is a `PositionTrimmed` whose removals land on tiers the position holds no lots in — `splitTierRemoval` returns `removed: []` when `tierTotal === 0` (`packages/engine/src/events/fold.ts:933`).

That route is **also gate-closed**: `parse.ts:569` requires a positive removal quantity, and `crossref.ts:1245` rejects a removal against an unheld tier on sufficiency. And it is a **credit**, not a debit, so it would never have gone through a `reserveDeltas*` sufficiency check in the first place — sufficiency is a question about money leaving.

## The residual question, which is real and is not about this arm

Every gate above runs at **ingest**. `foldEvents` is also called **directly, with no parse and no crossref**, on the durable log:

- `loadFoldedReview` — `packages/event-store/src/event-store.ts:225` — `return foldEvents(genesis, load.events, asOf);`
[...]

PR #382, opened 2026-08-18
fix(engine): give the discard vocabulary one meaning — a reserve miss discards whole

### LINKED ISSUE #371: Ruling needed: the discard vocabulary is split — reserve-absent half-applies where provenance-absent discards whole
**This is a ruling request, not a fix request. Do not attach it to a lane before it is ruled.**

`FoldSkipReason` is a deliberately closed vocabulary — its doc comment in `packages/engine/src/contracts.ts` says widening it is a spec change, not a build decision. After #366 it has three members, and **two of them mean different things by "discarded".**

## The split

| reason | what the fold does |
|---|---|
| `position-absent` | records the skip, **applies nothing** |
| `provenance-absent` | records the skip, **applies nothing** (#366) |
| `reserve-absent` | records the skip, **and then pushes the closed-book row and deletes the position** |

On `PositionClosed`, when the settlement reserve is missing, the arm records `reserve-absent` and carries on: `buildClosedPosition` pushes a row carrying the full `settlement.proceeds`, and `positions.delete` retires the position. **Realized P&L rises by the full proceeds while no reserve is ever credited.**

That is the exact failure #366 was opened to fix, arrived at through a different miss. #366 fixed it for `provenance-absent` and deliberately did not touch `reserve-absent`.

## Why this is worth filing even though the defect is unreachable

The defect itself is hard to reach: the cross-ref existence gate rejects an unknown reserve before the fold runs, so on the gated path it cannot happen. Reachable only through callers that fold the durable log directly, bypassing both gates — `loadFoldedReview` at `packages/event-store/src/event-store.ts:225`. Migration-shaped, same caveat as the rest of #366.

**The stronger argument is not the defect. It is the inconsistency.**

Nothing in the type, the prose, or `SKIP_DETAIL` tells a reader which members discard whole and which half-apply. The distinction lives only in the shape of the four call sites. So the next person to add a reason code — and the vocabulary is explicitly designed to make that a deliberate, reviewed act — will pick one of the two behaviours by whichever call site they happen to read first, with nothing to tell them a choice is being made at all.

A closed vocabulary whose members disagree about what membership *means* has lost the property that made closing it worth doing.

The `SKIP_DETAIL` prose sharpens it: `provenance-absent` ends *"No balance moved."*, which #366 made true by making the discard total. `reserve-absent` ends *"No balance moved."* too — and on the close arm that is **false**, because the closed-book row and the retirement both moved. The same sentence is honest under one member and a lie under another.

## Options, unranked

1. **Make `reserve-absent` discard whole**, matching the other two. Smallest conceptual surface; the vocabulary regains one meaning. Consequence: a close naming a missing reserve leaves the position open instead of retiring it — the same trade #366 accepted, for the same reason (loud and inspectable beats silent and balanced-looking).
2. **Split the vocabulary explicitly** — encode "total" vs "partial" in the type so a reader cannot miss which is which, and so adding a member forces the choice.
3. **Leave it, and document it** — state in `contracts.ts` that members differ in application semantics, and say why. Cheapest, and still better than today, where the difference is undocumented.

Option 1 is the one #366's reasoning points at, but it changes behaviour on an arm that has shipped this way since before ADR-020, so it wants a ruling rather than a follow-on commit.

## Provenance of this issue

Found while making the `provenance-absent` discard total in #366, by the implementer, who flagged it rather than fixing it in scope. It is **pre-existing on `main`** and was not introduced by that PR. #366's body states plainly that it knowingly leaves the vocabulary split, and why.

Related: #367 (the durable-log read path — the same "unreachable through ingest, reachable through direct `foldEvents` callers" caveat applies to both).

AmetAlvirde:
Rules #371. Every member of `FoldSkipReason` now means the same thing about what the fold did, and this is what actually closes #371.

## The ruling

#371 offered three options. **Option 1** — make `reserve-absent` discard whole, matching the other two — is the one taken, on the reasoning #366 already established: loud and inspectable beats silent and balanced-looking.

Two things found while ruling changed the scope from what the issue described.

**The split was inside one member, not between members.** `reserve-absent` fired from seven sites and did two different things. On `Deposit` and `Withdraw` it meant "nothing applied" — there is nothing else in those arms. On the four LOT-DERIVED arms it meant "everything except the cash". And `Transfer` was a third case, half-applying by explicit design.

| arm | before | after |
|---|---|---|
| `Deposit`, `Withdraw` | nothing applied | unchanged |
| `PositionOpened` | position registered + cost anchor seeded, **no debit** | discarded whole |
| `PositionAddedTo` | lot appended, **no debit** | discarded whole |
| `PositionClosed` | closed-book row + `positions.delete`, **no credit** | discarded whole |
| `PositionTrimmed` | partial row + lots mutated, **no credit** | discarded whole |
| `Transfer` | source debited, destination absent | discarded whole |

**`PositionOpened` and `PositionAddedTo` were worse than the close arm #371 named.** A close at least inflates realized P&L against a real disposal. Those two grow the fund by a holding that no reserve paid for — NAV invention rather than misattribution.

## Transfer

The arm applied each leg independently so cash could leave and never arrive, leaving reserves quietly unbalanced. Its comment defended this as the state the channel exists to report. Reaching a broken state in order to report it is worse than not reaching it, so both reserves are now checked before either moves.

It records **once per event, not once per leg**: `dedupeFoldSkips` keys on `(eventId, reason)`, so two records for one transfer would both survive dedupe and print two operator lines for a single drop.

## What holds now

`FoldSkipReason` carries the invariant explicitly: a member names **what was missing**, never how much of the event survived the miss. There is no partial application anywhere in the vocabulary, so a new reason code inherits total discard rather than choosing it unawares — which was #371's actual complaint, over and above the defect.

That also makes `SKIP_DETAIL`'s fixed prose true everywhere. Both cash-leg reasons now say the EVENT was dropped rather than the leg, and `reserve-absent` says "a reserve" because a `Transfer` names two.

A consequence, pinned as a test: **no single event can now produce two skip records**, so `skipped.length` and `discardedEventCount` agree by construction within one fold. `dedupeFoldSkips` is not thereby vestigial — ingest re-folds the whole log once per accepted event (ADR-015), and collapsing the same standing drop across those n folds is its real job.

## Reachability

Unchanged, and migration-shaped — the same caveat #366 and #367 carry. The cross-ref existence gate rejects an unknown reserve before the fold runs, so on the gated path none of this occurs. It is reachable through callers that fold the durable log directly, such as `loadFoldedReview`.

## Tests

Nine net new tests. Four pin one arm each and were confirmed **red against the pre-change fold** before the fix was restored; three cover the `Transfer` legs; one pins that the gate stops only what it names, applying all four lot-derived arms in full against a present reserve.

The old behaviour was **not** pinned anywhere directly, but it was depended on implicitly: `fold.test.ts`'s helpers defaulted their cash legs to a reserve the genesis did not hold, so the leg no-opped and position-focused tests could ignore it. 15 tests failed on the first probe for that reason alone. `emptyGenesis` gains a funded desk reserve and the defaults name it; the ADR-002 tier test owns its own reserve, funded to drain to zero, because it asserts fund-wide percentages any standing balance would distort.

The one test that pinned the old `Transfer` shape is rewritten rather than deleted, and says what it reverses. The event-store digest test lost its fixture — one `Transfer` missing both legs was the only thing that made `skipped.length` differ from `discardedEventCount` — so it moves to two distinct dropped events, with the now-stronger invariant pinned beside it.

`pnpm typecheck` clean across all packages; `pnpm vitest run` 2442 passed, 52 skipped.

## Docs

`docs/coverage-rationale.md` is re-measured at `9e888d7`, from the uncovered-line report rather than the branch-location spans the first pass quoted. The three open `fold.ts` ranges survive as claims with every line number moved, and one claim is retired on evidence: the `Transfer` arm's `reserve-absent` skip is no longer uncovered, since the pre-flight check replaced two per-leg branches that no test reached.

## The ADR

`context/adr/ADR-021-one-meaning-for-the-discard-vocabulary.md` — *One meaning for the discard vocabulary: a reserve miss discards the whole event* — landed in `016c9dd`, with its row in `context/adr/INDEX.md`.

It was ADR-shaped for the reason flagged here from the start: this reverses a documented, reasoned in-code decision (`applyTieredLeg`'s *"A RESERVE MISS IS NOT A DISCARD and returns `true`"*) on an explicitly closed vocabulary. The ADR reconciles that with ADR-020's *"prefer a new reason over widening an existing one's meaning"* rather than claiming an exception — that rule guards a member's meaning being **widened** to cover new situations, and this **narrowed** one that was already carrying two senses, so read literally it would have protected the split rather than the readers. ADR-020 governs the token set; ADR-021 governs the semantics behind every token. Nothing in ADR-020 is superseded or amended.

It also records, plainly rather than buried, that the defect is migration-shaped — reachable only through callers that fold the durable log directly — so the ruling was taken on the inconsistency and not on the defect.


AmetAlvirde (commented):
## Review — `pr-review` (`review` discipline, Opus 5 high)

Baseline reproduced in a clean worktree of `feature/discard-vocabulary-one-meaning`: `pnpm typecheck` clean, `pnpm vitest run` **2442 passed / 52 skipped**, matching the PR body.

The ruling itself is implemented correctly and the fold-side reasoning holds up under attack. **One test lost its teeth in the fixture repair**, and the coverage re-measurement is numerically off in five places. Two comments elsewhere in `fold.ts` state contracts this diff invalidated. Details inline; the unanchored items are below.

### Hypotheses checked and refuted — no defect

- **`PositionAddedTo` hoist is safe.** `reserveDeltasForOpen([event.lot], …)` reads only `event.lot`, and `applyToReserve` → `applyReserveDelta` touch only `reserve.amount` / `reserve.lots`, which are disjoint from `positions`, `costAnchors`, `latestMark` and `entryWacFallback`. The fallback refresh still runs after the append and still sees the blended lots. Observationally equivalent on the applying path, as claimed.
- **The `Transfer` pre-flight predicate matches `applyToReserve`'s.** `applyToReserve` gates on `reserves.get(id)` truthiness; the only `reserves.set` site (`fold.ts:765`, `ReserveOpened`) stores a spread object literal, and the genesis map is built from `structuredClone`. No entry can be `undefined`, so `has` and `get` cannot disagree and the second leg cannot apply after a first-leg failure.
- **Nothing downstream depended on per-leg records.** Every consumer — `durable-log.ts:101`, `event-store.ts:316,368`, `ingest-walk.ts:200,221,240`, and the TUI formatters — routes through `dedupeFoldSkips` and counts *events*. `dedupeFoldSkips` does still have a job on the ingest path, which re-folds the whole log once per accepted event (ADR-015).
- **`SKIP_DETAIL` prose is clean against the operator-channel rule.** It is a record of fixed literals with no interpolation, and `recordSkip` assigns `detail: SKIP_DETAIL[reason]` verbatim — no event content, no fund figure. The "No state moved" claim was checked at every site that can now raise `reserve-absent`, including the trim arm: `splitTierRemoval` is genuinely pure (it copies lots into new arrays), so breaking out leaves `trimming.lots` untouched.
- **`emptyGenesis`'s new funded reserve distorts nothing.** No expectation in `fold.test.ts` was loosened — the diff is repointing plus additions. `percentOfFund` is asserted in exactly one place (the ADR-002 test), and reserves contribute 0 to `totalUnrealizedPnlUsd`, so the other total-asserting tests are insensitive to the standing balance. The ADR-002 test's own-reserve claim is true and the test is still sensitive: made to inherit the shared reserve, it fails loudly (`percentOfFund` 50 → 100.075).
- **`provenance-absent` was the right occurrence to leave on `"no-such-reserve"`** (`fold.test.ts:1116`), and it still proves precedence — inserting a `reserves.has()` check ahead of the provenance check in `applyTieredLeg` reddens exactly that one test, suite-wide.
- **The `Transfer` rewrite genuinely inverts.** `genesis()` is a fresh factory returning `amount: 10000`, so `toEqual(genesis().reserves)` fails at 9500 if the source is debited. The two added mirror tests are net new coverage.

### Findings with no single anchor line

**4. `docs/coverage-rationale.md:227-228` — the `crossref.ts` bullet still cites `fold.ts`'s `_never` latch at its pre-branch location `:737-740,742-743`.** That range moved on this branch; `const _never: never = event` is now at `fold.ts:781`. Corrected: `:781-784,786-787`. As it stands the doc gives two different line numbers for the same latch — the stale one at :228 and the re-measured one at :247.

**5. `docs/coverage-rationale.md:275-276` — `splitTierRemoval` cited at `fold.ts:987` with its `if (tierTotal === 0)` early return at `:996-998`.** Both were correct on `main`, and this branch's own `fold.ts` edits invalidated them; the re-measurement pass missed this paragraph. `fold.ts:987` is now `realizedPnlUsd: tierProceeds - tierCost,` inside an unrelated tier-attribution loop. Actual: `splitTierRemoval` is declared at `fold.ts:1031`, its early return at `:1040-1042`. The paragraph's substantive claim (it measures covered) is still true.

**6. `packages/engine/src/events/gate-carries-discards.test.ts:304-308` — prose drift the branch missed.** The title and comment describe "a Transfer's two absent legs collapse… one event dropped twice for one reason", a shape this branch deleted. The test still has teeth (it reddens under an identity `dedupeFoldSkips`), but for a different reason than it states — it is `walkPendingInbox` re-folding that produces the repeat, not two per-leg records. Same claim the diff correctly updated in `event-store.test.ts` and the `Transfer` arm, left stale here.

**7. Low — `docs/coverage-rationale.md:209` percentages.** `events/fold.ts` now measures **98.04%** lines (doc says 98.02%) and the aggregate is **93.88%** (doc says 93.85%). Both are explicitly tagged to the older vintage `8393bf6` rather than misattributed, so this is labelling drift, not a false claim — noted only because the bullet now carries two vintages while its `fold.ts` paragraph claims `a253dcd`.

### Measurement note

Coverage was re-run at HEAD (`npx vitest run --coverage`, v8 provider, 159 files passed / 4 skipped) with uncovered lines computed from `coverage-final.json` via `istanbul-lib-coverage`'s `getUncoveredLines()` — the same source the text reporter's "Uncovered Line #s" column uses. **Actual `fold.ts` uncovered lines at HEAD: `719-720`, `781-784,786-787`, `1089-1090`.**

The `a253dcd` vintage is legitimate and not stale by construction: HEAD (`8178019`) is docs-only, so `a253dcd` is the last commit touching `fold.ts` and a measurement at either point is identical.

Two claims in the doc were verified true: **the retired `Transfer` claim is correct** — the pre-flight `recordSkip` at `fold.ts:743` is genuinely covered by three tests in `fold-discard-channel.test.ts` (`:354`, `:386`, `:408`), all of which also assert reserves are byte-identical to genesis — and **`Withdraw` really is now the only `reserve-absent` record in the file no test reaches** (`fold.ts:347`, `:714`, `:743` all measure covered).

packages/event-store/src/event-store.test.ts
```diff
@@ -217,22 +217,50 @@ describe("loadEventLog — log-line quarantine", () => {
...
     });
     const folded = await loadFoldedReview(paths);
-    expect(folded.skipped).toHaveLength(2); // two records…
+    expect(folded.skipped).toHaveLength(2);
 
     const digest = deriveHeadDigest(folded, "transfer-both-ghost", "0.7.2");
     const verdict = unattendedFoldVerdict(folded);
 
-    expect(digest.discardedEventCount).toBe(1); // …one event.
+    expect(digest.discardedEventCount).toBe(2);
```
AmetAlvirde: **High — this test went vacuous in the fixture repair.**

The whole job of this test is to separate `deriveHeadDigest`'s *deduped* count from a raw `skipped.length`. On `main` the fixture was one `Transfer` with both legs ghosted: 2 skip records against 1 event, so the two numbers provably differed and only a deduped implementation passed.

The rewrite uses two **distinct** dropped events (`closeGhost` + `transferBothLegsGhost`). That is 2 records for 2 events, so `dedupeFoldSkips` is the identity here and `skipped.length === discardedEventCount === 2` by construction. The assertion can no longer distinguish the two definitions.

**Verified escaping mutation** — in `packages/engine/src/durable-log.ts:101`, change:

```ts
discardedEventCount: dedupeFoldSkips(skipped).length,
```
to
```ts
discardedEventCount: skipped.length,
```

The **entire** suite still passes: 159 files, 2442 tests, zero failures. On `main` that same mutation reddened this exact test (`toBe(1)` against `skipped.length === 2`).

The regression is not cosmetic. `deriveHeadDigest` is exported engine API, and its docstring at `durable-log.ts:57-67` promises the deduped key specifically so it can never disagree with `unattendedFoldVerdict` — which *is* called on concatenated multi-fold skips (the backfill's per-anchor lift, `event-store.ts:294-316`). A caller handing `deriveHeadDigest` a lifted envelope would over-count, and nothing would now catch it.

The replacement test below (`:242-261`) does not restore the teeth — `expect(...discardedEventCount).toBe(folded.skipped.length)` *encodes* the tautology the mutation relies on.

`dedupeFoldSkips` itself is still well covered (making it the identity reddens 4 tests in `gate-carries-discards.test.ts`, `fold-discard-channel.test.ts`, and the backfill report). It is only its use **inside `deriveHeadDigest`** that is now unguarded. Restoring the old both-legs-ghosted `Transfer` no longer works for this, so a fixture where the two counts genuinely differ now needs a source of duplicate records — e.g. a re-folded/concatenated envelope, which is also the shape the docstring's promise is really about.

docs/coverage-rationale.md
```diff
@@ -237,21 +237,28 @@ tests moved with the code into `packages/event-store/src/event-store.test.ts`.
   already-closed rejection branches this bullet used to cite as open — they
   measure covered in the fresh run.)
 
-  `events/fold.ts` (re-measured at `8393bf6`, after #329) has **three** ranges
-  open, and every line number this paragraph used to carry has moved:
+  `events/fold.ts` (re-measured at `a253dcd`, after #371) has **three** ranges
+  open — the same three as the previous vintage, all renumbered by that change:
 
-  - `:691-692` — the `Withdraw` arm's `recordSkip(event, order,
+  - `:718-720` — the `Withdraw` arm's `recordSkip(event, order,
```
AmetAlvirde: **The range is off by one at the start.** `packages/engine/src/events/fold.ts:718` is the `if (!applyToReserve(...))` test itself, which is executed (the `Withdraw` arm is exercised) and measures **covered**. Only the `recordSkip` body is uncovered.

Corrected: **`:719-720`** — which is also exactly the old `:691-692` shifted by +28, matching this bullet's own "only the lines moved" framing.

The likely cause across all three ranges in this bullet: these match v8's *branch* location spans (which start at the `if`/`switch` keyword — 718-720, 767-787, 1088-1090) rather than the uncovered-line report the rest of the document uses.

docs/coverage-rationale.md
```diff
@@ -237,21 +237,28 @@ tests moved with the code into `packages/event-store/src/event-store.test.ts`.
...
     "reserve-absent")`, taken when `applyToReserve` finds no such reserve. A
-    Discard-Channel skip record rather than a numeric guard, and a branch this
-    paragraph never named before.
-  - `:737-740,742-743` — the `foldEvents` exhaustiveness latch (`const _never:
+    Discard-Channel skip record rather than a numeric guard. Cited as
+    `:691-692` before; only the lines moved.
+  - `:781-787` — the `foldEvents` exhaustiveness latch (`const _never:
```
AmetAlvirde: **Wrong shape, not just wrong numbers — this range is not contiguous.** `fold.ts:785` (`` `handle — add its arm to the switch above.`, ``) measures **covered**; 781-784 and 786-787 are uncovered.

Corrected: **`:781-784,786-787`** — the same two-range shape as the previous vintage's `:737-740,742-743`, shifted by +44.

The bullet claims the range was only renumbered, but it was also silently merged across the hole, which loses the fact that the `throw`'s message-continuation line is reached.

docs/coverage-rationale.md
```diff
@@ -237,21 +237,28 @@ tests moved with the code into `packages/event-store/src/event-store.test.ts`.
...
     never = event; throw new Error(...)`), categorized like the other `_never`
     latch rows (§7's `skip-message.ts`/`orders/ingest.ts`/`orders/select.ts`):
     unreachable at runtime by construction, kept live so a new verb fails the
-    build rather than falling through silently. Cited as `:492-498` before;
-    only the lines moved.
-  - `:1045-1046` — the zero-`totalQuantity` early `return 0` in
-    `weightedAverageCost`. Cited as `:688-689` before; only the lines moved.
+    build rather than falling through silently. Cited as `:737-740,742-743`
+    before; only the lines moved.
+  - `:1088-1090` — the zero-`totalQuantity` early `return 0` in
```
AmetAlvirde: **Off by one at the start.** `fold.ts:1088` is `if (totalQuantity === 0) {`, which measures **covered**. Only the `return 0` body is uncovered.

Corrected: **`:1089-1090`** — the old `:1045-1046` shifted by +43, consistent with the bullet's "only the lines moved".

packages/engine/src/events/fold.ts
```diff
@@ -691,18 +719,34 @@ export function foldEvents(
...
+        //
+        // The existence check is deliberately separate from `applyToReserve`'s own
+        // boolean: that boolean reports whether a leg APPLIED, and by the time it could
+        // answer for the second leg the first has already moved.
+        if (!reserves.has(event.fromReserveId) || !reserves.has(event.toReserveId)) {
           recordSkip(event, order, "reserve-absent");
+          break;
         }
+        applyToReserve(reserves, event.fromReserveId, [{ tier: event.tier, amount: -event.amount }]);
+        applyToReserve(reserves, event.toReserveId, [{ tier: event.tier, amount: event.amount }]);
```
AmetAlvirde: **These two calls make `applyToReserve`'s stated contract false, and the docstring was not updated.**

`fold.ts:1070-1072` still reads:

> Every one of the eight call sites — open funding, close settlement, trim settlement, add funding, `Deposit`, `Withdraw`, and BOTH `Transfer` legs — must record on a `false`, one record per miss.

After this change neither of those two sites records on a `false` — both discard the boolean entirely, because the arm decides via the pre-flight `reserves.has` above. And "one record per miss" is precisely the rule this PR replaced with one record per *event*. So the contract is now wrong at 2 of its 8 named sites and wrong in its counting rule.

**Failure scenario:** a later change adds a ninth cash-leg call site, reads this docstring as the governing rule, and records per miss. That reintroduces exactly the double-record inflation of `discardedEventCount` and the duplicated `formatFoldDiscards` line that the comment at `:732-737` warns against — the docstring is the natural place a reader would look for that rule, and it currently instructs the opposite.

The contract that is actually true now is narrower: *the six sites that route through `applyTieredLeg` or call it directly must record on a `false`; the `Transfer` legs are pre-checked and deliberately ignore the return.*

packages/engine/src/events/fold.ts
```diff
@@ -691,18 +719,34 @@ export function foldEvents(
...
+        // That is the silent-and-unbalanced state the Discard Channel exists to end, not
+        // a state worth reaching in order to report it. So both reserves are checked
+        // BEFORE either leg applies, and a miss applies neither.
+        //
+        // ONE RECORD PER EVENT, not per leg. `dedupeFoldSkips` keys on
+        // (`eventId`, `reason`), so two `reserve-absent` records for a single Transfer
+        // would BOTH survive dedupe — inflating `discardedEventCount` and emitting two
+        // `formatFoldDiscards` lines for one drop. The channel would over-report the
+        // very thing it exists to report exactly once. This is the same precedence
+        // reasoning `applyTieredLeg` makes for `provenance-absent` over `reserve-absent`.
```
AmetAlvirde: **This reasoning is sound, but it leaves the matching example in `dedupeFoldSkips` stale.** `fold.ts:76-77` still reads:

> A `Transfer` whose BOTH legs name absent reserves is one event dropped twice for ONE reason and collapses to one finding, correctly.

That scenario is exactly what this arm now makes unreachable — a `Transfer` missing both reserves records once, so `dedupeFoldSkips` never sees the pair. The comment justifies the pair-key with an example the fold can no longer produce.

**Failure scenario:** the docstring is the stated rationale for keying on `(eventId, reason)` rather than `eventId` alone. A reader checking whether the pair-key still earns its keep finds a within-one-fold justification, tests it, finds it unreachable, and concludes the key can be simplified — losing the case the first half of the same comment actually describes (one event dropped for two *different* reasons). The live justification is the one the PR body gives and the new test at `event-store.test.ts:249-251` states — ingest re-folds the whole log once per accepted event (ADR-015), so the same standing drop repeats across folds. That belongs here, where the key is defined.

packages/event-store/src/event-store.test.ts
AmetAlvirde: Fixed in `9e888d7` — the fixture is a multi-fold union again rather than two distinct dropped events. It folds the log, then re-folds it grown by one event (the shape `walkPendingInbox` and the backfill actually produce), so `close-ghost` is reported once per fold: **3 records, 2 dropped events**. The test asserts the repeat explicitly (`union.skipped` length 3, `close-ghost` twice) *before* asserting `discardedEventCount === 2`, so it cannot go vacuous again.

Verified by escaping mutation, both directions: swapping `dedupeFoldSkips(skipped).length` for `skipped.length` at `packages/engine/src/durable-log.ts:106` fails it with `expected 3 to be 2`; reverting makes it pass.

The by-construction property you identified was kept, as its own test, labelled as guarding the arms rather than the dedup and explicitly noted as not standing in for it.

docs/coverage-rationale.md
AmetAlvirde: Accepted and fixed in `c0e00a4`. Your diagnosis of the cause was the more valuable half and it was right: the bullet was quoting the v8 *branch*-location spans while the rest of the file is written in the uncovered-line report. Which measurement is being quoted is now stated in the file, because the two units disagree and nothing in the numbers says which one you are reading.

The number differs from your `:719-720` only because the paragraph was re-measured at a later vintage (`9e888d7`). It now reads **`:732-733`**, with the covered `if` at `:731` and the reason the range starts inside it written beside it.

docs/coverage-rationale.md
AmetAlvirde: Accepted and fixed in `c0e00a4` — including the shape, which was the real finding. Re-measured at `9e888d7`, it now reads **`:794-797,799-800`**: two ranges, the hole explicit, with the reason the message-continuation line of the `throw` is reached stated beside it rather than silently merged across.

docs/coverage-rationale.md
AmetAlvirde: Accepted and fixed in `c0e00a4`. Re-measured at `9e888d7` → **`:1119-1120`**; the covered `if (totalQuantity === 0)` is no longer claimed as open.

packages/engine/src/events/fold.ts
AmetAlvirde: Fixed in `5ed1747`. The docstring no longer promises one record per miss across eight call sites. It now reads *"A `false` DISCARDS THE WHOLE EVENT, RECORDED ONCE (#371)"*, states that six of the eight sites read the boolean to discard the whole event while both `Transfer` legs deliberately discard it because the arm settles existence for the pair above them, and names the rule a ninth call site inherits — one record per EVENT, not one per call — with the cost of getting it wrong spelled out: two `reserve-absent` records for one event both survive the (`eventId`, `reason`) dedup.

packages/engine/src/events/fold.ts
AmetAlvirde: Fixed in `5ed1747`. The unreachable `Transfer` example is gone, and its absence is explained rather than quietly dropped: the comment now records that re-folding is the only way a repeat arises, and that the per-LEG report is where the original justification for this key came from.

The pair key is re-justified on its own terms instead — an id-only key would dedupe the current logs identically and begin silently swallowing findings the moment one event carried two reasons; that it never does is a decision each arm makes in the open (the precedence ruling in `applyTieredLeg`, the both-legs-or-neither pre-flight in the `Transfer` arm), not a property this key may assume on their behalf.

PR #383, opened 2026-08-18
The operator notice's twin property: bound the data half, take the job half off 5b

### LINKED ISSUE #376: The operator notice's job half is composed at the one moment it is guaranteed stale
Found while reviewing #375 (the lost-day delivery seam). Recorded rather than patched: every candidate remedy is a design decision on a converged grill, not a fix-cycle change. **This does not block #375** — see "Why shipping it recorded is safe" below.

## The defect, in one line

**Step 5b's heartbeat half is structurally staler than step 0's, and adds nothing but a window in which it can be wrong.**

## Mechanism

`job-heartbeat.json` has exactly one writer — `trap write_heartbeat EXIT` at `ops/price-feed/run-daily-fetch.sh:441`. Step 5b runs at `:902`. So when `pnpm operator-notice` composes the notice, the heartbeat on disk still describes the **previous** run, and `formatHeartbeatWarning` (`packages/event-store/src/heartbeat.ts:251`) emits its FAILED line on `exitCode !== 0` with **no staleness gate at all**.

Two consequences, both real:

- **Cry-wolf.** Run N−1 fails; run N — the recovery run — succeeds. At 5b it reads N−1's heartbeat and writes `the daily price job FAILED on <yesterday>`. The EXIT trap then records success. That false FAILED line stands in the operator's terminal for ~24h, *after the run that fixed things*. This fires on the recovery path specifically, on a channel whose entire design premise is that it has one account of operator attention to spend.
- **False all-clear.** Run N−1 succeeded; run N reaches 5b with clean data (notice truncated to empty = healthy), then dies at `backfill`. An all-clear stands over a failed run until the next run's step 0 — ~19h on the 23:00 fire, so the morning terminal shows health.

Name it plainly: **this is a channel that can say "healthy" while a run is failing, inside the increment built because every surface was pull-only and a lost day reached no one.**

## Why shipping it recorded is safe

- **The lost-day half — the actual point of #357 — is unaffected.** It derives live from the durable log; heartbeat staleness does not touch it. The 2026-08-14 / 08-15 demonstration in #375 stands.
- **The false all-clear is about job status, not lost days.** A run that dies at `backfill` has already written its marks (5b runs after `gap-report`), so the notice's silence about lost days is *correct*. What is delayed is the announcement that the run failed.
- **Step 0 already delivers the job half, fresher.** It reads `exitCode`/`lastStep` at run start, when the trap has fired. 5b's heartbeat lines are a second copy of the same fact, composed at the one moment it is guaranteed stale.

## Candidate remedies (none chosen here, deliberately)

1. **Drop the heartbeat lines from 5b entirely**, making step 0 the sole voice on job status. Cleanest separation, and it follows directly from "step 0's copy is strictly fresher." Cost: a run that fails after 5b says nothing about the job until the next fire.
2. **Write an in-progress heartbeat before 5b** (`exitCode 0`, `lastStep "operator-notice"`), letting the EXIT trap overwrite it with the truth. Kills the cry-wolf mirror outright, and is safe against a hard death mid-run because step 0's cleanliness test is `exitCode == 0 AND lastStep == complete`, so an in-progress breadcrumb correctly reads as *not* clean. Cost: reaches into #191's mark-window logic inside `write_heartbeat`.
3. **Record and accept**, which is what #375 does today.

Option 1 is the one the framing above points at.

## Related coverage note

Six harness cases drive the real wrapper through step 5b (case 1, case 7, both halves of case 6, both halves of case 8) and assert `pnpm operator-notice` was invoked, case 6 pinning its exact position between `gap-report` and `backfill`. **No harness case asserts the CONTENT of `operator-notice.txt` as written at 5b**, because the harness fakes `pnpm` wholesale and the real CLI never runs inside it. Any fix here should consider whether that content path deserves end-to-end coverage.

### LINKED ISSUE #381: The operator notice is never empty: venueDark has no recency bound, and from 2027-08-08 it counts days pnpm gap-report will not show
Found on 2026-08-18 while wiring #375's shell snippet into a real `~/.zshrc`, immediately after recovering 2026-08-14 and 2026-08-15 (17 marks, `accumulus@8bae007`). With the lost days gone, the notice did not go empty — it printed the venue-dark line, and it will print that line on every new shell forever. Recorded, not patched: the remedy collides with a stated design constraint (below), so it is a ruling, not a fix-cycle change.

Sibling of #376, which found the same channel's **job** half stale. This is its **data** half. Neither subsumes the other.

## The defect, in one line

**`venueDark` has no recency bound anywhere on this channel, so the notice is never empty on a store with any dark day in its history — and "empty means healthy" is the contract the whole delivery seam rests on.**

## Mechanism

`formatVenueDarkCount` (`packages/event-store/src/operator-notice.ts:217-226`) emits its line on `report.venueDark.length === 0` being false. No window, no recency filter — presence, at any age.

The window it reads is unbounded at the floor. `operator-notice-cli.ts` passes no window, so the derivation defaults to `LAUNCHD_ERA_START` — the fixed day `2026-07-03` (`gap-report.ts:145`, `:285`). The set of venue-dark days is therefore **monotonic**: about ten US market holidays a year enter it and nothing ever leaves.

Measured on the real store today: 1 venue-dark day (`2026-07-03`, twelvedata, 9 owed), zero lost days, 46/46 anchors checked — and a 125-byte notice on every new shell.

**The composer's own docstring already frames the split correctly and stops one step short.** It says the lost side spends its ceiling via `MAX_NOTICE_LOST_DAYS` (`:135`) and that a venue-dark **enumeration** on this channel would be "cry-wolf channel death arriving on a schedule, inside the fix." That reasoning bounded the *volume* — one line instead of ten — and left the *presence* unbounded. D5 of [the delivery grill] argued the count "never needs a threshold argued," which is true of the count and not of the window it counts over.

## Two consequences, the second one dated

**1. The empty file is unreachable on this store, from 2026-07-03 onward.** `operator-notice.ts`'s "EMPTY MEANS HEALTHY / self-clears with NO DISMISSAL STATE" contract is intact in the code and dead in practice: the file is non-empty on a perfectly healthy day, and the operator's eye learns that within a week. The lost-day half — the point of #357 — then arrives *appended to a line the reader has been trained to skip*.

**2. From 2027-08-08, the notice counts days its own instruction cannot show.** The line ends "Enumerate them with `pnpm gap-report`." That command's default floor is **not** the era start — it is `boundedEraFloor(now, MAX_WINDOW_DAYS)` with `MAX_WINDOW_DAYS = 400` (`apps/web/src/push/gap-report-core.ts:70`, `:101`), which starts rolling on 2027-08-08 and whose docstring says so plainly: *"the default report is a trailing window, not a complete history, from 2027-08-08 on."* After that date the notice's count includes days the enumeration it names will not print, and the divergence grows by one day per day. This is the same date-bomb shape `gap-report.ts:251-260` was written to warn about, reproduced across two callers instead of inside one.

## Why it is not a one-line fix

`operator-notice-cli.ts:13-17` states the constraint deliberately: *"The window is the derivation's own default — the launchd era floor and a ceiling of yesterday — which is exactly what the TUI's liveness banner uses, and this notice is that banner's push twin: the two must not describe different windows."*

So bounding the notice alone breaks the stated twin property, and bounding both changes what the TUI banner reports. That is a decision on a converged grill, not a patch.

## Candidate remedies (none chosen here, deliberately)

1. **Bound the venue-dark half on both twins** — a shared recency window (14 or 30 days) applied to `venueDark` in the banner and the notice together, preserving the twin property and restoring the empty file. Also closes consequence 2, since a bounded floor cannot outrun `gap-report`'s. Cost: an old, real venue outage stops being visible on either pull surface without `--since`. This is D5's own recorded flip-trigger, arriving from the direction D5 did not expect — not through a missed outage, but through the channel never falling silent.
2. **Bound only the notice, and retire the twin property explicitly** — the pull surface keeps the full history (it is scanned deliberately), the push surface carries only what is recent enough to act on. Honest, but it must be written down where the twin claim currently is, or the next reader will call it a bug.
3. **Have the notice inherit `gap-report`'s default floor** rather than the era start — kills consequence 2 by construction, since the count and the enumeration would share one window. Does nothing for consequence 1 until 2027-08-08, when 400 days start rolling.
4. **Accept it**, and delete the empty-means-healthy claim from `operator-notice.ts` so the docstring stops asserting a property the deployed store cannot exhibit.

## Coverage note

`operator-notice.test.ts` exercises `formatVenueDarkCount` at zero and non-zero, which is the whole of its current contract — so nothing here is a test gap under the code as designed. Any remedy that introduces a window needs a case pinning that a venue-dark day **older than the window** produces an empty notice, which is the assertion that would have caught this shape at authoring time.

AmetAlvirde:
Closes #381. Closes #376.

Two defects filed against the operator notice from opposite sides — a channel that can never be empty, and a channel that can be empty while a run is failing. They were grilled together on 2026-08-18 (`codebase-grill`, nine nodes, nine locked decisions, zero parked) because both turned out to be consequences of one unstated contract. The ruling is on both issue bodies as a comment; the remedy menus in the original bodies are superseded.

## The rule both halves fell out of

**A line earns this notice only if there is a move the operator can make today that deletes it.** That is what the existing ENUMERATE-`lost` / COUNT-`venueDark` split was already reaching for — it bounded the **volume** of what gets printed and left the **admission** open. The volume was never the problem.

Traced against a healthy store on 2026-08-19 — 2026-07-03 dark on record, previous run exited 1 at `post-check`, current run clean — the notice printed two lines and offered zero moves. Under the admission rule it is **empty**.

## Four commits

**1 — `refactor(event-store)`: the window width moves beside the era floor.** `MAX_WINDOW_DAYS` and `defaultGapReportSince` come down into `@numisma/event-store`; `gap-report-core.ts` re-exports them, so no importer changes. Behaviour-identical, and it is the seam the other three stand on: the package and `apps/tui` cannot import from `apps/web`, and the alternative was a second copy of 400 — the exact failure `venue-calendar.ts:12-20` records. Same shape as the #266 move that brought `venue-calendar.ts` into the engine.

**2 — the data half (#381).** `MAX_NOTICE_VENUE_DARK_DAYS = 7`, presentation-only at the leaf exactly like `MAX_NOTICE_LOST_DAYS`; `computeGapReport` still walks the whole window, because narrowing the derivation would narrow the **lost** half too and lost days are permanent. A venue-dark day is admitted only when recent, and that is the admission rule applied rather than an exception to it: a real outage's day is recoverable by `pnpm prices:fetch --as-of=<date>` and then leaves `venueDark`, because the report re-derives the finding from marks every run; a US market holiday's never will. Recency is the only proxy the derivation has, and it has the property that matters — the line leaves on its own even when the operator can do nothing.

The line also regains the holiday clause #266 D7 requires of every surface rendering this expectation. The store's one venue-dark day is 2026-07-03, observed Independence Day — the known false positive, pinned into a push channel with the sentence that would explain it stripped out. And it names its own window, because a count whose scope is invisible is a count the reader assumes is total.

Both twins now default their floor to `defaultGapReportSince`. The notice tells the operator to run `pnpm gap-report`; on the fixed era floor those diverge on 2027-08-08, and the cheap repair is refused by the code — a `--since=<era floor>` names a window wider than `MAX_WINDOW_DAYS`, which that command rejects outright. So it was not "the default won't show it" but "no invocation can". The default lives at both composers rather than at the two shells, so the twin binding is structural.

**The twin property is not retired.** It binds the **window**, not the **rendering** — the banner enumerates, the notice counts, and `operator-notice.ts` already required them to render this finding differently. That sentence is now in the CLI's header, because its absence is what made #381 read as unfixable.

**3 — the job half (#376).** All three heartbeat triggers come off step 5b; `formatOperatorNotice` loses its `heartbeatLines` argument. A correction to how the issue states the mechanism: the trap is EXIT-only, so at 5b the file holds run N−1's breadcrumb **exactly as step 0 saw it** — the two read the same bytes. 5b is not staler, it is identically sourced and differently scoped, and on the recovery path the file is right and the sentence is wrong. Two of the three triggers were anti-correlated with the truth of the run composing them: 5b's job half was written by a run that knows its own status and reads someone else's. `formatHeartbeatWarning`, `loadHeartbeatLines` and the TUI are untouched — the banner keeps all three triggers, and that their suites pass unedited is the evidence.

**4 — the bash.** The EXIT trap writes the notice too, on a non-zero exit only, closing the ~19-hour all-clear that commit 3 would otherwise leave standing over a run that dies after 5b. On a zero exit it does not touch the file. Step 0 and the trap share one writer function and one wording; only the scope line's tail varies by argument. Step 0 is kept as the only cover for a trap-less death.

## Assurance

- **Unit, the case #381 named:** a venue-dark day older than the window yields an **empty** notice, with every number driven off `MAX_NOTICE_VENUE_DARK_DAYS` rather than restated — plus the boundary from both sides, a mixed window, and a guard that the **lost** half is untouched by the bound.
- **Unit, the shared floor:** the notice's floor and `defaultGapReportSince` are the same value, asserted at a clock past 2027-08-08 where the two old floors genuinely differ. Commit 1 is what makes that expressible.
- **Unit, the silence:** a FAILED breadcrumb on disk beside a clean window produces an **empty file**. That is #376's false-FAILED-on-the-recovery-path case, and it goes red the instant anyone re-adds the argument.
- **Harness, real bash — case 9e:** the fake `pnpm operator-notice` writes an authored sentinel, the run dies at `backfill`, and the trap's FAILED lines must have replaced it. Two writers, one file, ordering matters. Cases 9a–9c now run to completion, which is the only shape in which step 0's own bytes survive to be read — and that buys the zero-exit restraint in the same three cases.

`pnpm typecheck` clean. `pnpm test` armed: **160 passed | 3 skipped** files, **2476 passed | 19 skipped** tests. `NUMISMA_WRAPPER_TEST=always pnpm test:wrapper`: **61 passed**. `shellcheck` and `bash -n` clean on the wrapper.

No ADR — the twin property is not being retired, and the constant move has the #266 precedent, which took none either. The docstrings that already carry this reasoning are where the ruling landed.

## Known residue, not closed here

Step 0's unique cover — a trap-less death by SIGKILL, OOM or power loss — has no harness case; producing one needs an external kill of the wrapper shell plus new residue handling. It is named in step 0's rewritten comment block.

### DIFF
--- a/apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
+++ b/apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
@@ -1954,9 +1954,16 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
    *
    * THE RUN SUCCEEDS, so this is the no-writer case end to end (#376): step 0 has no
    * breadcrumb to speak from, the EXIT trap's branch is gated on a non-zero exit, and the
-   * unarmed fake at 5b writes nothing. On a fresh machine's FIRST healthy run the file must
-   * therefore not exist at all — the strongest form of "say nothing, write nothing" the
-   * channel has.
+   * unarmed fake at 5b writes nothing. IN THIS HARNESS, therefore, no writer touches the
+   * file and it must not exist at all — the strongest form of "say nothing, write nothing"
+   * the two BASH writers have.
+   *
+   * SAID PRECISELY, BECAUSE THE PRODUCTION SENTENCE IS THE OPPOSITE: with a real toolchain
+   * a fresh machine's first healthy run DOES create `operator-notice.txt`, empty, because
+   * `writeOperatorNoticeFile` always writes including the empty case — that unconditional
+   * write is the self-clearing contract. What this case isolates is the bash half: with 5b
+   * inert, `undefined` vs `""` separates "no bash writer spoke" from "step 0 truncated
+   * unconditionally", and only the first is correct.
    */
   it(
     `case 9c — no heartbeat means no notice, ${RUNS_PER_CASE} consecutive times`,
--- a/apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
+++ b/apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
@@ -138,8 +138,12 @@ import {
 import {
   AUTHORED_CLEAN_OUTCOME,
   AUTHORED_FAILED_OUTCOME,
+  AUTHORED_SENTINEL_FRAGMENT,
   AUTHORED_SENTINEL_NOTICE,
   AUTHORED_WEDGED_OUTCOME,
+  EXIT_TRAP_SCOPE_TAIL,
+  STEP_0_SCOPE_TAIL,
+  armFakeOperatorNoticeWrite,
   expectedFailureFragment,
   readOperatorNotice,
   seedHeartbeatOutcome,
@@ -1772,14 +1776,28 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
   // the suite header already gives about `launchctl` — this harness delivers the signal,
   // not the sender.
   //
-  // ALL FOUR RUN OUT OF THE MARK WINDOW, so no run stamps itself and the heartbeat
-  // assertions stay about step 0 rather than about the carry-forward contract cases 6 and
-  // 8 own. And all four end the run EARLY — before step 5b — on purpose: a run that
-  // reached 5b would have the fake `pnpm operator-notice` succeed without writing anything,
-  // and the file's final contents would then be step 0's by default rather than by proof.
-  // Ending early is also the only shape production actually produces here, which is the
-  // trap this board has paid for before: a guard that pins a shape the real producer never
-  // emits is a guard that cannot fail.
+  // ALL FIVE RUN OUT OF THE MARK WINDOW, so no run stamps itself and the heartbeat
+  // assertions stay about the notice rather than about the carry-forward contract cases 6
+  // and 8 own.
+  //
+  // ── THERE ARE NOW TWO BASH WRITERS, AND THAT SETS THESE CASES' SHAPE (#376) ──────────
+  // The EXIT trap writes this same file, through the same shared function, on any non-zero
+  // exit — because a notice that carries only DATA findings is EMPTY on a run whose data
+  // was clean and which then died at `backfill`, and an empty notice reads as health for as
+  // long as it takes the next fire to notice. The trap is therefore the LAST writer on
+  // every failing run, which is exactly what 9e asserts.
+  //
+  // The consequence for 9a-9c is that step 0's own write is no longer observable at the end
+  // of a run that FAILED: the trap has overwritten it by then, correctly. So those three
+  // now run to COMPLETION and read the file afterwards. That is not a weaker vehicle, it is
+  // two claims for the price of one — the notice standing at the end of a clean run can
+  // only be what step 0 left there, AND the trap's zero-exit restraint (it must not
+  // truncate, `:>`, or otherwise touch the file when the run succeeded) is what makes that
+  // sentence true. A trap that cleared the file unconditionally fails all three at once.
+  //
+  // WHICH WRITER WROTE IS ASSERTED THROUGH THE SCOPE TAIL, never through the FAILED line:
+  // the two callers share that line verbatim on purpose, and `STEP_0_SCOPE_TAIL` /
+  // `EXIT_TRAP_SCOPE_TAIL` are the only place they differ.
 
   /**
    * CASE 9a · THE PREVIOUS RUN FAILED, SO THE NOTICE SAYS SO — naming BOTH fields.
@@ -1788,6 +1806,13 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
    * PATH problem on this machine and exit 124 at `timeout:backfill` is a wedged network
    * call, and a notice that could not tell them apart would send the operator to the wrong
    * place on the one channel that reaches them unasked.
+   *
+   * THE RUN ITSELF IS HEALTHY, and since #376 that is the only shape in which this claim
+   * can be read at all: on a failing run the EXIT trap overwrites step 0's lines with this
+   * run's own failure, which is the whole point of the trap. A clean run leaves the trap
+   * silent, so the file still holds what step 0 put there — a notice about the PREVIOUS run
+   * standing over a successful one, which is exactly what production does until step 5b
+   * replaces it.
    */
   it(
     `case 9a — a FAILED previous heartbeat is written into the notice, ${RUNS_PER_CASE} consecutive times`,
@@ -1796,10 +1821,6 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
         const options = markCaseOptions("out", CASE_OPTIONS);
         const dirs = makeCaseDir(options);
         seedHeartbeatOutcome(dirs.dataDir, AUTHORED_FAILED_OUTCOME);
-        // THIS run dies at `prices-fetch`, which is a DIFFERENT step from the seeded
-        // `resolve-tools`. Without that difference "step 0 reported the previous run" and
-        // "something reported this one" would be the same observation.
-        setFakeBehavior(dirs.caseDir, "prices:fetch", "exits-127");
         const label = `case 9a run ${run}/${RUNS_PER_CASE}`;
 
         const record = await launchWrapper({
@@ -1812,9 +1833,9 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
         });
 
         assertTriple(record, dirs.caseDir, {
-          exitCode: 127,
-          lastStep: "prices-fetch",
-          pnpmReached: ["prices:fetch"],
+          exitCode: 0,
+          lastStep: "complete",
+          pnpmReached: WRAPPER_PNPM_COMMANDS,
           mark: options.mark,
           label,
         });
@@ -1832,13 +1853,18 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
         expect(notice ?? "", `${label}: the notice's lines do not begin \`Numisma: \``).toMatch(
           /^Numisma: /,
         );
-        // AND IT IS THE PREVIOUS RUN'S OUTCOME, NOT THIS ONE'S. This run died at
-        // `prices-fetch`; a notice naming that step would mean step 0 had somehow read a
-        // heartbeat written after it, which is the one thing its placement rules out.
+        // AND IT WAS STEP 0 THAT WROTE IT. The FAILED line is shared verbatim with the EXIT
+        // trap, so only the scope tail can say which caller produced the file — and a trap
+        // that had written here on a ZERO exit would have replaced a real notice with one
+        // about a run that succeeded.
         expect(
           notice ?? "",
-          `${label}: the notice named THIS run's step — step 0 cannot know it`,
-        ).not.toContain("prices-fetch");
+          `${label}: the notice standing after a CLEAN run is not step 0's — something else wrote it`,
+        ).toContain(STEP_0_SCOPE_TAIL);
+        expect(
+          notice ?? "",
+          `${label}: the EXIT trap wrote a notice on a run that exited 0`,
+        ).not.toContain(EXIT_TRAP_SCOPE_TAIL);
       }
     },
     600_000,
@@ -1858,10 +1884,16 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
    * exists to end. Self-clearing is step 5b's job alone, because step 5b runs only once the
    * day's work is actually done.
    *
-   * So: an authored sentinel notice, a CLEAN heartbeat, and a run that dies before 5b. The
+   * So: an authored sentinel notice, a CLEAN heartbeat, and a run that finishes. The
    * sentinel must survive BYTE FOR BYTE — `toBe`, never `toContain`, because a step 0 that
    * appended to it rather than truncating it would also be wrong and `toContain` would
    * shrug.
+   *
+   * AND IT IS NOW THE SAME CASE FOR THE EXIT TRAP (#376). Two writers of one file are held
+   * to one restraint here: the run exits 0, so the trap's notice branch must not fire, and
+   * a trap that wrote unconditionally would delete this sentinel just as surely as a step 0
+   * that truncated. `pnpm operator-notice` is deliberately left UNARMED, so the fake writes
+   * nothing at 5b and the bytes on disk at the end can only be the seed.
    */
   it(
     `case 9b — a CLEAN previous heartbeat leaves the notice untouched, ${RUNS_PER_CASE} consecutive times`,
@@ -1871,7 +1903,6 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
         const dirs = makeCaseDir(options);
         seedOperatorNotice(dirs.dataDir, AUTHORED_SENTINEL_NOTICE);
         seedHeartbeatOutcome(dirs.dataDir, AUTHORED_CLEAN_OUTCOME);
-        setFakeBehavior(dirs.caseDir, "prices:fetch", "exits-127");
         const label = `case 9b run ${run}/${RUNS_PER_CASE}`;
 
         const record = await launchWrapper({
@@ -1884,23 +1915,24 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
         });
 
         assertTriple(record, dirs.caseDir, {
-          exitCode: 127,
-          lastStep: "prices-fetch",
-          pnpmReached: ["prices:fetch"],
+          exitCode: 0,
+          lastStep: "complete",
+          pnpmReached: WRAPPER_PNPM_COMMANDS,
           mark: options.mark,
           label,
         });
 
-        // THE RUN REALLY DID DIE BEFORE 5b, stated rather than assumed: if it had reached
-        // the notice step, the fake would have "succeeded" without writing and the survival
-        // below would prove nothing about step 0's restraint.
+        // THE RUN REALLY DID REACH 5b, stated rather than assumed: the unarmed fake writing
+        // nothing there is what makes the survival below attributable to the two bash
+        // writers rather than to a run that stopped before anything could write.
         expect(
           existsSync(join(dirs.caseDir, "sentinels", sentinelNameFor("operator-notice"))),
-          `${label}: the run reached step 5b, so this case cannot say what step 0 did`,
-        ).toBe(false);
+          `${label}: the run never reached step 5b, so this case cannot say what it proves`,
+        ).toBe(true);
         expect(
           readOperatorNotice(dirs.dataDir),
-          `${label}: step 0 rewrote or truncated a notice a HEALTHY previous run had left — ` +
+          `${label}: a notice a HEALTHY previous run had left was rewritten or truncated — ` +
+            "by step 0 at the start, or by the EXIT trap on a run that exited 0; either way " +
             "an empty file over a real lost day reads as health",
         ).toBe(AUTHORED_SENTINEL_NOTICE);
       }
@@ -1919,6 +1951,12 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
    * The assertion is `toBeUndefined`, and the distinction from `""` is the case: an empty
    * notice is the healthy SELF-CLEARED state, and a step 0 that truncated unconditionally
    * would produce exactly that and sail past a reader which flattened the two.
+   *
+   * THE RUN SUCCEEDS, so this is the no-writer case end to end (#376): step 0 has no
+   * breadcrumb to speak from, the EXIT trap's branch is gated on a non-zero exit, and the
+   * unarmed fake at 5b writes nothing. On a fresh machine's FIRST healthy run the file must
+   * therefore not exist at all — the strongest form of "say nothing, write nothing" the
+   * channel has.
    */
   it(
     `case 9c — no heartbeat means no notice, ${RUNS_PER_CASE} consecutive times`,
@@ -1927,7 +1965,6 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
         const options = markCaseOptions("out", CASE_OPTIONS);
         const dirs = makeCaseDir(options);
         // Deliberately no seed of any kind — `makeCaseDir` leaves a fresh data dir.
-        setFakeBehavior(dirs.caseDir, "prices:fetch", "exits-127");
         const label = `case 9c run ${run}/${RUNS_PER_CASE}`;
 
         const record = await launchWrapper({
@@ -1940,17 +1977,18 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
         });
 
         assertTriple(record, dirs.caseDir, {
-          exitCode: 127,
-          lastStep: "prices-fetch",
-          pnpmReached: ["prices:fetch"],
+          exitCode: 0,
+          lastStep: "complete",
+          pnpmReached: WRAPPER_PNPM_COMMANDS,
           mark: options.mark,
           label,
         });
 
         expect(
           readOperatorNotice(dirs.dataDir),
-          `${label}: step 0 invented a failure report on a machine with no breadcrumb — ` +
-            "cry-wolf on day one, from the one channel whose only asset is being believed",
+          `${label}: a failure report was invented on a machine with no breadcrumb and a ` +
+            "run that succeeded — cry-wolf on day one, from the one channel whose only " +
+            "asset is being believed",
         ).toBeUndefined();
       }
     },
@@ -1971,8 +2009,16 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
    * the fake bin: the isolation contract still passes (every entry is absolute and inside
    * the case dir), the real `pnpm` is in neither the prepend nor the inherited launchd
    * PATH, and the wrapper's own `command -v pnpm` guard is what ends the run. That the
-   * notice is on disk afterwards is the proof that step 0 ran BEFORE that guard and
+   * notice is on disk afterwards is the proof that the channel ran BEFORE that guard and
    * survived it.
+   *
+   * WHICH OF THE TWO BASH WRITERS IS ON TRIAL HERE CHANGED WITH #376, and the case is
+   * sharper for it. Both write on this path — step 0 about the SEEDED wedged run, then the
+   * EXIT trap about THIS run's exit 127 — and the trap is last. So the file at the end must
+   * name `resolve-tools`, not the seed, and that single assertion carries two facts: the
+   * notice channel still speaks with no toolchain at all, and the trap is the writer that
+   * stands. Step 0's own bytes are unobservable from here for the same reason; 9a is where
+   * they are read.
    */
   it(
     `case 9d — the notice survives an unresolvable toolchain, ${RUNS_PER_CASE} consecutive times`,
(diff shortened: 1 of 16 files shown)


docs/price-feed-ops.md
```diff
@@ -886,22 +904,40 @@ A clean machine prints nothing on a new shell. That is the whole contract, and i
...
+breadcrumb and the "has not completed since" staleness line. Both live on in the
+TUI banner, which reads the same heartbeat live; neither is actionable on the one
+channel that arrives unasked, because the run printing it has usually just fixed
+the thing it names.
 
-**Job lines** (from `job-heartbeat.json`) name a run that failed and the step it
-died at, a breadcrumb dated ahead of today (a wrong machine clock), or a job that
-has not completed since some date. These are the wrapper-level failures; triage
-them with the sections above.
+**A data report** is what step 5b writes, and it is everything below.
```
AmetAlvirde: **MEDIUM — the venue-dark section of this same doc (lines 956–971, just below this line) still documents the pre-#381 line, and this PR did not touch it.**

This hunk rewrote "What it means when it does print" down to `**A data report** is what step 5b writes, and it is everything below.` — and then *everything below* is the old behaviour. Line 959 still prints the string commit `e1db115` replaced:

```text
Numisma: 3 venue-day(s) dark — not lost days: the feed ran and the days are anchored. Enumerate them with pnpm gap-report.
```

The shipped line (`operator-notice.ts:387-392`) is now:

```text
Numisma: 3 venue-day(s) dark in the last 7 days — not lost days: the feed ran and the days are anchored, and the venue was silent or the market was closed for a holiday. Enumerate them with pnpm gap-report.
```

And the prose at 963–971 still describes the count as unbounded and total — "A venue-dark day is **permanent and it accumulates**… So the notice carries the number" — with no mention of `MAX_NOTICE_VENUE_DARK_DAYS` anywhere in the file. It even keeps the old rationale ("they would grow without bound") as the *only* reason for the count/enumerate split, which is exactly the argument D1 says bounded the volume and left the admission open.

**Concrete failure.** 2026-11-26 (Thanksgiving) goes dark; on 2026-12-05 the notice is empty again. The operator, working from this doc, reads the count as the store's running total, so an empty notice means "every venue-dark day I ever had has been cleared" — the reverse of what the bound means. Or the other way: on 2026-11-27 the notice says `1 venue-day(s) dark in the last 7 days` while `pnpm gap-report` — the command the line names — enumerates fifteen. The doc gives the reader nothing to reconcile that with.

This is the operator-facing home of the behaviour #381 changed, and the doc is otherwise the most carefully maintained artefact in the change. Fix: update the sample line to the emitted text, and add the 7-day bound (and the flip trigger of 3) to the paragraph beneath it.

packages/event-store/src/operator-notice.ts
```diff
@@ -44,14 +67,65 @@
...
+ * Two of the three are ANTI-CORRELATED with the truth of the run composing them.
+ *
+ * THE ALTERNATIVE IS REFUSED BY THE BASH, NOT BY TASTE. "Write an in-progress
+ * heartbeat before 5b" would make the file describe THIS run — and `write_heartbeat`'s
+ * first act after capturing `$?` is a SIGKILL of the watchdog plus a `pgrep -P` reap
+ * of its `sleep`, so an early write would leave 5b AND `backfill`, the longest step,
+ * running with no timeout. It buys back only the future-dated trigger.
+ *
+ * SO THE CHANNELS SPLIT BY LANGUAGE: this notice is purely the DATA channel, and the
+ * job channel is purely bash — the wrapper's step 0, which knows its own run's scope
```
AmetAlvirde: **MEDIUM — this header is the designated home of the #376 ruling, and commit `88ea2d1` made two of its claims false without updating it.**

This paragraph was written in commit `30d51d7`, when step 0 really was the only bash writer. Commit `88ea2d1` then gave the `EXIT` trap `write_operator_failure_notice` — and touched `operator-notice-io.ts` (+8 lines, correctly) but not this file at all. So the module header now says:

1. Line 121-123: *"the job channel is purely bash — the wrapper's **step 0**, which knows its own run's scope because it is inside it."* The job channel is now two bash writers, and the trap is the one that carries the case this header spends fifty lines motivating. Step 0 is a backstop (`run-daily-fetch.sh:368-376` says so in those words).
2. Lines 101-104, the `exitCode !== 0` bullet: *"**Step 0** already carries this, correctly scoped, and step 0 reaches the two paths 5b never sees at all."* After `88ea2d1` the trap carries it, and it carries it about **this** run rather than the previous one — which is the whole reason the FAILED sentence stopped being anti-correlated with the truth of the run composing it.

Also, the clause "which knows its own run's scope because it is inside it" is the exact inverse of what step 0 does: it is inside run N and speaks about run N−1. That is fine and deliberate for a backstop — but it is the trap, not step 0, that the sentence actually describes.

**Concrete failure.** The next reader auditing the channel starts here — D9 ruled "the docstrings that already carry this reasoning are where the ruling lands." They read that the job half is step-0-only, work out the 23:00-fire / dies-at-`backfill` hole for themselves, conclude it is open, and either re-file it or re-implement the trap write that is already forty lines away in `write_heartbeat`. That is exactly the cost the four-commit split was structured to avoid.

(By the same token `run-daily-fetch.sh:1073` — commented separately — and the header's "EMPTY MEANS HEALTHY" section, which is still true, but is now true of a file three writers share.)

ops/price-feed/run-daily-fetch.sh
```diff
@@ -966,9 +1056,28 @@ pnpm gap-report -- --write
...
+#     the heartbeat lines here until then, and this is the one moment they are
+#     guaranteed to be misread: the EXIT trap is EXIT-ONLY, so at 5b the breadcrumb
+#     still holds the PREVIOUS run's bytes — exactly what step 0 read at the top of
+#     this same run. Not staler, identically sourced and DIFFERENTLY SCOPED: step 0
+#     says the PREVIOUS run failed, `formatHeartbeatWarning` says "the daily price
+#     job FAILED", present tense, about the job. On a recovery run — the previous
+#     run died at `prices-fetch`, this run is landing the missed days right now —
+#     the file is right and the sentence is wrong, and the run writing it is the run
+#     that just fixed it. So the channels split by LANGUAGE: this notice is the DATA
+#     half, and the job half is bash that knows its own run — step 0 today, correctly
```
AmetAlvirde: **MEDIUM — same staleness as `operator-notice.ts:122`, and this one is in the file commit `88ea2d1` rewrote.**

"the job half is bash that knows its own run — **step 0 today**, correctly scoped to the PREVIOUS run and reaching the two paths this step never sees at all."

Commit `30d51d7` wrote that; commit `88ea2d1` changed +130 lines of this same file — including step 0's own comment block at `:368-376`, which was correctly updated to say *"It is no longer the only bash channel for a partway death"* — and left this paragraph alone. The two blocks now contradict each other about the same mechanism, 700 lines apart, and this is the one a reader of step 5b hits first.

Specifically wrong here after `88ea2d1`:
- "step 0 **today**" — the trap is the primary writer of the job half now, and this block is the only place in the file that describes the split without mentioning it.
- "correctly scoped to the **PREVIOUS** run" — the trap's call is scoped to *this* run, which is the point of D7 and the reason the ~19-hour all-clear closed.

**Concrete failure.** Someone maintaining the wrapper reads step 5b top-to-bottom to answer "what reports a run that dies after 5b?", gets "step 0 today", and concludes the answer is "nothing until the next fire." That is the defect this commit closed, restated as documentation in the file that closed it.

ops/price-feed/run-daily-fetch.sh
```diff
@@ -396,8 +447,11 @@ if [[ -n "$PREVIOUS_EXIT_CODE" && -n "$PREVIOUS_LAST_STEP" ]]; then
     # things of different vintages in one voice, which is the credibility spend this whole
     # design refuses. So: one bounded, currently-true message, and `NOTICE_SCOPE_LINE` above
```
AmetAlvirde: **LOW — dangling identifier: `NOTICE_SCOPE_LINE` no longer exists anywhere in the repo.**

This surviving line says *"one bounded, currently-true message, and `NOTICE_SCOPE_LINE` **above** says out loud that lost days are not in it."* Commit `88ea2d1` deleted `NOTICE_SCOPE_LINE` (and `NOTICE_FAILED_LINE`) from this block and re-created them as `notice_scope_line` / `notice_failed_line` inside `write_operator_failure_notice` at `:219-221`.

```
$ grep -rn 'NOTICE_SCOPE_LINE' .
ops/price-feed/run-daily-fetch.sh:448:    # design refuses. So: one bounded, currently-true message, and `NOTICE_SCOPE_LINE` above
```

One hit, and it is this comment. Two things are wrong at once: the name is gone, and "above" now means ~230 lines up in a different function rather than the four lines this sentence used to point at. The reference is load-bearing prose — it is what discharges the trade the paragraph spends twelve lines accepting.

Suggested: `` …and the scope line in `write_operator_failure_notice` says out loud… ``

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
```diff
@@ -1919,6 +1951,12 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
    * The assertion is `toBeUndefined`, and the distinction from `""` is the case: an empty
    * notice is the healthy SELF-CLEARED state, and a step 0 that truncated unconditionally
    * would produce exactly that and sail past a reader which flattened the two.
+   *
+   * THE RUN SUCCEEDS, so this is the no-writer case end to end (#376): step 0 has no
+   * breadcrumb to speak from, the EXIT trap's branch is gated on a non-zero exit, and the
+   * unarmed fake at 5b writes nothing. On a fresh machine's FIRST healthy run the file must
```
AmetAlvirde: **LOW — this sentence generalises a fake-induced outcome into a production claim, and the production claim is false.**

"On a fresh machine's FIRST healthy run the file must therefore not exist at all — the strongest form of 'say nothing, write nothing' the channel has."

It does not hold with a real toolchain. `writeOperatorNoticeFile` (`operator-notice-io.ts:130-137`) opens with *"IT ALWAYS WRITES, INCLUDING THE EMPTY CASE, and that is the whole self-clearing contract"* and unconditionally `writeFile(path, "")`. So on a fresh machine's first healthy run, step 5b **creates** `operator-notice.txt`, empty. `toBeUndefined()` passes here only because `pnpm operator-notice` is left unarmed and the fake is inert — which the preceding sentence says, and then the conclusion ignores.

The assertion itself is fine and worth keeping (it is still the `undefined` vs `""` distinction, about the two bash writers). It is the last sentence that needs to say "in this harness", not "the channel".

**Worth flagging alongside it, and I am not asking for a change here:** the block comment this reshape deleted at 1772-1786 warned that *"a guard that pins a shape the real producer never emits is a guard that cannot fail"*, and gave that as the reason 9a-9d ended early. After the reshape all three of 9a/9b/9c end in a state production cannot produce — a completed run whose `operator-notice.txt` was **not** rewritten by 5b. The reasoning for the trade is written out at 1780-1794 and it is a legitimate one (an inert 5b is what isolates the bash writers, and it buys the zero-exit restraint three times over). But the sentence that used to guard against this specific shape went away in the same edit, and case 9c's docstring is where the consequence surfaced. Case 9e is the one that carries a real 5b write, and it does it well.

packages/event-store/src/operator-notice.ts
```diff
@@ -135,22 +210,67 @@ export function formatNoticeCheckFailure(error: unknown): string {
 export const MAX_NOTICE_LOST_DAYS = 10;
```
AmetAlvirde: **LOW — this constant's docstring (lines 177-178, just above) still says the CLI gets the *derivation's* default, which commit `e1db115` made false.**

"The CLI passes no window, so production gets **the derivation's default (~90 days)**"

After `e1db115` the CLI's floor is filled in by the *composer*: `loadOperatorNoticeLines` does `since: window.since ?? defaultGapReportSince(now)` (`operator-notice-io.ts:85`), so `computeGapReport`'s own `window.since ?? LAUNCHD_ERA_START` default (`gap-report.ts:338`) is never reached from this path again. Today the two resolve to the same string, so nothing observable changes — and that is exactly what makes the sentence a trap: it diverges on 2027-08-08, and this is one of the two places a reader would look to find out what window the ninety-line incident was measured against.

The evidence this is an oversight rather than a choice: the twin sentence in `operator-notice-io.test.ts:167-170` *was* updated in this same commit, from "the derivation's own floor (`LAUNCHD_ERA_START`)" to "the composer's own floor (`defaultGapReportSince` — the era start today, a rolling 400-day floor from 2027-08-08)". This one was missed.

(Separately, "~90 days" does not match either floor at any point in this repo's history — `LAUNCHD_ERA_START` is 2026-07-03 — but that predates this PR and I am not counting it as a finding here.)

packages/event-store/src/index.ts
AmetAlvirde: **LOW — `index.ts:60-61` (file-level comment, because the line is outside the diff hunk): the package's own export-block comment for the notice is false after commit `30d51d7`.**

```ts
// The operator notice (#357): the liveness banner's PUSH twin. The same two signals
// the TUI prints, composed into a file the shell profile cats on every new terminal
```

It is no longer "the same two signals the TUI prints". `formatOperatorNotice` lost its `heartbeatLines` argument in `30d51d7`; the notice composes **one** signal, the gap findings, and `operator-notice.ts:121-127` rules that the banner keeping all three heartbeat triggers while the notice keeps none is correct and permanent.

This PR edited this export block — `MAX_WINDOW_DAYS` and `defaultGapReportSince` were added 31 and 35 lines above — so the comment was in the diff's immediate neighbourhood.

**Concrete failure.** This is the package's public index; the comment is the one-paragraph description a consumer of `@numisma/event-store` reads before deciding what `loadOperatorNoticeLines` returns. "The same two signals the TUI prints" tells them the heartbeat warning is in there, which is precisely the re-add `operator-notice.test.ts`' arity assertion exists to make go red.

Suggested: "the DATA half of what the TUI prints — the gap findings only; the job half is the wrapper's bash (see `operator-notice.ts`, #376)."

packages/event-store/src/operator-notice.ts
```diff
@@ -209,17 +337,56 @@ function formatWithheldLostDays(withheld: number): string {
...
+ * ── THE WINDOW IS ANCHORED AT `report.until`, NOT AT `now` ────────────────────
+ * This half is PURE and reads no clock; `report.until` is already clamped to
+ * yesterday by the derivation, which is the same instant a clock read here would
+ * resolve to and is reproducible in a test besides. A report narrower than the bound
+ * simply filters nothing — no clamp is needed, because the floor only ever moves
+ * earlier than the report's own.
  */
 function formatVenueDarkCount(report: GapReport): string[] {
-  if (report.venueDark.length === 0) {
+  const floor = addDays(report.until, -(MAX_NOTICE_VENUE_DARK_DAYS - 1));
```
AmetAlvirde: **NIT (verified, not speculative — flagging it because the arithmetic was worth checking and it is correct).**

The window is right: `floor = until − 6`, `date >= floor`, so `[until−6, until]` is exactly seven calendar days inclusive, and `operator-notice.test.ts`' `OLDEST_ADMITTED` / `NEWEST_REFUSED` pair pins it from both sides off the constant. No off-by-one. The narrower-report case is also right — `floor` can only move earlier than `report.since`, so it filters nothing and cannot throw.

Two residual wrinkles, both presentation-only:

1. **"in the last 7 days" is measured from `until`, which is *yesterday*.** The docstring at 375-377 defends the anchor as "the same instant a clock read here would resolve to" — true, `dueThrough(now)` is yesterday — but the sentence the operator reads says "the last 7 days" and means `[today−7, today−1]`. So a venue-dark day exactly seven days before the morning they read it is counted, and today is not in the window at all (it never can be). Concretely: dark on Mon 2026-11-02, read on Mon 2026-11-09 → still counted, on the eighth calendar day back. Harmless, and arguably unfixable without either a clock read here or "in the 7 days ending <until>".

2. **A caller that hands in a report narrower than the bound still gets "in the last 7 days".** `operator-notice.test.ts` pins that as intended ("filters nothing when the report itself is narrower than the bound"), and the zero-argument CLI can never produce it — but the sentence names a window the report did not open. Only reachable from a test or a future caller passing `since`.

Neither is worth a change on its own; recording them so the next person doing this arithmetic does not redo it.

AmetAlvirde:
## Review summary — adversarial pass over all four commits

Eight findings, all posted as inline threads. **No high-severity findings, and no correctness defect in the bash or the TypeScript.** The three MEDIUMs are all documentation asserting behaviour the code no longer has — which this codebase treats as a defect, not a nit — and every one of them is a claim that a *later commit in this same PR* falsified.

### Findings, most severe first

| # | Severity | Where | One line |
| --- | --- | --- | --- |
| 1 | MEDIUM | `docs/price-feed-ops.md:956-971` | The venue-dark section still prints the pre-#381 line and describes the count as an unbounded running total; `MAX_NOTICE_VENUE_DARK_DAYS` appears nowhere in the doc |
| 2 | MEDIUM | `packages/event-store/src/operator-notice.ts:101-104, 121-123` | The header D9 designates as the ruling's home still says the bash job channel is step 0 only; `88ea2d1` made the EXIT trap the primary writer and never came back |
| 3 | MEDIUM | `ops/price-feed/run-daily-fetch.sh:1072-1075` | Same staleness ("step 0 today", "scoped to the PREVIOUS run") in the file `88ea2d1` rewrote, contradicting step 0's own updated block 700 lines up |
| 4 | LOW | `ops/price-feed/run-daily-fetch.sh:448` | Dangling `NOTICE_SCOPE_LINE` — one grep hit in the repo, and it is this comment |
| 5 | LOW | `run-daily-fetch.test.ts` case 9c docstring | "on a fresh machine's FIRST healthy run the file must not exist at all" is false in production — `writeOperatorNoticeFile` always writes, including empty |
| 6 | LOW | `operator-notice.ts:177-178` | "the CLI…gets the derivation's default" — the composer now supplies the floor; the twin sentence in `operator-notice-io.test.ts` was updated and this one was not |
| 7 | LOW | `packages/event-store/src/index.ts:60-61` | "the same two signals the TUI prints" — the notice composes one signal since `30d51d7` |
| 8 | NIT | `operator-notice.ts:381` | The 7-day arithmetic is correct; two presentation wrinkles recorded so nobody redoes the check |

### What was checked and found sound

**`run-daily-fetch.sh` — the EXIT trap.** No defect found, and I went at this hardest.
- **Zero-exit restraint holds.** `if [[ "$HEARTBEAT_STATUS" -ne 0 ]]` is the only gate; on a clean run nothing touches `operator-notice.txt` — no truncate, no `:>`. Cases 9a/9b/9c now assert it three ways.
- **It cannot replace the exit code.** `HEARTBEAT_STATUS=$?` is still the first act; the new branch is a plain `if` (so `errexit` never fires on a false condition), the call is `|| true` over a function that is `return 0` unconditionally with its one `printf` guarded `2>/dev/null || true`, and the trap still ends in `return 0` — which does not alter an EXIT trap's inherited status.
- **Guard ordering holds on every early path.** The notice branch sits *after* `[[ -d "$DATA_DIR" ]] || return 0`. `exit 78` (blank `NUMISMA_DATA_DIR`) is at `:117`, before the trap is installed at `:628`, so it is uncovered by design and step 0 still backstops it. Both `exit 127`s are after `:628` and reach the writer, which is what case 9d now proves.
- **No `set -u` exposure.** `write_operator_failure_notice` reads only `$OPERATOR_NOTICE_FILE` (`:176`) and its three arguments; `$HEARTBEAT_LAST_STEP` is assigned at `:574`, unconditionally, before the branch. The `MARK_WINDOW`-in-a-trap hazard the file documents is still discharged — `MARK_WINDOW=false` at `:281`, `MARKS_LANDED_STEPS` at `:295`, `CARRIED_MARK_WINDOW_AT=""` at `:312`, all before the trap install.
- **No variable collision.** The five unlocalised globals are all `notice_*`-prefixed and appear nowhere else in the file; step 0's call completes ~170 lines before the trap exists, so there is no re-entrancy.
- `bash -n` and `shellcheck` both clean on the wrapper.

**`formatVenueDarkCount`.** `addDays(until, -(7-1))` with `date >= floor` is seven calendar days inclusive — verified from both sides, and the suite drives it off the constant rather than restating 7. Narrower-than-bound filters nothing, as claimed. Details in finding 8.

**The inherited floor.** `window.since ?? defaultGapReportSince(now)` at both composers — an explicit `since` still wins, and nothing is silently overridden. The only two production callers (`app.ts:57`, `operator-notice-cli.ts:57`) pass no window. Today's behaviour is identical on both surfaces (`boundedEraFloor(now, 400)` is still `LAUNCHD_ERA_START` until 2027-08-08); the TUI does not change. The `MAX_WINDOW_DAYS` move keeps `gap-report-core.ts`'s one-import structural property intact, because the bare `export {}` carries no `from`.

**The harness.** 9a still asserts both seeded fields and gains a stronger writer-discriminator (the scope tail) than the step-name check it replaced; 9b is strictly stronger than before (it now binds the trap's restraint as well as step 0's); 9d's re-aim is sound and its `not.toContain` on the wedged seed cannot collide with the run-under-test's `127`/`resolve-tools`; 9e's geometry is right (5b precedes `backfill`) and the fake fails closed at exit 96 rather than degrading to `succeeds`. Only 9c's prose overreaches — finding 5.

### Verification run locally

- `NUMISMA_WRAPPER_TEST=always pnpm test:wrapper` → **61 passed**, 1 file, 502s. Cases 9a-9e all green, including the two mutation guards.
- `pnpm typecheck` → clean, exit 0.
- `bash -n` + `shellcheck` on `ops/price-feed/run-daily-fetch.sh` → clean.

No verdict posted — findings only.

docs/price-feed-ops.md
AmetAlvirde: Fixed in 370815b. The sample line is now the emitted text (holiday clause and named window included), and the paragraph beneath says the count is bounded to `MAX_NOTICE_VENUE_DARK_DAYS`, why the bound exists (a holiday no command can clear would otherwise print forever and "empty means healthy" would be unreachable), and that the count self-extinguishes by age. Added the two reconciliations you named: an empty notice means nothing fell in the last seven days rather than everything was cleared, and `pnpm gap-report` enumerating more is the design, since the bound is presentation-only at the leaf and the derivation window is never narrowed. Swept the rest of the doc for the same class — the "empty means healthy" section already credits the EXIT trap, and nothing else in the file asserts step 0 as the sole bash writer.

packages/event-store/src/operator-notice.ts
AmetAlvirde: Fixed in 370815b. The `exitCode !== 0` bullet now says the bash carries it with the trap as the primary carrier, scoped to this run at the moment it dies, which is why the FAILED sentence stopped being anti-correlated. The "SO THE CHANNELS SPLIT BY LANGUAGE" paragraph now describes two bash writers explicitly: the EXIT trap as primary (inside the failing run, its own scope) and step 0 as backstop (scoped to N-1, covering alone a death with no trap — SIGKILL, OOM, power loss — plus a run that dies before the trap is installed). Also added a line pointing at the two `run-daily-fetch.sh` blocks and saying the three must be kept in agreement.

ops/price-feed/run-daily-fetch.sh
AmetAlvirde: Fixed in 370815b, worded to match the `operator-notice.ts` header and step 0's own block: two bash writers, the EXIT trap primary and scoped to THIS run, step 0 the backstop scoped to the PREVIOUS run and covering alone the trap-less deaths. "step 0 today" is gone.

ops/price-feed/run-daily-fetch.sh
AmetAlvirde: Fixed in 370815b — took your suggestion: the sentence now points at the scope line in `write_operator_failure_notice` (`notice_scope_line`), and says it is defined with the writer near the top of the file rather than "above".

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
AmetAlvirde: Fixed in 370815b. The conclusion is now scoped to the harness, and a following paragraph states the production fact you named: with a real toolchain the first healthy run does create the file, empty, because `writeOperatorNoticeFile` always writes and that unconditional write is the self-clearing contract. The `undefined` vs `""` distinction is restated as what it actually isolates — the bash half with 5b inert. Left the assertion and the deleted-guard trade alone, as you asked.

packages/event-store/src/operator-notice.ts
AmetAlvirde: Fixed in 370815b, matching the twin sentence in `operator-notice-io.test.ts`: production gets the composer's floor via `defaultGapReportSince(now)` (era start today, a rolling 400-day floor from 2027-08-08), and `computeGapReport`'s own `LAUNCHD_ERA_START` default is never reached from this path. The "~90 days" claim went with it; the ninety-line smoke-run measurement is kept and now says which floor it was measured against.

packages/event-store/src/index.ts
AmetAlvirde: Fixed in 370815b — took your suggested wording: "the DATA half of what the TUI prints — the gap findings only; the job half is the wrapper's bash (#376, see `operator-notice.ts`)".

packages/event-store/src/operator-notice.ts
AmetAlvirde: Recorded in 370815b rather than changed, which is what you asked for. The anchor docstring now carries the checked arithmetic (`[until-6, until]`, seven inclusive, pinned both sides off the constant by `OLDEST_ADMITTED`/`NEWEST_REFUSED`) and both wrinkles as presentation-only and accepted: the window reads `[today-7, today-1]` because it is anchored at yesterday, with the cost of fixing the wording named (a clock read in the pure half, or "in the 7 days ending <until>"); and a narrower report still names the seven-day window, unreachable from the zero-argument CLI. Behaviour untouched.

Issue #370, opened 2026-08-18
readline's own error still reaches the operator on two paths #346 did not close — TTY Ctrl-D, and record-fill's eager interface


AmetAlvirde:
#346 closed the piped/non-TTY path in `orders:import`: the shell now builds its readline lazily, and on a stdin that is no terminal it never builds one at all — it names the missing terminal in its own voice, returns an empty answer, and lets the flow refuse as `no-reserve-declared`. Two sentences, two layers, no readline internal.

**Two paths of the same defect class survive, and they are one issue because they need the same thing to fix honestly: a harness that can stand at a pty.**

## 1. Ctrl-D at a real terminal still surfaces Node's own string

`import-orders-cli.ts` guards on `process.stdin.isTTY`. When a terminal *is* present the guard does not fire, so Ctrl-D at the funding question rejects inside `rl.question` and the outer catch prints Node's wording verbatim.

```
printf '\004' | script -q /dev/null npx tsx apps/tui/src/import-orders-cli.ts <export.csv>
```

Observed: `Aborted with Ctrl+D` on stderr, exit 1, **no** "No terminal on stdin" sentence and **no** `REFUSED — no funding reserve was declared for this batch`.

A blank line at a TTY refuses correctly, so this is a gap rather than a regression — but it is the same complaint #346 exists to answer: readline's error, not ours, reaching the operator.

Likely shape of the fix is `rl.question(q).catch(() => "")`, composing with the existing design — the empty answer is the one honest answer, and the domain refuses in its own voice. It should not be taken on faith; see the harness note below.

## 2. `record-fill-cli.ts` still has the pre-#346 shape

Untouched by #346 and uncharacterized by its own tests:

- `record-fill-cli.ts:34` — `createInterface` at module scope, eagerly consuming stdin
- `:95` — `ask: (question) => rl.question(question)`
- `:106` — unconditional `rl.close()`
- no `isTTY` guard anywhere

A piped run prints `readline was closed` verbatim, which is exactly the sentence #346 removed from the other shell.

This shell matters more than the one already fixed: it has **nine `ask` call sites** against `import-orders`' single reachable question. A no-terminal `orders:import` run puts exactly one question and then refuses, which is why its once-per-run notice guard is real but unobservable. Here the notice would actually be emitted repeatedly, so "say it once, not once per question" becomes observable behaviour rather than a defensive flag.

## Why one issue

[...]

PR #384, opened 2026-08-19
ADR-017 in post-landing tense with its >= boundary, and record-fill's readline internal (#379, #370 symptom 2)

### LINKED ISSUE #379: docs(ADR-017): status says the implementation "does NOT land" — it landed, and the ADR now describes the opposite of the code

AmetAlvirde:
`ADR-017`'s status line said the implementation **does NOT land with this ADR** and that
*"the code today does the opposite, deliberately and under test."* It landed — **PR #341,
merged 2026-08-14**, in four commits (`e282b25` add-to, `ff50da5` red test + as-of world
contract, `ec71936` trim, `fa41693` same-dated prefix). The pin the status cited as
evidence for the opposite had already been inverted in place. What actually closes #379 is
the rewrite below plus the three retargeted citations.

## What the seal rule does at the `>=` boundary — verified, and the drafting text had it wrong

Three sites, all in `crossref.ts`, all read against the code and against green tests:

| site | equality | evidence |
| --- | --- | --- |
| `requirePositionUntouchedAfter` | **ACCEPTS** — success on `asOf >= latest.asOf`, reject only on strict `<` | `it("case 1 — accepts a same-day trim-then-close, and folds BOTH legs")`, load-bearing assertion `costBasisUsd: 50` |
| `crossReferenceTrim` / `crossReferenceAddedTo` | **REFUSE** — reject on `event.asOf >= retiredAsOf` | `describe("ADR-017 — the trim boundary")`, `describe("ADR-017 — the boundary")` |
| `worldAsOf` prefix | **INCLUDES** — `priorEvents.filter(e => e.asOf <= asOf)` | `fa41693` |

So the two guards give **opposite answers on the close date**. The drafting-time text
called them *"matching"* and said they *"must not disagree there — otherwise a batch's
verdict would depend on which of the pair arrived first."* Both halves are wrong:

- They disagree on purpose. At equality the **log index** decides, not the date —
  `foldEvents` orders by `(asOf, then log index)`. Trim-then-close folds; close-then-trim
  cannot. Each site takes the reading that matches what the fold will actually do.
- The verdict *does* depend on arrival order, and that is correct. It is the fold's own
  tie-break showing through the gate; a gate that hid it would admit events the fold
  silently drops, which is the class ADR-015 exists to eliminate.

The same tie-break bounds `fold.ts`'s detect-by-absence pass, which reports an
`InvalidationMarked` dated strictly after retirement and leaves the same-date tie silent
(#297's named residual).

## The rewrite

Post-landing tense throughout, and every site named by **symbol** — the drafting-time line
numbers had all gone stale, which is what #379 catalogued. Retargeted: the four-site verb
table (`crossReferenceTrim` / `crossReferenceAddedTo` / `crossReferenceClose` /
`crossReferenceInvalidation`), the as-of world (shipped as `EventReference.worldAsOf`,
lazy, and a full reference rather than a fragment), the settlement-magnitude hole it closes
structurally, `checkDebit`, `positionLots`, and the inverted pin plus the replacement pin
that was owed and has landed. A new section states the boundary above.

Two paragraphs had stopped being true and are corrected rather than left standing:

- the `InvalidationMarked` detect-by-absence exception **shrank to a same-date tie** under
  #297 (`2bd9b7a`) — the ADR still described it as wholly undetected;
- **#293 closed as the Discard Channel without covering this ADR's case.** An *admitted*
  backdated trim is discarded by nothing, so it produces no `skipped` record. The
  silently-moving closed row this ADR names is still unannounced, and the ADR now says so
  instead of pointing at #293 as its companion.

`INDEX.md`'s row 21 is rewritten to match, status column included.

## The three blocked citations

`crossref.ts:817-820` contains no comparison — it is prose about ADR-015's third delta
class. #349 deliberately left these three rather than swapping the line number, because the
surrounding claim was unverified and a correct line number is exactly what a reviewer
checks. Now verified, they name **functions** instead: `ADR-017`'s boundary section, plus
the comments in `position-backdated-add-to.test.ts` and `position-backdated-trim.test.ts`.

Three **in-source** comments carried the identical *"must not disagree"* sentence — the
`closedPositionAsOf` field doc and both dated guards — and are corrected in the same pass.
Leaving them asserting the opposite of the ADR would rebuild the defect this issue exists
to close.

`pnpm typecheck` clean. No `crossref.ts:819` citation remains in the tree.

---

# Part 2 — #370, symptom 2: `record-fill`'s readline internal

Bundled here because part 1 is docs-only. **Refs #370, does not close it** — symptom 1
(Ctrl-D at a real terminal) is a different trigger and stays open.

## The harness decision, settled on measurement

#370 says both symptoms are *"only honestly fixable behind a pty"* and that a fix landed
without one *"would ship with an assertion that cannot fail."* Two findings change that:

**Symptom 2 needs no pty.** An eager module-scope interface against piped stdin reproduces
it exactly — `ERR_USE_AFTER_CLOSE` / `readline was closed`. A spawn's stdin *is* the pipe
that provokes it, so the existing `record-fill-cli.test.ts` harness (already running with
`input: ""`) has been printing the internal all along. The three new cases **fail against
the old shell with exactly that string** — verified by reverting the shell and re-running.

**`script -q /dev/null` fails as a harness.** With piped stdin it delivers Ctrl-D to the
pty before the child starts, so all three inputs collapse:

| input | observed |
|---|---|
| `printf '\004'` | `AbortError` / `Aborted with Ctrl+D` |
| `printf '\n'` (blank line) | `AbortError` / `Aborted with Ctrl+D` |
| `printf 'hi\n'` (real answer) | `AbortError` / `Aborted with Ctrl+D` |

It cannot express the blank-line-versus-Ctrl-D contrast the issue's own argument rests on,
and an assertion built on it would pass for the wrong reason — the #365 pattern. It also
needs a platform branch (`script -q /dev/null cmd args` on macOS vs
`script -qc "cmd args" /dev/null` on util-linux) while CI is ubuntu and dev is darwin.

**Ruling: extract the channel, no pty.** Recorded in `prompt-channel.test.ts`'s docblock so
the rejected option is on the record rather than re-litigated.

## The change

`prompt-channel.ts` holds the three decisions #346 gave the import shell and left inline —
lazy interface, no-terminal notice returning `""`, notice once per run. `isTTY` goes in as
a **value** and the interface as a **factory**, so the channel is testable with no terminal,
no process and no spawn. Both shells now use it; `import-orders-cli.ts`'s behaviour is
unchanged and its existing suite is the proof.

A piped `record-fill` run now says two sentences from two layers — the shell names the
missing terminal, the flow refuses as `unknown-rung` and states nothing was written.

## The ordering dependency is pinned, not just commented

`""` is honest only while the **first** question reached is one that refuses on it. Three of
`recordFill`'s nine `ask` sites read `""` as a ratification — the quantity question takes the
remaining quantity, the book questions accept the proposed verdict. `record-fill.test.ts`
now asserts a no-answer interview asks **exactly one** question and writes nothing, so a
reorder goes red instead of silently letting an unattended run write. The once-per-run
notice also becomes observable for the first time: the import shell puts one question before
refusing, `recordFill` has nine.

## Verification

`pnpm typecheck` clean, `pnpm test` green — **2466 passed, 51 skipped, 164 files** (+13).
The three end-to-end cases were confirmed red against the pre-fix shell before being kept.


AmetAlvirde (commented):
## Review — #384

Checked the claims rather than the prose, since that is what this PR asks for.

**Everything load-bearing in part 1 verifies against the source at `2268556`:**

- `requirePositionUntouchedAfter` (`crossref.ts`) returns `null` when `latest === undefined || asOf >= latest.asOf` — equality ACCEPTED, strict `<` the only rejection. ✅
- `crossReferenceTrim` and `crossReferenceAddedTo` both reject on `event.asOf >= retiredAsOf`. ✅
- `worldAsOf` folds `priorEvents.filter((event) => event.asOf <= asOf)`. ✅
- `foldEvents` sorts `(a.event.asOf === b.event.asOf ? a.order - b.order : …)` — (asOf, then log index). ✅

The unifying explanation holds too: at equal `asOf` the log index decides, each gate takes the reading that matches what the fold does, and the verdict genuinely turns on arrival order.

**Citation sweep clean.** `it("case 1 — accepts a same-day trim-then-close, and folds BOTH legs")` exists and its `costBasisUsd: 50` really is the non-vacuity assertion; `describe("case C is admitted, the backdated trim ADR-017 relaxed")`, `it("and the book it folds to is the one that argued for admitting it")`, `describe("ADR-017 — the trim boundary")`, `describe("ADR-017 — the boundary")`, `openLate`/`trimPayload`/`closePayload`/`caseC`, `FoldedReview` in `contracts.ts`, `position-absent` on all three fold arms, `held ? world.lastClose.get(…)` / `if (held && last !== undefined)`, `crossReferenceAddedTo` never touching `positionLots`, `requirePositionUntouchedAfter` called from `crossReferenceClose` only — all present and all saying what is claimed. Commits `e282b25` / `ff50da5` / `ec71936` / `fa41693` / `2bd9b7a`, PRs #301 (merged 08-11) and #341 (merged 08-14), and issues #293 / #294 / #297 all check out. No bare `.ts:NNN` citation survives in ADR-017, and `crossref.ts:819` is gone from the whole tree.

**Part 2 is green here**: 60 tests across `prompt-channel.test.ts`, `record-fill.test.ts`, `record-fill-cli.test.ts`, `import-orders-cli.test.ts`. The `import-orders-cli.ts` rewiring looks behaviour-preserving (same notice bytes, same `""`, same lazy build, same `rl?.close()`), and dropping `record-fill-cli.ts`'s unconditional `rl.close()` is a strict improvement — nothing is opened, so nothing needs closing.

---

**Eight findings, inline.** The first three are the class this PR exists to close: prose that confidently asserts something the code does not do. The fourth is an assertion that cannot fail, and this repo has already written down why that exact assertion cannot fail.

1. `record-fill-cli.ts:44` — the interview is seventeen questions, not nine; `io.ask` is handed on to `authorLadderTarget`, and one of the questions there ratifies on blank.
2. `record-fill-cli.ts:52` — three `[y/N]` confirmations, not two, and one of them does not abandon.
3. `record-fill-cli.ts:59` — cites `record-fill-cli.test.ts` for the ordering pin; the pin is in `record-fill.test.ts`.
4. `record-fill-cli.test.ts:257` — `expect(notices).toBe(1)` cannot fail, and "observable here for the first time" is wrong twice.
5. `prompt-channel.test.ts:152` — `not.toContain("readline")` is tautological after the `toEqual([NOTICE])` above it.
6. `ADR-017 …md:188` — "undecidable" contradicts `fold.ts`, which says the tie IS decidable and left undecided by choice.
7. `ADR-017 …md:128` — "is consulted at four sites" is now two; a post-landing-tense slip in the document whose subject is post-landing tense.
8. `INDEX.md:21` — `asOf <= asOf` renders as a tautology.

**One gap, not a defect introduced here.** The PR pins `recordFill`'s ordering dependency and leaves `import-orders-cli.ts`'s stated in prose only — `import-orders-cli.ts:62-68` says a reorder "must move in the same commit" with nothing that goes red if it doesn't. `import-orders-cli.test.ts` asserts the shell sentence and the `no-reserve-declared` refusal but never that the funding question is asked FIRST. The symmetric pin (a no-answer import interview puts exactly one question and writes nothing) would cost one test and close the asymmetry the docstrings now describe as closed.

apps/tui/src/record-fill-cli.ts
```diff
@@ -28,10 +28,46 @@ import {
...
+/**
+ * THE PROMPT CHANNEL (#370, symptom 2). This shell built its interface HERE, at module
+ * scope, which is the pre-#346 shape: `createInterface` eagerly consumes stdin, so a piped
+ * run ended the stream before the first question was put and `ask` rejected with
+ * `ERR_USE_AFTER_CLOSE` — `readline was closed`, a readline internal printed verbatim to
+ * the operator by the outer catch below. The channel builds lazily and, on a stdin that is
+ * no terminal, names the missing terminal in this shell's own voice and returns "".
+ *
+ * `""` IS HONEST HERE ONLY BECAUSE OF THE ORDER `recordFill` ASKS IN, and that is a
+ * dependency rather than a property of the empty string. This flow has NINE `ask` sites and
```
AmetAlvirde: **The interview is not nine questions — `recordFill` hands `io.ask` onward.**

`record-fill.ts` has nine literal `io.ask(` calls, but at `record-fill.ts:807` it passes `io.ask` straight into `authorLadderTarget`, which puts **eight more** (`record-fill-ladder-target.ts:63-67, 111, 121, 125`). The reachable interview is seventeen questions.

That matters past the arithmetic, because one of the eight is a ratification this census does not cover:

```ts
// record-fill-ladder-target.ts:111
if (isNegative(await ask(`Append this lot to '${ladder.positionId}'? [Y/n]: `))) {
```

`isNegative("")` is `false`, so a blank answer **ratifies the append** — a `[Y/n]` gate that says yes to silence, and the only gate standing between an unattended run and attaching a lot to an existing Position. `  Tempo [${reserve.tempo}]:` (`:125`) is a second one: blank takes the default. So "three of nine read `\"\"` as a ratification" undercounts both halves of the fraction.

Same nine-site count is asserted in `prompt-channel.ts:26`, `record-fill.test.ts:1124` and `record-fill-cli.test.ts:248`. Either count the delegated sites or say the census covers `record-fill.ts` only and name `authorLadderTarget` as the part that is not surveyed — the current sentence reads as a complete census and is not one.

apps/tui/src/record-fill-cli.ts
```diff
@@ -28,10 +28,46 @@ import {
...
+ * they read `""` three different ways:
+ *
+ *   - `record-fill.ts`'s "Which rung filled?" — the FIRST question the flow reaches —
+ *     matches no rung and refuses as `unknown-rung`. This is the refusal that makes the
+ *     empty answer safe, and nothing has been written when it fires.
+ *   - "Filled quantity [n]" reads `""` as TAKE THE REMAINING QUANTITY, and the per-rung
+ *     book questions read it as ACCEPT THE PROPOSED VERDICT. Silent ratifications, both.
+ *   - the two "[y/N]" confirmations read `""` as NO, and abandon.
```
AmetAlvirde: **Three `[y/N]` confirmations, not two — and the middle one does not abandon.**

`record-fill.ts:736` ("Confirm these derived verdicts?"), `:745` ("Also record N confirmed cancellation(s) in this act?") and `:900` ("Write BOTH?"). The middle one:

```ts
if (cancelled.length > 0) {
  if (isAffirmative(await io.ask(`Also record ${cancelled.length} confirmed cancellation(s) in this act? [y/N]: `))) {
    alsoCancelled = cancelled.map((verdict) => verdict.orderId);
  }
}
```

`""` → `isAffirmative` false → `alsoCancelled` stays `[]` and **the flow continues** to the write gate. That is a silent ratification of "record none of them", so it belongs in the ratification bullet above, not in this one. Plus `record-fill-ladder-target.ts:111`'s `[Y/n]`, which ratifies the opposite way on blank (see the other comment).

As written this bullet tells a future reader that every remaining confirmation is a safe refusal on `""`. Two of the four are not.

apps/tui/src/record-fill-cli.ts
```diff
@@ -28,10 +28,46 @@ import {
...
+ *
+ * Only the first is ever reached without a terminal. Reorder the interview — ask the book
+ * observation before the rung pick, or make the rung question skippable — and this same
+ * `""` ratifies a quantity nobody stated and the run WRITES, unattended, with nobody
+ * having answered anything. If that ordering moves, this must become a refusal the domain
+ * cannot mistake for consent (a sentinel the questions reject, not a blank), and it must
+ * move in the same commit. `record-fill-cli.test.ts` pins the ordering itself so the move
```
AmetAlvirde: **Wrong file.** The ordering pin is `record-fill.test.ts:1131` — `describe("an interview nobody can answer refuses at the FIRST question and writes nothing")`. `record-fill-cli.test.ts` contains no assertion about question order at all; its three new cases assert stderr content and a notice count, and its own comment at `:253` points the reader at `record-fill.test.ts` for the pin.

A reader who follows this citation to check the claim finds nothing that pins it, which is the same failure mode as the `crossref.ts:819` citations part 1 of this PR is retargeting.

apps/tui/src/record-fill-cli.test.ts
```diff
@@ -176,3 +176,84 @@ describe("record-fill-cli — the shell refuses a partial durable log (finding 2
...
+  it("says it ONCE, though the interview has nine questions behind that one", async () => {
+    const result = runFill(await cleanDataDir());
+
+    // The once-per-run guard, observable here for the first time: on the import shell
+    // exactly one question is ever put, so the flag was correct and unpinnable. It stays
+    // one here only because the rung pick refuses first — see the ordering pin in
+    // `record-fill.test.ts` — so this assertion and that one fail together if the
+    // interview is ever reordered.
+    const notices = result.stderr.split("No terminal on stdin").length - 1;
+    expect(notices).toBe(1);
```
AmetAlvirde: **This assertion cannot fail, and the repo already wrote down why.**

The notice is emitted only from inside `PromptChannel.ask`, and this run reaches exactly one `ask`: the case two tests up asserts `result.stderr` contains `no resting rung matches`, which is `recordFill`'s refusal at `record-fill.ts:587` — fired by the FIRST question. One `ask` ⇒ one notice, whether or not `toldThereIsNoTerminal` exists. Delete the flag, write the notice unconditionally on every `ask`, and this case stays green.

So the comment at `:251` — "The once-per-run guard, observable here for the first time" — is wrong twice:

1. `import-orders-cli.test.ts:452-460` records this exact experiment and its verdict, on the sibling shell: *"This used to carry a `toHaveLength(1)` count over the notice, which could not fail — a no-terminal run puts EXACTLY ONE question. … Delete the `toldThereIsNoTerminal` flag entirely, write the notice unconditionally, and all nine cases here stay green. … Counting a one-element set proved none of it."* This case re-adds the removed assertion against a shell with the identical one-question shape.
2. The guard **is** observable in this PR — `prompt-channel.test.ts:145-152` forces nine asks and counts one notice. That is the first (and real) observation, and it is a unit test, not this one.

The case isn't worthless as a smoke test, but the comment should say what `import-orders-cli.test.ts` says in the same situation: the count is over a one-element set and proves nothing about the guard.

apps/tui/src/prompt-channel.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    await expect(harness.channel.ask("a? ")).resolves.toBe("");
+  });
+
+  it("writes the shell's own sentence, not readline's", async () => {
+    const harness = channelOn(false);
+
+    await harness.channel.ask("a? ");
+
+    expect(harness.errors).toEqual([NOTICE]);
+    expect(harness.errors.join("")).not.toContain("readline");
```
AmetAlvirde: **Cannot fail.** `errors` is pushed to only by `io.err`, and `createPromptChannel` calls `io.err` with exactly one value — `io.noTerminalNotice`, which this test sets to its own `NOTICE` constant. The line above already asserts `toEqual([NOTICE])`, so the array's full contents are pinned; `not.toContain("readline")` is then a statement about a string literal defined ten lines up.

There is also no path by which readline could contribute here: on this branch the channel never constructs one, which the case above ("never touches readline at all") pins directly. The version of this assertion that can fail is the end-to-end one in `record-fill-cli.test.ts` (`not.toContain("readline was closed")`), which per the PR body was verified red against the old shell.

context/adr/ADR-017-a-close-is-the-positions-last-dated-event.md
```diff
@@ -46,95 +55,146 @@ Nothing may be dated after the close, because the fold applies events in
...
+**does** depend on which of the pair the log received first — that is not a flaw
+to be designed away, it is the fold's own tie-break showing through the gate, and
+a gate that hid it would be admitting events the fold silently drops, which is
+the whole class ADR-015 exists to eliminate.
+
+The same tie-break governs one more site, and it is the honest edge of the rule:
+`fold.ts`'s detect-by-absence pass reports an `InvalidationMarked` dated
+**strictly after** its position's retirement and stays silent on a mark dated
+**on** it, because the closed book does not carry the close's log index and the
+tie is undecidable from what that pass holds (#297's named residual, deliberately
```
AmetAlvirde: **"Undecidable" is the word `fold.ts` was written to prevent.**

The detect-by-absence comment (`packages/engine/src/events/fold.ts`, the RESIDUALS block) says the opposite, and explains that it is saying it on purpose:

> RESIDUALS, NAMED AND ACCEPTED … **the comment must not assert a boundary the code does not hold.**
>
> 2. THE SAME-DATE TIE: a mark dated ON its position's retirement date. The fold breaks an equal-`asOf` tie by log index, **so this IS decidable** — and, to be exact about why it is nevertheless undecided: the index is not missing state. `order` is in scope in the `PositionClosed` arm above, right where the row is pushed … What is missing is only the CLOSED BOOK's carriage of it — a consequence of where this derivation is SITED, **a choice, not a constraint that binds this function.**

The ADR's "deliberately not widened" is right; "the tie is undecidable from what that pass holds" is the framing the source explicitly disclaims. Suggest tracking `fold.ts`'s own sentence: decidable in principle, left undecided because the retirement date is re-derived from `closedPositions` after the loop and that row carries no log index — a siting choice, not an impossibility.

In a rewrite whose subject is an ADR that described behaviour the code did not have, this is the one paragraph that re-creates the defect in miniature.

context/adr/ADR-017-a-close-is-the-positions-last-dated-event.md
```diff
@@ -46,95 +55,146 @@ Nothing may be dated after the close, because the fold applies events in
...
+The frequency argument is what turned a wrong answer into a decision. **Fills are
 recorded by hand.** A trim discovered while reconciling an exchange export two
 days after the close is not an exceptional event in this fund's operation; it is
 the ordinary shape of hand-recorded history. A rule whose only remedy is
 out-of-band log surgery (option A) is the wrong rule for the common case.
 
 ## Why exactly two verbs
 
-`closedPositionIds` is consulted at four sites. The relaxation reaches two of
+`closedPositionIds` is consulted at four sites. The relaxation reached two of
```
AmetAlvirde: **Present tense, no longer true.** `closedPositionIds` is consulted at exactly **two** sites today — `crossReferenceClose` (`crossref.ts:1110`) and `crossReferenceInvalidation` (`crossref.ts:786`). The other two now read `closedPositionAsOf`, which is what the table three lines below correctly records in its ruling column.

The status block sets the standard this trips over: *"This document is written in post-landing tense throughout; where it says the gate does something, the gate does it today."* Reads correctly as "`closedPositionIds` **was** consulted at four sites" — which is also the premise the section needs, since the whole point is that two of those four moved off it.

context/adr/INDEX.md
```diff
@@ -18,7 +18,7 @@
...
...-date follow-up has the whole world to ask. `PositionAddedTo` needed only the comparison (`crossReferenceAddedTo` never reads `positionLots`), so the two halves were not symmetric in cost and the add-to landed first (`e282b25` ahead of `ec71936`). **Other consequences:** retroactivity is UNBOUNDED and a bounded variant was rejected by name (a clock inside a pure fold makes the same batch admissible Tuesday and refused Friday); a closed row can now move with nothing announcing it — `FundReviewData` still has no warnings field, and ADR-020's Discard Channel does not close this, since an ADMITTED backdated trim is discarded by nothing and produces no record, so #293 closed without covering it; the pin inverted, its first test rewritten to assert acceptance while the second became the positive expectation, and the owed replacement pin LANDED as `describe("ADR-017 — the trim boundary")` / `describe("ADR-017 — the boundary")` in the two `position-backdated-*.test.ts` files; the reserve-balance-at-date gap under `checkDebit` is pre-existing and merely widened, named so it is not mistaken for a new hole. `EVENT_SCHEMA_VERSION` stayed `2` — no verb added, no record reshaped, no migration. |
```
AmetAlvirde: Nit, but this row is the summary a reader trusts without opening the ADR: `worldAsOf`'s prefix is rendered `` `asOf <= asOf` ``, which is a tautology as written — both operands have the same name. The code is `priorEvents.filter((event) => event.asOf <= asOf)`, i.e. the prior event's date against the as-of date. ADR-017's own boundary section has it right (`(event) => event.asOf <= asOf`); the row should carry the same form.

apps/tui/src/record-fill-cli.ts
AmetAlvirde: Fixed in 9999b9b. The count was low here too: `record-fill.ts` delegates `io.ask` to **both** `authorLadderTarget` (8 sites) and `resolveFunding` (2 sites), so the census is **19**, not 17 — and the ratifying set is **6**, not 3. `Cash debited [n]` in `record-fill-funding.ts` is a further silent ratification the prose missed. Verified site by site.

apps/tui/src/record-fill-cli.ts
AmetAlvirde: Fixed in 9999b9b — three confirmations, and `:745` moved into the ratification bullet since it records none and the flow continues.

apps/tui/src/record-fill-cli.ts
AmetAlvirde: Fixed in 9999b9b — repointed to `record-fill.test.ts`.

apps/tui/src/record-fill-cli.test.ts
AmetAlvirde: Agreed, and fixed in 47d9310. The whole `it` block is deleted — the count was its only assertion. Confirmed by mutation: replacing the guard with `if (true)` reddens **only** `prompt-channel.test.ts`, while both spawn suites stay green. The false "observable here for the first time" claim is corrected too.

apps/tui/src/prompt-channel.test.ts
AmetAlvirde: Fixed in 47d9310 — tautology removed.

context/adr/ADR-017-a-close-is-the-positions-last-dated-event.md
AmetAlvirde: Fixed in 4585581. The ADR now tracks `fold.ts`'s own sentence: the tie **is** decidable, `order` is in scope in the `PositionClosed` arm, and only the closed book's carriage of it is missing — a siting choice, not an impossibility.

context/adr/ADR-017-a-close-is-the-positions-last-dated-event.md
AmetAlvirde: Fixed in 4585581 — reframed as "**was** consulted at four sites" (the premise the section needs) plus an explicit statement of today's two.

context/adr/INDEX.md
AmetAlvirde: Fixed in 4585581 — now `(event) => event.asOf <= asOf`, with a clause saying what it means.

PR #385, opened 2026-08-19
fix(data-dir): route every dataDir door through one predicate, pinned by one table

### LINKED ISSUE #369: Two data-dir resolvers accept a relative path that the other two refuse — the relative arm of #348's family
Two data-dir resolvers silently accept a **relative** path that the other two refuse loudly, and resolve it against the process CWD.

## The asymmetry

| resolver | relative input `"data"` |
| --- | --- |
| `resolveDataDir` (`packages/engine/src/data-dir.ts`) | **throws** — "must be an absolute path or start with `~/`" |
| `resolveSidecarPath` (`packages/preferences/src/sidecar-io.ts`) | **throws** — "must be an absolute path" |
| `resolvePreferencesPath` (`packages/preferences/src/preferences.ts:55`) | **silently** `<cwd>/data/preferences.jsonl` |
| `resolveEventStorePaths` (`packages/event-store/src/event-store.ts:72`) | **silently** `<cwd>/data/events.jsonl` |

Both permissive resolvers take a default *parameter* and then call `resolve(dataDir)` directly, so nothing validates a caller-supplied value:

```ts
export function resolvePreferencesPath(dataDir = resolveDataDir()): string {
  return join(resolve(dataDir), "preferences.jsonl");
}

export function resolveEventStorePaths(dataDir = resolveDataDirDefault()): EventStorePaths {
  const base = resolve(dataDir);
  // …
}
```

Observed at runtime (not inferred):

```
resolvePreferencesPath('data')      => <cwd>/data/preferences.jsonl
resolveEventStorePaths('data').log  => <cwd>/data/events.jsonl
```

## Why it matters

ADR-006 states the `dataDir` invariant as **absolute and homedir-derived, never CWD-relative**, and requires a relative override to be *rejected loud* — because a durable, git-tracked artifact written relative to whatever directory a process happened to start in is a split-brain ledger. Two resolvers enforce that. Two do not, and one of the two that does not owns `genesis.json` and `events.jsonl`.

## Relationship to #348

#348 fixed the **empty-string** arm of this same family: `""` used to mean "not configured" and fell through to the real ledger, and now it is refused as a misconfigured value while `undefined` still means not-configured. That work extended to these same two resolvers, so after it lands all four agree on `undefined` and on `""`/whitespace.

The **relative-path** arm was deliberately left alone. It is the same family but a different rule, and folding it into a lane already at its merge gate would have grown the change past its warrant. Filing it here so it does not die in a lane report.

## Not urgent

No production call site passes a relative value today — every production call to these two resolvers is a bare no-arg call, so they take their default, which is always absolute. This is a latent contract gap and a drift hazard, not active data loss.

## Suggested shape

Route all four resolvers through one shared validation so the rule cannot drift again, rather than adding a fourth hand-written copy of it. A shared input→outcome table (`undefined` → default, `""`/whitespace → throw, `~`/absolute → normalize, relative → throw) exercised against every resolver would turn a future one-sided edit into a failing test instead of a production surprise. Note the resolvers do not agree on `~` today either — `resolveDataDir` expands it, `resolveSidecarPath` rejects it as non-absolute — so that table needs a deliberate ruling on tilde before it is written, not just a transcription of current behaviour.

AmetAlvirde:
The last ruling on the board. Two data-dir resolvers silently accepted a relative `"data"` and resolved it against the process CWD while two refused it loudly — and one of the permissive two owns `genesis.json` and `events.jsonl`.

## The fifth door

The issue named four. There were five: `resolvePriceFeedPaths` called `resolve()` on its argument **before** delegating to `resolveEventStorePaths`, so `"data"` arrived at the hardened resolver already disguised as `<cwd>/data` — absolute, valid-looking, and a different store on every CWD. That is the same laundering #348 measured for blank, on the relative arm. Folding it in was the difference between fixing the bug and moving it one file over.

## The tilde ruling

The five doors disagreed on `~` **three** ways:

| door | `~/scratch` before | after |
| --- | --- | --- |
| `resolveDataDir` (env) | expands | expands |
| `resolveSidecarPath` | throws — non-absolute | expands |
| `resolvePreferencesPath` | `<cwd>/~/scratch` | expands |
| `resolveEventStorePaths` | `<cwd>/~/scratch` | expands |
| `resolvePriceFeedPaths` | `<cwd>/~/scratch` | expands |

**`~` expands at every door.** The hazard the rule guards is CWD-dependence, and `~/x` has none — it is absolute and homedir-derived, which is ADR-006's invariant verbatim.

The counter-argument, recorded so it is not re-litigated: `~` is a shell affordance, `node:path` has no notion of it, so a TypeScript caller writing it is likelier confused than deliberate. It loses on two counts. The argument boundary is not purely programmatic — `resolvePriceFeedPaths(config.dataDir)` is fed a value that originated as `NUMISMA_DATA_DIR` in a launchd plist, which cannot expand `~` itself; it arrives pre-expanded only because of the order the plumbing runs in, not because of any contract. And decisively: an env-only rule needs an `allowTilde` flag on the shared predicate, making one function carry two behaviours and splitting the table into a column that varies per door — this issue's defect rebuilt inside its own fix.

`~user` is **not** supported syntax and is refused as relative rather than silently becoming `<cwd>/~user`. Expansion is `os.homedir()` at runtime; no `/Users/...` literal enters the tree, in the source or in the tests.

## The shape

One predicate — `normalizeDataDirOverride` in `@numisma/engine` — for every **present** value: blank refused (#348), `~` expanded, absolute normalized, relative refused. The `undefined` → default arm stays per-door on purpose: the five defaults genuinely differ (accumulus root / `resolveDataDir()` / no default at all), and folding that in would mean a parameter whose value varies per door.

Per-door error prose is preserved verbatim through a `DataDirVoice`. #348 deliberately spent that difference — the preferences door names a Reserve floor served from a file nothing writes, the event-store door names a SECOND genesis seeded beside the job's CWD — and a shared predicate must not flatten it. Every existing message assertion still passes unchanged.

One input→outcome table (`@numisma/engine/testkit`) runs against all five doors from their own packages. No single package can import all five — the dependency graph forbids it, correctly — and none needed a new edge, since every package with a door already depends on `@numisma/engine`.

## Verification

`pnpm verify` green: 2546 passed, 51 skipped, typecheck clean, TUI startup smoke passes.

The table was mutation-checked, not just observed passing: reverting `resolveEventStorePaths` to its pre-fix `resolve(dataDir ?? default)` body fails **15 rows**, each naming the input it disagreed on.

## What this closes

Closes #369.

### DIFF
--- a/apps/price-feed/src/paths.test.ts
+++ b/apps/price-feed/src/paths.test.ts
@@ -13,9 +13,10 @@
  * Every fixture is an authored string. Nothing here touches a filesystem — this is pure
  * path algebra — so no test in this file can reach a real data dir.
  */
-import { join, resolve } from "node:path";
+import { dirname, join, resolve } from "node:path";
 import { describe, expect, it } from "vitest";
 import { PRICE_STORE_DIR_SEGMENT } from "@numisma/engine";
+import { assertDataDirContract } from "@numisma/engine/testkit";
 import { resolvePriceFeedPaths } from "./paths.js";
 
 describe("resolvePriceFeedPaths — a BLANK dataDir is REFUSED before it can be laundered", () => {
@@ -88,4 +89,19 @@ describe("resolvePriceFeedPaths — a BLANK dataDir is REFUSED before it can be 
     const spaced = resolve("/tmp/numisma authored root 348");
     expect(resolvePriceFeedPaths(spaced).log).toBe(join(spaced, "events.jsonl"));
   });
+});
+
+// The LAUNDERING door runs the same table as the other four (#369). It is the door the
+// table most needs, because its failure mode is invisible at the boundary it delegates
+// across: it used to `resolve()` its argument first, which turns every value the upstream
+// resolver refuses into one it accepts. `resolve("data")` is `<cwd>/data` — absolute,
+// valid-looking, and a different store on every CWD.
+//
+// It is also the one door with NO `undefined` arm: its `dataDir` is required, so there is
+// no default for a caller to fall into. The table asserts that absence rather than
+// skipping it, so a later edit that adds a default has to come back through here.
+assertDataDirContract({
+  name: "resolvePriceFeedPaths",
+  subject: /a price-feed data directory|NUMISMA_DATA_DIR/,
+  root: (dataDir) => dirname(resolvePriceFeedPaths(dataDir).log),
 });
--- a/apps/price-feed/src/paths.test.ts
+++ b/apps/price-feed/src/paths.test.ts
@@ -13,9 +13,10 @@
  * Every fixture is an authored string. Nothing here touches a filesystem — this is pure
  * path algebra — so no test in this file can reach a real data dir.
  */
+import { homedir } from "node:os";
 import { dirname, join, resolve } from "node:path";
 import { describe, expect, it } from "vitest";
-import { PRICE_STORE_DIR_SEGMENT } from "@numisma/engine";
+import { INBOX_PATH_SEGMENTS, PRICE_STORE_DIR_SEGMENT } from "@numisma/engine";
 import { assertDataDirContract } from "@numisma/engine/testkit";
 import { resolvePriceFeedPaths } from "./paths.js";
 
@@ -100,8 +101,49 @@ describe("resolvePriceFeedPaths — a BLANK dataDir is REFUSED before it can be 
 // It is also the one door with NO `undefined` arm: its `dataDir` is required, so there is
 // no default for a caller to fall into. The table asserts that absence rather than
 // skipping it, so a later edit that adds a default has to come back through here.
+//
+// The root is read back off `pricesDir`, NOT off `log`. `log`, `genesis` and `inbox` are
+// all built by the delegated `resolveEventStorePaths`, so a table driven through any of
+// them measures the EVENT-STORE door's normalization a second time and says nothing about
+// this one. `pricesDir` is the only field this function assembles from its own `base`, so
+// it is the only field through which the table can see this door at all: a regression
+// that fed `join(dataDir, PRICE_STORE_DIR_SEGMENT)` the raw argument would keep every
+// delegated field correct and split the price marks off to `<cwd>/~/scratch/prices`,
+// invisibly, if the table kept reading `log`.
 assertDataDirContract({
   name: "resolvePriceFeedPaths",
-  subject: /a price-feed data directory|NUMISMA_DATA_DIR/,
-  root: (dataDir) => dirname(resolvePriceFeedPaths(dataDir).log),
+  subject: /a price-feed data directory/,
+  root: (dataDir) => dirname(resolvePriceFeedPaths(dataDir).pricesDir),
+});
+
+describe("resolvePriceFeedPaths — all four locations agree on ONE base", () => {
+  // The table drives this door through `pricesDir` alone (see above), which proves this
+  // function's own normalization but not that its own half and its delegated half landed
+  // on the SAME root. That agreement is the whole reason this function exists rather than
+  // each caller assembling four paths itself, and it is exactly what a split would break:
+  // marks written under one root while the log, genesis and inbox live under another is
+  // not a crash, it is a fund whose prices silently stop reaching its positions.
+  //
+  // The two inputs are the two the argument doors used to get wrong. `~/…` is #369's
+  // ruling (three doors produced `<cwd>/~/…`), and a padded absolute path is the trim's
+  // new reach at the argument doors — the one whose halves would disagree if only one of
+  // them normalized.
+  const assertOneBase = (dataDir: string, base: string): void => {
+    const paths = resolvePriceFeedPaths(dataDir);
+    expect(paths.pricesDir).toBe(join(base, PRICE_STORE_DIR_SEGMENT));
+    expect(paths.genesis).toBe(join(base, "genesis.json"));
+    expect(paths.log).toBe(join(base, "events.jsonl"));
+    expect(paths.inbox).toBe(join(base, ...INBOX_PATH_SEGMENTS));
+  };
+
+  it("a `~` path expands once, for the price store and the delegated log alike", () => {
+    // `homedir()` at runtime — never a `/Users/...` literal, which ADR-006 forbids and
+    // which would publish the operator's directory layout in a public repo.
+    assertOneBase("~/scratch/numisma-authored-price-root", join(homedir(), "scratch", "numisma-authored-price-root"));
+  });
+
+  it("a whitespace-padded path is trimmed once, not by one half only", () => {
+    const root = resolve("/tmp/numisma-authored-price-root-369");
+    assertOneBase(`  ${root}  `, root);
+  });
 });
--- a/packages/engine/src/data-dir-contract.testkit.ts
+++ b/packages/engine/src/data-dir-contract.testkit.ts
@@ -148,6 +148,14 @@ export interface DataDirDoor {
    * The door's own voice, so the shared predicate is proven NOT to have flattened the
    * per-door error messages #348 deliberately wrote (a Reserve floor served from a file
    * nothing writes, vs a SECOND genesis seeded beside the job's CWD).
+   *
+   * It must match THIS door's subject and nothing else. An alternation that also admits
+   * another door's wording — `/a sidecar data directory|NUMISMA_DATA_DIR/` — disarms the
+   * assertion against the one regression `DataDirVoice` exists to prevent: a door that
+   * stopped calling `normalizeDataDirOverride` with its own voice and started delegating
+   * to `resolveDataDir({ NUMISMA_DATA_DIR: dataDir })` would emit the ENGINE's message,
+   * every root row would still land on the same path, and the whole table would stay
+   * green while the operator lost the sentence naming which artifact was at stake.
    */
   subject: RegExp;
   /**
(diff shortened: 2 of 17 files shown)


AmetAlvirde (commented):
Review of the shared `dataDir` predicate (#369). **The production change is sound** — I probed `normalizeDataDirOverride` directly against `~`, `~/`, `~//x`, `~/.`, `~/..`, `~\\x`, `" ~ "`, `~user`, `~ /x`, `~/~`, trailing-slash and whitespace-padded absolutes, and every one lands where the ADR-006 ruling says it should. The four affected suites are green (110 tests). Nothing here is a merge blocker.

What I found is concentrated in the **table's teeth**, not in the predicate: three of the findings are about assertions that cannot fail on the regression they name. Details inline.

### Findings that could not be anchored to a diff line

**4. Docs drift — the new `./testkit` subpath export is undeclared in every place that enumerates them (low).** `packages/engine/package.json` now exports four subpaths, but:

- `README.md:31` — "Exposes `.`, `./format`, and `./calendar` subpath exports."
- `README.md:44` and `docs/codebase-map.md:40` — "the **three** deliberate subpath exports (`@numisma/engine/format`, `@numisma/engine/calendar`, `@numisma/event-store/testkit`)"
- `packages/engine/README.md:131` — "**Two** subpath exports exist beside the package root"

That codebase-map paragraph is the canonical list a contributor consults before adding an import, and it is stated as an exhaustive count, so it now reads as "`@numisma/engine/testkit` is an undeclared deep import" — the exact thing the paragraph forbids. `@numisma/event-store`'s equivalent got its README line (`packages/event-store/README.md:41`); this one did not. Nothing enforces the count mechanically, so the suite stays green either way.

**6. `docs/durable-log-ops.md:93` still states the pre-#348 rule (low, pre-existing).** The bullet reads "unset / **empty / whitespace** → the `<fund>` default", which `normalizeDataDirOverride` refuses loudly. Pre-existing drift, not introduced here — but this PR is the one that re-canonicalizes that exact five-row table in ADR-006 and in the predicate's docstring, so it is the natural place to bring the third copy along.

packages/preferences/src/data-dir-doors.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import { resolvePreferencesPath } from "./preferences.js";
+import { resolveSidecarPath } from "./sidecar-io.js";
+
+// The file name is incidental to the table — the door reports the ROOT it resolved, and
+// which leaf a sidecar appends is pinned by that sidecar's own suite.
+const SIDECAR_PROBE_FILE = "orders.jsonl";
+
+assertDataDirContract({
+  name: "resolveSidecarPath",
+  subject: /a sidecar data directory|NUMISMA_DATA_DIR/,
```
AmetAlvirde: **1. The `|NUMISMA_DATA_DIR` alternative makes the "own voice" assertion unable to fail on the regression it names (medium — test strength).**

Same line at `:42`, and at `packages/event-store/src/event-store.test.ts:399` and `apps/price-feed/src/paths.test.ts:105`.

The testkit asserts, at `data-dir-contract.testkit.ts:191`, that "the refusal must be this door's own, not some incidental downstream error" — and that is the check standing between this PR and the flattened-prose failure the whole `DataDirVoice` parameter exists to prevent. With `|NUMISMA_DATA_DIR` in the alternation it cannot do that job for four of the five doors.

Verified two ways:

- I ran all seven refusal inputs from the table (`""`, `"   "`, `"\t\n"`, `~scratch`, `data`, `./data`, `../data`) against all four non-env doors. **Not one** produces a message containing `NUMISMA_DATA_DIR` — the branch is dead against every row the table has.
- The engine's two messages satisfy *every* assertion in the refusal arm: `/a sidecar data directory|NUMISMA_DATA_DIR/` matches `NUMISMA_DATA_DIR must be an absolute path or start with "~/"…` and `NUMISMA_DATA_DIR is set to an empty value…`; so do the blank pattern (`set to an empty value`) and the relative pattern. Same for the preferences, event-store and price-feed regexes.

So: a future edit that regresses `resolveSidecarPath` to `join(resolveDataDir({ NUMISMA_DATA_DIR: dataDir }), fileName)` — dropping this door's voice entirely and speaking the env knob's — keeps **the whole table green**. That matters most here, because `resolveSidecarPath` and `resolvePreferencesPath` have no *other* test pinning their relative-refusal prose: grepping `must be an absolute path` across the repo turns up only `packages/engine/src/data-dir.test.ts` and two `apps/tui` CLI tests. For those two doors this alternation is the only guard on the message, and it does not guard.

Dropping `|NUMISMA_DATA_DIR` from the four non-env doors costs nothing (no row reaches it) and restores the assertion.

apps/price-feed/src/paths.test.ts
```diff
@@ -89,3 +90,18 @@ describe("resolvePriceFeedPaths — a BLANK dataDir is REFUSED before it can be
...
+// resolver refuses into one it accepts. `resolve("data")` is `<cwd>/data` — absolute,
+// valid-looking, and a different store on every CWD.
+//
+// It is also the one door with NO `undefined` arm: its `dataDir` is required, so there is
+// no default for a caller to fall into. The table asserts that absence rather than
+// skipping it, so a later edit that adds a default has to come back through here.
+assertDataDirContract({
+  name: "resolvePriceFeedPaths",
+  subject: /a price-feed data directory|NUMISMA_DATA_DIR/,
+  root: (dataDir) => dirname(resolvePriceFeedPaths(dataDir).log),
```
AmetAlvirde: **2. This adapter measures `resolveEventStorePaths`'s normalization, not price-feed's (low/medium — coverage gap).**

`paths.ts:65` reads `const { genesis, log, inbox } = resolveEventStorePaths(base)`, so `.log` is the **delegate's** arithmetic. The one field price-feed computes from its own `base` is `pricesDir` (`join(base, PRICE_STORE_DIR_SEGMENT)`), and the table never looks at it.

Concrete regression the table would not catch: keep the `normalizeDataDirOverride` guard exactly as written, but let `pricesDir` be built from the raw argument instead of `base` — `join(dataDir, PRICE_STORE_DIR_SEGMENT)`. Every refusal row still throws in price-feed's own voice; every `root` row still returns the right `log` root, because `resolveEventStorePaths` re-normalizes on its own. Meanwhile `resolvePriceFeedPaths("~/scratch").pricesDir` is `<cwd>/~/scratch/prices` — a directory literally named `~`, the exact artifact this PR's ADR-006 amendment says it kills — while the event log correctly lands in `$HOME/scratch`. Marks and log split across two roots, table green.

The existing suite does not cover this either: `"an ABSOLUTE dataDir still passes through untouched"` is the only `pricesDir` assertion, and it passes an already-`resolve()`d root, so it is invariant under the same edit.

`root: (dataDir) => dirname(resolvePriceFeedPaths(dataDir).pricesDir)` is the same value on the happy path and is this door's own arithmetic — it makes the delegate's re-normalization stop covering for it.

packages/engine/src/data-dir-contract.testkit.ts
```diff
@@ -0,0 +1,238 @@
...
+        expect(() => door.root(testCase.input)).toThrow(door.subject);
+        expect(() => door.root(testCase.input)).toThrow(
+          testCase.outcome === "throws-blank"
+            ? /must not be empty|set to an empty value/
+            : /absolute path or start with "~\/"/,
+        );
+      });
+    }
+
+    it("never resolves against the process CWD — the whole point of the table", () => {
```
AmetAlvirde: **3. This `it` cannot fail unless a per-row `it` has already failed (low — dead assertion).**

Its failure set is a strict subset of the per-row tests':

- **`root` rows** are already pinned exactly at `:181` by `expect(door.root(input)).toBe(testCase.root())`. Once that holds, `not.toBe(process.cwd())` and `not.toBe(join(cwd, "data"))` are determined by whether `testCase.root()` itself equals one of them — and the table's five roots are `resolve(homedir())`, two paths under `homedir()`, and two authored `/tmp/numisma-authored-…` literals. None can be the vitest CWD.
- **refusal rows** `continue` on throw, and the per-row test already requires them to throw.

So there is no state where this test is red and the rows above are green. The comment calls it "the class-level statement", which is a fair intent — but as written it is documentation with an `expect` around it, and it re-invokes `door.root` twice per row (once per `forbidden`) to assert nothing new. Worth either deleting, or giving it teeth the rows lack — e.g. asserting the property over inputs that are *not* in the table (a fuzzed set), which would genuinely be a class-level statement rather than a restatement.

packages/engine/src/data-dir.ts
```diff
@@ -22,28 +28,114 @@ function accumulusDataDirDefault(): string {
...
+ * `<cwd>/~/scratch` — a directory literally NAMED `~`, beside wherever the process
+ * started. No door wanted that arm; it was what "call `resolve()` and hope" happened
+ * to do.
+ *
+ * Expansion is `os.homedir()` at RUNTIME. Never a `/Users/...` literal, in this file or
+ * in any test that pins it — ADR-006 forbids the literal, and a committed one would leak
+ * the operator's real directory layout into a public repo.
+ */
+export function normalizeDataDirOverride(raw: string, voice: DataDirVoice): string {
+  const trimmed = raw.trim();
```
AmetAlvirde: **5. The trim is a new normalization at three doors, and it retargets rather than refuses (low — judgment call, likely acceptable).**

`resolveEventStorePaths`, `resolvePreferencesPath` and `resolveSidecarPath` previously did `resolve(dataDir)` with no trim. Measured before/after for a directory whose *name* carries whitespace:

```
resolveEventStorePaths("/tmp/a b ").log
  before → /tmp/a b /events.jsonl
  after  → /tmp/a b/events.jsonl      (verified against this branch)
```

macOS and Linux both allow a trailing space in a directory name, so this is the one input class where the new rule silently points a caller at a **different existing directory** instead of either honoring it or refusing it — the same shape of quiet redirection the PR is otherwise closing, just at a much narrower input.

I read this as deliberate: the table's `"  /tmp/numisma-authored-contract-root  "` row states the trimming rule explicitly, it matches what the env knob has always done, and a padded env value resolving to a sibling store is by far the likelier real-world mistake. Flagging it only because it is a behavior change at three doors that the ADR-006 amendment does not mention — if the trade is intended (I think it is), one clause on that row's `why` saying trimming *retargets* rather than refuses would close it.

AmetAlvirde:
All six review findings addressed in `ae9bae0`, `217cda4`, `92c5815`.

**1. Toothless voice regexes** — dropped the `|NUMISMA_DATA_DIR` alternative from all four non-env doors. Mutation-verified independently: pointing `resolveSidecarPath` at `resolveDataDir({ NUMISMA_DATA_DIR: dataDir })` now fails **7 rows** (3 blank + 4 relative), each on `expected to throw error matching /a sidecar data directory/ but got 'NUMISMA_DATA_DIR must be an absolute …'`. With the old alternation and the same mutation in place: 30 passed, fully green. The finding was exactly right — that assertion was the only guard on `resolveSidecarPath`/`resolvePreferencesPath`'s relative-refusal prose, and it could not fail.

**2. Price-feed measured the wrong door** — the table's adapter now reads `dirname(...pricesDir)`, the only field price-feed builds from its own `base`. Added a `describe` asserting all four returned locations agree on one base, for a tilde input and a padded input.

**3. Dead CWD assertion** — deleted. Its failure set was a strict subset of the refusal rows', which assert the door returns *nothing* — strictly stronger. A comment at that spot records what it claimed and which rows actually hold the guarantee.

**4. Docs drift** — `./testkit` now enumerated in `README.md`, `docs/codebase-map.md`, and `packages/engine/README.md`, with a note that it is test scaffolding deliberately kept out of `index.ts` and the `.` entry point. The engine README's `data-dir.ts` row also gained `normalizeDataDirOverride` / `DataDirVoice`.

**5. The trim's new reach** — documented, not changed. `normalizeDataDirOverride` now carries a block naming the concrete case (`/tmp/a b ` → `/tmp/a b/events.jsonl`, previously a sibling directory ending in a space) and why a retarget is right here: the env knob already landed on the intended root, and a door refusing what the knob accepts would be a fifth disagreement inside the function written to leave none.

**6. `docs/durable-log-ops.md:93`** — split into `unset → default` and `empty/whitespace → rejected loudly (#348)`, naming unsetting as the deliberate way to choose the default.

`pnpm verify` green: 2543 passed, 51 skipped, typecheck clean, openTUI startup smoke passes. Count moves 2546 → 2543 as expected: −5 for the deleted dead case (one per door), +2 for the new price-feed agreement cases.

Issue #372, opened 2026-08-18
Flake (recorded, not chased): schedule-window 'refuses a mark hour that is not an hour' reds once under full parallel run


AmetAlvirde:
**Filed as a recorded observation, not a request to chase it.** The point is that the next lane to see this red has a prior observation to check against instead of spending an hour deciding whether it caused it.

## The observation

`apps/price-feed/src/schedule-window.test.ts`

> `the launchd fetch window > refuses a mark hour that is not an hour, which would classify every run out-of-window`

- **Failed once** in a full parallel `pnpm test` run.
- **Passed 15/15 in isolation** immediately afterwards (`pnpm vitest run apps/price-feed/src/schedule-window.test.ts`).
- **Did not reproduce** on a second full `pnpm test` run — that run was 2287 passed / 48 skipped / 2 failed, the 2 being the known `ops/testkit/gitignored-path-globs` pair.
- Observed from the Lane E worktree (`feature/zero-cost-tier-attribution-fallback`, PR #366), whose diff is confined to `packages/engine/src/events/fold.ts` and `packages/engine/src/fold.test.ts`. **Not plausibly the cause** — the failing assertion is a config-refusal path (`NUMISMA_MARK_HOUR: "noon"` must exit 1 and refuse loudly) that never reaches the engine's fold.

## Corroborated independently by a second lane, before this issue was an hour old

**Lane D observed the same mechanism family in a different harness**, without knowledge of this filing: its armed wrapper harness went **3/55 red under self-inflicted parallel load, and 55/55 alone.**

That is a second lane, a second file, an independent observation, and the same signature — reds only under parallelism, clean in isolation. It moves the "suspected mechanism" below from a hypothesis inferred from one failure's shape to a pattern with two sightings, and it is the strongest argument for measuring the harness rather than re-reading either test.

## Suspected mechanism, unverified

The test **shells out to the real wrapper script** via `runWrapper`, creating temp dirs and spawning a process, then asserts on exit status and output. That family of test is the one most exposed to parallel resource contention under a full run. This is a hypothesis from the failure's shape, not a diagnosis — nobody has instrumented it.

## Why this gets an issue rather than a line in a lane report

[...]

PR #386, opened 2026-08-19
fix(tui): a rejected question is a missing answer, not a dead run

### LINKED ISSUE #370: readline's own error still reaches the operator on two paths #346 did not close — TTY Ctrl-D, and record-fill's eager interface

### LINKED ISSUE #378: docs(ADR-004): the "three stale claims" passage asserts a comment that no longer exists

AmetAlvirde:
What actually closes #370 is the fix below to symptom 1. What actually closes #378 is the ADR-004 rewrite below.

## #370, symptom 1 — a rejected question is a missing answer, not a dead run

At a real terminal `isTTY` is true, so the prompt channel's no-terminal guard never fires. Ctrl-D makes `rl.question()` reject, and before this fix that rejection unwound straight through each shell's outer catch — `import-orders-cli.ts` and `record-fill-cli.ts` both print `error.message` there — putting Node's own wording, `Aborted with Ctrl+D`, on the operator's screen and ending the run. The domain's refusal, `REFUSED — no funding reserve was declared for this batch`, was never reached.

The fix is one line, `rl.question(q).catch(() => "")` in `apps/tui/src/prompt-channel.ts` — clause 4, the second door into the decision clause 2 already made. A rejected question is an ANSWER that did not arrive, not a run that cannot continue, so the channel resolves it with `""`, the same honest answer the no-terminal path already gives, and the domain refuses in its own voice instead of Node.

### The Node contract, measured rather than assumed

node v24.14.0, `node:readline/promises` over an in-process `PassThrough` pair with `terminal: true`, a raw `\x04` byte written to a pending `question()`:

```
name "AbortError", code "ABORT_ERR", message "Aborted with Ctrl+D"
```

The same measurement found the second rejection a caller can provoke: a question put after the interface has closed rejects `ERR_USE_AFTER_CLOSE` / "readline was closed" — the readline internal symptom 2 already removed from the operator's screen. It also found what does NOT reject: plain stream EOF and `rl.close()` leave a pending `question()` unsettled forever rather than rejecting anything.

That's why the catch is wide rather than narrowed to `ABORT_ERR`. The abort closes the interface, so every question asked after it hits `ERR_USE_AFTER_CLOSE`. Narrowing the catch would fix the first question and hand the operator the readline internal on the second — the exact string clause 1 exists to keep off the screen.

### What the tests cover, and what they don't

There is no pty, no spawn and no child process anywhere in `prompt-channel.test.ts`. It pins what the channel does with a `question()` that rejects with the measured shape above — nothing more. It is NOT end-to-end coverage of a real terminal delivering a real Ctrl-D.

That's deliberate. The issue's proposed `script -q /dev/null` harness was measured and rejected earlier: with piped stdin it delivers Ctrl-D to the pty before the child even starts, so a raw `\x04`, a blank line, and a real answer all collapse to the identical `AbortError`. It cannot express the blank-line-versus-Ctrl-D contrast the issue's own argument rests on, and an assertion built on it would pass for the wrong reason.

### The accepted residual — flagging this for review, not burying it

`""` is only safe as an answer while the FIRST question a run reaches unanswerable is one that refuses on it. The no-terminal path only ever reaches the first question, so ordering alone made it safe. Ctrl-D doesn't respect that ordering — it lands wherever the operator presses it, and every question after it also answers `""`, including the questions both shells' own enumeration reads as a silent ratification rather than a refusal: six of `record-fill`'s nineteen, three of the import shell's.

Open question for review: whether a mid-interview Ctrl-D can now reach a durable write with ratified defaults where it previously aborted the run outright. The module docstring records this as an accepted residual. The durable fix is the sentinel both construction sites already name — something the questions reject rather than a blank they can't distinguish from a real answer.

## #378 — ADR-004's correction 2 asserted a comment that no longer exists

ADR-004's "three stale claims" item 2 quoted `preferences.ts`'s docstring as still saying the resolver is "latent today (no runtime caller)", and said "the comment itself is left untouched by this increment." Spec #320 rewrote that docstring — the quoted string exists nowhere in the repo anymore, and the debt the correction described as owed has been paid.

Rewritten to post-#320 tense: what was wrong when the correction was written, and what's true now, kept separate rather than blurred. The chain is renamed by FUNCTION instead of pinned line numbers, so it can't go stale a fourth time: `loadPreferences(resolvePreferencesPath())` inside `loadReserveFloorAsOf`, called by `buildGlanceForAnchor`, called by both `push.ts` and `backfill-core.ts`.

A grep against that chain turned up a correction to the ADR's own claim: `push.ts` and `backfill-core.ts` call `buildGlanceForAnchor` only — neither names `loadPreferences` or `resolvePreferencesPath` directly. `push-core.ts` is the only non-test file under `apps/web/src` that does. The sidecar read is reached through one door, not the three the ADR implied. Agrees with `context/adr/INDEX.md:8`, which is unchanged.

## Verification

- `pnpm vitest run apps/tui/src/prompt-channel.test.ts` — red first (4 failed | 9 passed), then 13 passed.
- `pnpm typecheck` clean across all seven projects.
- Full `pnpm test`: 162 files passed / 3 skipped, 2547 passed / 51 skipped, 29.08s. Neither known red present — `ops/testkit/gitignored-path-globs` (#365/#368) passed, and `apps/price-feed/src/schedule-window.test.ts` (#372) did not flake. The 3 skipped files are the Postgres suites gated on `NUMISMA_TEST_DATABASE_URL`.

Closes #370. Closes #378.

### DIFF
--- a/apps/tui/src/prompt-channel.test.ts
+++ b/apps/tui/src/prompt-channel.test.ts
@@ -20,8 +20,30 @@
  * CI is ubuntu while development is darwin.)
  *
  * So the channel takes `isTTY` as a VALUE and the interface as a FACTORY, and the thing
- * under test here is the channel rather than Node's terminal handling. Symptom 1 remains
- * open in #370 and gets its own decision; nothing below claims to cover it.
+ * under test here is the channel rather than Node's terminal handling.
+ *
+ * WHAT THE SYMPTOM 1 CASES BELOW DO AND DO NOT COVER. Symptom 1 is Ctrl-D at a REAL
+ * terminal, where `isTTY` is true, the guard does not fire, and `rl.question()` rejects.
+ * The channel's answer to that rejection is pinned here — and ONLY the channel's answer.
+ * There is no pty, no spawn and no child process anywhere in this file: nothing below
+ * presses Ctrl-D, and nothing below is end-to-end coverage of a terminal delivering one.
+ * What is pinned is: given a `question()` that rejects with the shape Node really
+ * produces, this channel resolves with `""`.
+ *
+ * THAT SHAPE WAS MEASURED, NOT ASSUMED — node v24.14.0, `node:readline/promises` over an
+ * in-process `PassThrough` pair with `terminal: true`, a pending `question()`, and a raw
+ * `\x04` byte written to the input:
+ *
+ *     name "AbortError", code "ABORT_ERR", message "Aborted with Ctrl+D"
+ *
+ * That message is the string #370 reports the operator seeing, printed by the shell's
+ * outer catch. The same measurement recorded the OTHER rejection a caller can provoke: a
+ * question put after the interface has closed rejects with code `ERR_USE_AFTER_CLOSE`,
+ * message "readline was closed" — the readline internal symptom 2 removed from the
+ * operator's screen. Both are reproduced by the authored fakes below. (It also recorded a
+ * non-rejection: plain stream EOF and `rl.close()` leave a pending `question()` unsettled
+ * forever rather than rejecting. Nothing here depends on that, but it is why a bare
+ * `.catch()` is the whole fix and no timeout is part of it.)
  *
  * Every fake is authored. Nothing here touches `process`, stdin, or a real readline.
  */
@@ -52,11 +74,57 @@ function fakeInterface(answers: string[]): PromptInterface & {
   };
 }
 
+/**
+ * Node's Ctrl-D rejection, reproduced from the measurement in this file's header rather
+ * than from the docs: an `AbortError` carrying `ABORT_ERR` and Node's own wording. This
+ * is the error that reached the operator through the shells' outer catch in #370.
+ */
+function abortedWithCtrlD(): Error {
+  const error: Error & { code?: string } = new Error("Aborted with Ctrl+D");
+  error.name = "AbortError";
+  error.code = "ABORT_ERR";
+  return error;
+}
+
+/** The other measured rejection: any question put after the interface has closed. */
+function usedAfterClose(): Error {
+  const error: Error & { code?: string } = new Error("readline was closed");
+  error.code = "ERR_USE_AFTER_CLOSE";
+  return error;
+}
+
+/**
+ * A readline stand-in that behaves the way the measured one does once Ctrl-D lands: the
+ * question in flight rejects with the abort, and every question after it rejects with
+ * `ERR_USE_AFTER_CLOSE`, because the interface is closed from that moment on. `close()`
+ * still counts, because the shell calls it in its `finally` either way.
+ */
+function ctrlDInterface(): PromptInterface & { asked: string[]; closed: number } {
+  const asked: string[] = [];
+  const state = { closed: 0 };
+  return {
+    asked,
+    get closed() {
+      return state.closed;
+    },
+    question: async (query: string) => {
+      asked.push(query);
+      throw asked.length === 1 ? abortedWithCtrlD() : usedAfterClose();
+    },
+    close: () => {
+      state.closed += 1;
+    },
+  };
+}
+
 /** The channel plus the two things a caller can observe about how it was built. */
-function channelOn(isTTY: boolean, answers: string[] = []) {
+function channelOn(
+  isTTY: boolean,
+  answers: string[] = [],
+  rl: PromptInterface & { asked: string[]; closed: number } = fakeInterface(answers),
+) {
   const errors: string[] = [];
   let built = 0;
-  const rl = fakeInterface(answers);
   const channel = createPromptChannel({
     isTTY,
     createInterface: () => {
@@ -177,5 +245,64 @@ describe("on a stdin that is no terminal the channel says so and returns an empt
     harness.channel.close();
 
     expect(harness.rl.closed).toBe(0);
+  });
+});
+
+describe("on a terminal where the question is aborted the channel ends the ANSWER, not the run", () => {
+  // #370, SYMPTOM 1. At a real terminal `isTTY` is true, so the no-terminal guard never
+  // fires; Ctrl-D makes `rl.question()` reject, and before this decision that rejection
+  // unwound through the shell's outer catch — `import-orders-cli.ts` and
+  // `record-fill-cli.ts` both print `error.message` there — putting Node's own wording,
+  // "Aborted with Ctrl+D", on the operator's screen and ending the run. The consequence
+  // the issue names: the domain's refusal, `REFUSED — no funding reserve was declared for
+  // this batch`, was never reached. Resolving with `""` is the same answer the
+  // no-terminal path gives, and it lets the domain refuse in its own voice.
+  it("RESOLVES with '' when the question rejects with Node's measured Ctrl-D abort", async () => {
+    const harness = channelOn(true, [], ctrlDInterface());
+
+    await expect(harness.channel.ask("funding reserve? ")).resolves.toBe("");
+
+    // It really did reach readline — this is the terminal path, not the guard.
+    expect(harness.built).toBe(1);
+    expect(harness.rl.asked).toEqual(["funding reserve? "]);
+  });
+
+  // Ctrl-D closes the interface, so the interview does not stop at the aborted question:
+  // whatever the flow asks next reaches a closed readline and rejects with
+  // `ERR_USE_AFTER_CLOSE` — the readline internal #370's symptom 2 took off the
+  // operator's screen. It must not come back through this door. Every later question
+  // answers `""` too, which is what carries a nineteen-question interview all the way to
+  // the domain's refusal instead of ending it at the first one.
+  it("keeps answering '' for the rest of the interview once the interface is closed", async () => {
+    const harness = channelOn(true, [], ctrlDInterface());
+
+    expect(await harness.channel.ask("first? ")).toBe("");
+    expect(await harness.channel.ask("second? ")).toBe("");
+    expect(await harness.channel.ask("third? ")).toBe("");
+
+    expect(harness.rl.asked).toEqual(["first? ", "second? ", "third? "]);
+    // Still one interface. An abort is not a reason to build another one.
+    expect(harness.built).toBe(1);
+  });
+
+  // Nothing is written to the error sink: the abort is not the shell's news to report.
+  // The domain's refusal is the message the operator gets, and it arrives on its own.
+  it("says nothing of its own about the abort", async () => {
+    const harness = channelOn(true, [], ctrlDInterface());
+
+    await harness.channel.ask("a? ");
+
+    expect(harness.errors).toEqual([]);
+  });
+
+  // The shells close in a `finally`, so `close()` runs on this path too. It was built,
+  // so it closes — once, and without throwing, even though readline is already closed.
+  it("still closes the interface it built", async () => {
+    const harness = channelOn(true, [], ctrlDInterface());
+
+    await harness.channel.ask("a? ");
+    harness.channel.close();
+
+    expect(harness.rl.closed).toBe(1);
   });
 });
--- a/apps/tui/src/prompt-channel.test.ts
+++ b/apps/tui/src/prompt-channel.test.ts
@@ -306,3 +306,62 @@ describe("on a terminal where the question is aborted the channel ends the ANSWE
     expect(harness.rl.closed).toBe(1);
   });
 });
+
+/**
+ * THE ABANDONMENT LATCH — the fact `""` cannot carry.
+ *
+ * Clause 4 must resolve `""` (that is what lets the domain refuse in its own voice), and
+ * `""` is a RATIFICATION at several questions of both shells' interviews. The string
+ * therefore cannot tell a flow whether an answer was typed or abandoned, so the channel
+ * remembers it beside the string. What the flow does with it is the flow's business and
+ * is pinned where that behaviour lives — `import-orders.test.ts` drives
+ * `importBitgetOpenOrders` with an abandoning `ask` and asserts nothing is appended.
+ * What is pinned HERE is only the latch: when it is set, when it is not, and that it
+ * never clears.
+ */
+describe("the channel remembers that a question was abandoned", () => {
+  it("is false before anything is asked", () => {
+    expect(channelOn(true).channel.aborted()).toBe(false);
+  });
+
+  it("is false after a question that was actually answered", async () => {
+    const harness = channelOn(true, ["reserve-a"]);
+
+    expect(await harness.channel.ask("funding reserve? ")).toBe("reserve-a");
+
+    expect(harness.channel.aborted()).toBe(false);
+  });
+
+  it("latches on the rejecting question, WITHOUT changing what it resolves", async () => {
+    const harness = channelOn(true, [], ctrlDInterface());
+
+    // Both halves in one case on purpose: a latch that suppressed clause 4's `""` would
+    // put `Aborted with Ctrl+D` back on the operator's screen, which is the whole defect
+    // #370 removed.
+    expect(await harness.channel.ask("funding reserve? ")).toBe("");
+    expect(harness.channel.aborted()).toBe(true);
+  });
+
+  it("never clears for the rest of the interview", async () => {
+    const harness = channelOn(true, [], ctrlDInterface());
+
+    await harness.channel.ask("first? ");
+    await harness.channel.ask("second? ");
+    await harness.channel.ask("third? ");
+
+    expect(harness.channel.aborted()).toBe(true);
+  });
+
+  // The no-terminal path never asks readline anything, so no question was ABANDONED —
+  // it was unaskable. That path is governed by each shell's ordering argument, and its
+  // first question refuses on `""` before any write door is reached. Latching here would
+  // report an abandonment that did not happen and would attribute the wrong refusal to a
+  // piped run.
+  it("stays false on a stdin that is no terminal", async () => {
+    const harness = channelOn(false);
+
+    await harness.channel.ask("a? ");
+
+    expect(harness.channel.aborted()).toBe(false);
+  });
+});
--- a/apps/tui/src/prompt-channel.ts
+++ b/apps/tui/src/prompt-channel.ts
@@ -5,10 +5,10 @@
  * It exists because the shape #346 gave the import shell was right and was NOT shared:
  * `record-fill-cli.ts` still built its interface at module scope, so a piped run printed
  * `readline was closed` — a readline internal, verbatim, to the operator (#370, symptom 2).
- * Two shells needing the identical three decisions is a real seam rather than a
- * hypothetical one, so the mechanism moved here and the shells kept their wiring.
+ * Two shells needing the identical decisions is a real seam rather than a hypothetical
+ * one, so the mechanism moved here and the shells kept their wiring.
  *
- * The three decisions, all of which were previously duplicated or missing:
+ * The decisions, all of which were previously duplicated or missing:
  *
  *   1. THE INTERFACE IS BUILT AT THE FIRST QUESTION, NEVER AT STARTUP. `createInterface`
  *      eagerly consumes stdin, so constructing it up front ends a piped stream before any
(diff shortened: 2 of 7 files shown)


apps/tui/src/prompt-channel.ts
```diff
@@ -93,7 +129,9 @@ export function createPromptChannel(io: PromptChannelIo): PromptChannel {
         return Promise.resolve("");
       }
       rl ??= io.createInterface();
-      return rl.question(question);
+      // CLAUSE 4. A rejected question is an ANSWER that did not arrive, not a run that
+      // cannot continue. See the module docstring for why the catch is this wide.
+      return rl.question(question).catch(() => "");
```
AmetAlvirde: **HIGH — a mid-interview Ctrl-D now reaches a durable write in `orders:import`. This answers the PR body's own open question: yes, and here is the sequence.**

The PR body flags this for review rather than burying it, which is the right instinct — but the answer is not "residual", it is a reachable durable write, and I confirmed it by execution rather than by reading the docstring.

**The sequence, at a real terminal, with `orders:import`:**

1. `Funding reserve for this batch: ` (`import-orders-funding-declaration.ts:69`) — the operator answers with a **real reserve id**. This is the gate the whole safety argument rests on, and it is now **behind** the operator.
2. `Override the funding reserve for any individual order? [y/N] ` (`import-orders-funding-declaration.ts:75`) — the operator presses **Ctrl-D** here. Before this line, that rejection unwound to `import-orders-cli.ts:134`, printed `Aborted with Ctrl+D`, exit 1, **nothing written**. After this line it resolves `""` → `isAffirmative("")` is false → "no overrides" → the flow continues.
3. `Rung picks proposed by price: … Accept all? [Y/n] ` — the interface is closed by the abort, so this rejects `ERR_USE_AFTER_CLOSE` and is likewise caught to `""`. `import-orders-rung-picks.ts:158-161` reads that blank as **"A BLANK LINE IS ACCEPTANCE"** and returns every proposal.
4. There is **no further gate**. `import-orders.ts:671` — `await io.appendOrders(io.ordersPath, fresh)`.

**Measured, not argued.** I drove `importBitgetOpenOrders` with an `ask` that answers once and then returns `""` forever (exactly what the channel now does post-abort):

- questions asked: **3** — the two above plus the rung-pick prompt;
- `appendOrders`: called **once, with 2 records**; the sidecar exists on disk with 2 lines;
- outcome `{"status":"imported","appended":2}`;
- and the written lines carry not just `fundingReserveId` but **`planId` + `rungId` joins that were never declared** — the blank at step 3 ratified the price-matched proposals into append-only lines the operator never saw.

That last part is why this is worse than a bare write. `import-orders-rung-picks.ts:178-181` states the rule this violates verbatim: *"RE-ASKED UNTIL IT IS A CHOICE, never guessed. An unreadable answer resolved to the default would write a durable join into an append-only file that nobody declared, and the operator would have no way to know which reading was taken."*

**The concrete operator scenario.** The operator types the funding reserve, sees the override question, realises the reserve they just typed is the wrong one, and presses Ctrl-D to bail. Previously: run over, nothing on disk. Now: the whole batch is imported and durably attributed to the wrong reserve, with fabricated rung joins, in an append-only file. A *bogus* reserve id would be caught by `checkFundingCoverage` (`unattributed`); a **valid but wrong** reserve passes coverage and lands. The refusal message for the adjacent case even says *"fix the declaration rather than the file"* — which is no longer possible once the line is appended.

**`record-fill` is NOT affected, and it is worth saying why**, because it is the reason this is one finding rather than two: every write there is behind `Write BOTH? [y/N]` at `record-fill.ts:900`, and `isAffirmative("")` is false, so a post-Ctrl-D run races through the remaining questions ratifying defaults and then abandons at the final gate. `writeLogImage`/`appendOrders`/`appendReconciliation` all sit after that line. The import shell has no equivalent gate — that asymmetry is the defect.

**On the trade.** Symptom 1 as #370 describes it is a *cosmetic* defect: the wrong wording on stderr on a run that wrote nothing. What lands in its place is a durability defect. For a Ctrl-D on question **one** the fix is a clear win and does exactly what #370 asked. For a Ctrl-D anywhere after that it is a strict regression, and the shells cannot tell the two apart because `""` is not distinguishable from an answer.

**The remedy is the one the docstring already names, and it is not deferrable.** The sentinel — something the questions reject rather than a blank — is what makes this correct. Alternatively the channel can latch "aborted" after the first rejection and have `ask` keep returning a value the flows refuse on, so the abort is sticky rather than silently ratified. Either way the fix belongs with this change, not after it.

apps/tui/src/prompt-channel.ts
```diff
@@ -31,17 +31,53 @@
...
+ * any question in the interview and every question after it. Both shells enumerate their
+ * questions and both name the ones that read `""` as a silent ratification rather than a
+ * refusal — six of `record-fill`'s nineteen, three of the import shell's. Those are now
+ * reachable by Ctrl-D, and the sentence "only the first is ever reached" is true of the
+ * no-terminal path only. This is the accepted residual of #370: it buys the domain's
+ * refusal at the first question, which is where Ctrl-D is actually pressed, at the price
+ * of a mid-interview Ctrl-D ratifying a default. The durable fix is the one both
+ * construction sites already name — a sentinel the questions reject rather than a blank —
+ * and it belongs to whichever increment makes an unanswered question distinguishable from
+ * an empty one. Each shell states the dependency at its own site and each pins it, because
```
AmetAlvirde: **MEDIUM — the residual is recorded here, but not at the two sites that own what `""` means, and one of them predicted this exact failure and told you to fix it in the same commit.**

This line says *"Each shell states the dependency at its own site and each pins it."* Neither site was touched by this PR, and both now read as safety arguments that are no longer complete.

`import-orders-cli.ts:51-57`, unchanged:

> The last three are ratifications. A no-terminal run never reaches them ONLY because funding is asked FIRST and refuses the whole batch before any rung question is put. **Reorder the interview** … and this same `""` silently ratifies every default and **the run WRITES, unattended, with nobody having answered anything.** If that ordering ever moves, this must become a refusal that the domain cannot mistake for consent (a sentinel the questions reject, not a blank), **and it must move in the same commit.**

That paragraph describes the outcome I confirmed empirically, almost word for word. It conditioned the hazard on a *reorder* — but a reorder is not the only way to reach those three questions with a blank, and clause 4 is the other way. The site's own stated trigger has fired by a route it did not anticipate, and its instruction ("it must move in the same commit") was not followed.

`record-fill-cli.ts:68` has the same shape — *"Only the first is ever reached without a terminal."* Still literally true as scoped to the no-terminal path, but a reader arriving at that construction site to reason about whether `""` is safe now gets an argument that only covers half the doors.

The docstring here is careful and honest about the widening. The problem is placement: the module that widened the hazard cannot see either shell, and the two files that *can* still carry the pre-widening argument. Whatever is decided about the finding on line 134, these two sites need to say that Ctrl-D is a second door into their ratifying questions — otherwise the next reader re-derives the old safety proof and believes it.

apps/tui/src/prompt-channel.test.ts
```diff
@@ -179,3 +247,62 @@ describe("on a stdin that is no terminal the channel says so and returns an empt
     expect(harness.rl.closed).toBe(0);
   });
 });
+
+describe("on a terminal where the question is aborted the channel ends the ANSWER, not the run", () => {
```
AmetAlvirde: **MEDIUM — this block pins the channel, and the channel is not where the behaviour changed. The consequence is pinnable today, without a pty.**

I want to be clear that this suite is *honest*: the header states plainly that there is no pty, no spawn, and that nothing here is end-to-end coverage of a real terminal delivering a real Ctrl-D. I checked every assertion in the new block and none of them can pass for the wrong reason — and the `usedAfterClose()` fake is genuinely load-bearing, because narrowing the catch to `ABORT_ERR` (the thing the docstring warns against) makes *"keeps answering '' for the rest of the interview"* fail at the second question. That is a well-built pin.

The gap is what is *not* pinned. What this change actually alters is not "the channel returns `''`" — it is "an `orders:import` run that would have ended now appends to `orders.jsonl`". Nothing anywhere asserts on that, and #370's argument for why ("only *honestly* fixable behind a pty; neither is reachable from a spawned run") does **not** cover it. The hazard sits at the *flow* seam, not the terminal seam, and that seam takes `ask` as an injected function:

```ts
// no pty, no spawn, no child process
let n = 0;
const ask = async () => (n++ === 0 ? "reserve-a" : ""); // answered once, then aborted
const outcome = await importBitgetOpenOrders({ csvPath, io: { ...io, ask } });
expect(appendOrders).not.toHaveBeenCalled(); // currently FAILS: called once, 2 records
```

That is the assertion this change needs, and it is three lines against a seam `import-orders.test.ts` already drives. It is also the assertion that would have surfaced the finding on `prompt-channel.ts:134` before the PR was opened.

Worth noting against the #363/#365 precedent the issue cites: those were assertions that *could not fail*. This is the inverse failure — a behaviour change with no assertion at all at the layer where it matters — and it is the more expensive of the two, because the layer that *is* covered looks green.

context/adr/ADR-004-preferences-sidecar.md
```diff
@@ -199,29 +199,40 @@ Each was verified against the code before being corrected here.
...
+   reading it live. The correction recorded the claim as false and deliberately
+   **left the comment standing** — a one-line fix owed to the next change that
+   opened the file.
+
+   **That debt is PAID (2026-08-13, spec #320)**, so this item is now a record of
+   a correction that has landed rather than one still owed. Slice 1 opened
+   `preferences.ts` for a reason of its own and rewrote the docstring; it now
+   states the read path is **ALREADY WIRED** and the split-brain hazard live, not
+   hypothetical, and its test-side twin says the same. The string this item
+   originally quoted no longer exists anywhere in the repo — quoted above only as
```
AmetAlvirde: **LOW — this sentence is literally contradicted by `context/adr/INDEX.md:8`, which is the one file the rewrite claims agreement with.**

*"The string this item originally quoted no longer exists anywhere in the repo."* Repo-wide, `latent today` returns three hits: this ADR's own two self-quotes at `:202` and `:204` (which the sentence excuses with "quoted above only as the historical claim"), and **`context/adr/INDEX.md:8`**, which still carries *"(2) `preferences.ts`'s "latent today (no runtime caller)" is **false**"* and is not excused.

The substantive claim — the string is gone as an actual source comment — is true; `packages/preferences/src/preferences.ts` contains no "latent" anywhere. Narrowing to "no longer exists **in the source**" or "as a comment" makes it exactly true and costs two words. Flagging it because the whole point of #378 is an ADR asserting a string state that had stopped being true, and this rewrite reintroduces a smaller version of the same thing.

Everything else in this item verified clean against the code — see the summary comment.

AmetAlvirde (commented):
## Review — fresh pass, formed from the diff rather than the PR body

Four findings, most severe first: **1 high, 2 medium, 1 low**, posted inline. Verdicts on the two things the PR body itself asks review to settle:

**A. Does the fix trade a cosmetic defect for a durability hazard? — YES, in `orders:import`, and I confirmed it by execution rather than by argument.** A Ctrl-D at any question after the funding declaration resolves `""`, the rung-pick prompt reads a blank as acceptance, and `import-orders.ts:671` appends — I drove the flow with an `ask` that answers once then returns `""` forever and got `appendOrders` called once with 2 records, outcome `imported / appended: 2`, carrying `planId`/`rungId` joins nobody declared. Pre-PR that same keystroke ended the run with nothing on disk. `record-fill` is safe, and for a specific reason: every write there sits behind `Write BOTH? [y/N]` (`record-fill.ts:900`), on which `isAffirmative("")` is false. The import shell has no equivalent gate. Detail and the operator scenario in the finding on `prompt-channel.ts:134`.

**B. Is the provenance honest? — YES, entirely, and it survived independent re-measurement.** I did not take the measured contract on faith; I re-ran it on node v24.14.0, `node:readline/promises` over a `PassThrough` pair with `terminal: true`, and every claimed value reproduces exactly:

| claim | re-measured |
| --- | --- |
| raw `\x04` on a pending `question()` | `AbortError` / `ABORT_ERR` / `Aborted with Ctrl+D` ✓ |
| question after the abort | `ERR_USE_AFTER_CLOSE` / `readline was closed` ✓ |
| question after explicit `close()` | `ERR_USE_AFTER_CLOSE` ✓ |
| plain stream EOF | never settles ✓ |
| `rl.close()` while pending | never settles ✓ |

The authored fakes are faithful: the real abort is `instanceof Error` with own properties exactly `message`/`code`/`name`, which is what `abortedWithCtrlD()` builds. I also checked the one claim the fakes *cannot* exhibit — the test comment on "still closes the interface it built" asserts real readline tolerates `close()` after an abort — and it does, twice over, no throw. Nothing in the suite or docstring implies pty or end-to-end coverage that does not exist; the header says the opposite, in plain words, twice. Suite is green (13 passed).

**On the wide `.catch(() => "")` — I went in expecting to object and came out agreeing with it.** Narrowing to `ABORT_ERR` really would leak `readline was closed` on the next question, which I confirmed. I then went looking for a real failure the wide catch hides and could not find one: EOF, `close()` while pending, and a destroyed output stream all leave the question **unsettled** rather than rejecting, so they never reach the catch at all. The two measured rejections are the realistic population, and both mean "no answer arrived". The width is defensible as written. What is not defensible is what `""` then *means* downstream — that is finding 1, and it is a problem with the sentinel, not with the catch.

**The ADR half verified clean, with one exception (finding 4).** Checked by grep against the code, not against the prose:

- `resolvePreferencesPath` is exported (`preferences.ts:69`); the docstring really does say **ALREADY WIRED** and "live, not hypothetical" (`:48-53`), and the test-side twin exists and says the same (`preferences-reliable.test.ts:539-542`) — that claim is not padding.
- The chain holds exactly as named: `loadPreferences(resolvePreferencesPath())` at `push-core.ts:202`, lexically inside `loadReserveFloorAsOf` (`:199-207`); `buildGlanceForAnchor` (`:250`) calls it at `:254`; `push.ts:95` and `backfill-core.ts:271` both call `buildGlanceForAnchor`.
- **"One door, not three" is exactly true**, and stronger than the ADR claims. Under `apps/web/src` only `push-core.ts` (non-test) and `discard-channel.test.ts` name either symbol. Repo-wide the only other runtime mention is inside the owning package's own `seedDefaultPreferences`, which has no runtime caller at all. `apps/tui`, `apps/price-feed` and `ops/` name neither. The `apps/web/src` scoping is conservative rather than misleading — good correction to make.
- **No stale line pin was reintroduced.** Item 2 contains zero `file:NNN` pins; the only one left in the document is `:73` in item 1, explicitly marked historical.
- **It still reads as a correction.** The original claim is quoted as history, stated as having been false and why, and the deliberate deferral ("left the comment standing — a one-line fix owed") survives. One clause did not: main's "which changes no code at all", which recorded that the correcting increment was documentation-only. Minor, mentioning it only so the drop is a choice rather than an accident.
- ADR-020's fifth clause citation checks out (`ADR-020…:60-65`), and the code ordering matches (`push-core.ts:546-568`: upsert first, then emit).
- Agrees with `INDEX.md:8` in substance. The index still opens present-tense ("is **false**") where the body now says "WAS false"; its own parenthetical resolves it, so no contradiction — noting it, not asking for it.

**Recommendation.** The ADR half is ready modulo a two-word narrowing. The TUI half should not land as-is: finding 1 is a durable, append-only write that a single keystroke now reaches where it previously could not, and the remedy is the sentinel both construction sites already name. `record-fill` being accidentally safe behind its final `[y/N]` gate is not a property the import shell shares, and nothing currently pins that it must.

Judgment-call disclosure, so the next agent can weigh rather than just act: finding 2 is a documentation-placement call and finding 4 is a wording nitpick — both are cheap and neither blocks. Findings 1 and 3 are defects and I would hold the merge on 1.

AmetAlvirde:
## Review addressed — all four findings, at the seams they were raised against

Two commits: `e54b8f4` (TUI) and `e1d1e12` (ADR).

### HIGH — `prompt-channel.ts:134` — a mid-interview Ctrl-D reached a durable write

Fixed as an **abort latch plus a refusal at the write door**, not by narrowing the catch and not by the sentinel.

- **Latch** — `prompt-channel.ts`: `PromptChannel.aborted()`, a closure over the channel's own interview, set the first time `question()` rejects and never cleared. It does **not** change what `ask` resolves: clause 4 stands, so the first rejecting question still comes back `""` and the domain still refuses in its own voice. The no-terminal path does not set it — nothing was abandoned there, it was unaskable.
- **Refusal** — `import-orders.ts`: new `OrdersImportIo.promptAbandoned?: () => boolean` (optional, so every existing caller and fixture is untouched), checked immediately before `io.appendOrders` and refused through the existing `reject()` with a new `OrdersImportRejection` member **`interview-abandoned`**. The operator gets the same `REFUSED — … Nothing was written to <path>` contract `write-failed` already gives, and the second sentence is literally true at that point.
- **Wiring** — `import-orders-cli.ts` passes `promptAbandoned: () => prompt.aborted()`.
- **Why the door and not the questions**: the guarantee is positional-independent — *if the terminal was abandoned, nothing is written* — so it holds regardless of which question caught the Ctrl-D, **including the last question of an interview**, which a latch that only re-answered later questions would still leak.
- **The sentinel is still owed, and the docstring says so** rather than implying the ordering dependency is gone: what the latch does not buy is an unanswered question distinguishable from an empty one *at the question* — the ratifications in between still read the blank as consent, and the operator still watches the batch be ratified before the door refuses it. A sentinel would move the refusal to the question actually abandoned and remove the dependency instead of surviving it.

### MEDIUM — `prompt-channel.ts:72` — the residual documented in the wrong place

Both construction sites updated in their own voice, each readable standalone:

- **`import-orders-cli.ts`** — after its existing enumeration: Ctrl-D is a **second door into those three ratifications that is not a reorder**; the site's own predicted trigger fired by a route it did not anticipate; and what stops the write anyway (latch → `promptAbandoned` → `interview-abandoned` before `appendOrders`), noting the guarantee survives a reorder even though the paragraph above it does not.
- **`record-fill-cli.ts`** — same second-door statement over its six ratifications, then what saves it: `Write BOTH? [y/N]` (`record-fill.ts:900`), `isAffirmative("")` false. **Verified before writing it down**, as asked: `writeLogImage` (`:907`) and `appendOrders` (`:915`) are inside the two-file act just past the gate, and `appendReconciliation` (`:492`) is reachable only via `reconcileRecordedFill`, called once at `:956` after the act is already durable — nothing writes ahead of the gate. `Confirm these derived verdicts? [y/N]` (`:736`) abandons on a blank too, so most abandoned runs stop earlier still. The site names the re-trigger: move a writer ahead of that gate, or rephrase it so silence means yes, and this act must take the latch too.
- The shared module's docstring no longer records this as an accepted residual; it now names both mechanisms and keeps the sentinel as outstanding work.

### MEDIUM — `prompt-channel.test.ts:251` — unpinned at the flow seam

Pinned where the behaviour changed, in `import-orders.test.ts`, in that file's voice — no pty, no spawn:

- `refuses as interview-abandoned and never reaches appendOrders` — exactly the reviewer's construction (`ask` answers `reserve-a` once, then `""` forever with the latch set), asserting `appendOrders` is **never called**, the outcome is `interview-abandoned`, the `REFUSED …/Nothing was written` line reached the operator, and the interview really got past the refusing question (so it cannot pass for the reason the adjacent `no-reserve-declared` case already covers). **Mutation-checked**: with the write-door check disabled it fails on `appendOrders` having been called.
- `still lets a Ctrl-D at the FIRST question refuse as no-reserve-declared` — #370's fix, unregressed.
- `writes normally when the terminal was never abandoned` — present-and-false behaves as absent.
- A channel-level latch suite stays in `prompt-channel.test.ts` (five cases: unset before asking, unset after a real answer, latches without changing the resolved `""`, never clears, stays false with no terminal).

### LOW — `ADR-004:216` vs `INDEX.md:8`

Narrowed to *"no longer exists **in the source**"*, with the quotation's survival stated positively: it lives on in the ADR record — quoted above as the claim being corrected, and in `INDEX.md`'s summary of the same correction — *"That is the record doing its job, and it stays."* Nothing removed from `INDEX.md`.

### Verification

- `pnpm vitest run apps/tui/src/prompt-channel.test.ts apps/tui/src/import-orders.test.ts` — **84 passed** (18 + 66).
- `pnpm typecheck` — clean across all seven projects.
- `pnpm test` — **162 files passed, 3 skipped; 2555 tests passed, 51 skipped**. No failures; neither known red appeared.

PR #387, opened 2026-08-19
docs(coverage-rationale): make §7's completeness claim true again

### LINKED ISSUE #377: docs(coverage-rationale): §7's table omits three `packages/preferences/src` files that measure below 100%

AmetAlvirde:
What actually closes #377 is §7's completeness contract being made true again: "every other currently measured file below 100% that the earlier sections do not individually name" is now a true sentence, because every file with nonzero uncovered lines or branches has been diffed against the document text with zero missing.

## What the issue asked, and what the deliverable actually is

#377 named three unrowed files in `packages/preferences/src` — `sidecar-io.ts`, `plans.ts`, `reconciliations.ts` — and asked for rows. But the real ask is §7's own completeness contract, not three rows in isolation. Getting that contract true again required treating it as a full audit: every file with nonzero uncovered lines or branches, checked against what the document actually says about it.

## Measured fresh at `d5fe02c`, not copied from the issue

The issue's table was taken at `7269a9f`, and by `d5fe02c` every number in it had drifted. The three named files, measured fresh:

- `sidecar-io.ts` — 92.47% (`:152-153,200-205`)
- `plans.ts` — 96.39% (`:160-161,203-204,206-207,301-302,325-326,391-392`)
- `reconciliations.ts` — 86.66%, 12 ranges — the same count as the issue's, not one more. `:524-526` is real and IS a round-trip self-check in `appendReconciliation`, distinct from the eleven per-field rejection ranges — but it is not new: it is the issue's own `:521-523`, shifted by the same +3 every other range in this row shifted by. (This row was later re-derived from scratch and corrected further — see the addendum below.)

## The churn this repairs

Roughly 22 existing rows had gone stale and were repaired against the fresh run. Most were number-only drift; the ones where the CONTENT changed — the ones a reader would otherwise trust wrongly — are called out specifically:

- `import-orders.ts` — 93.75%→89.21%, ranges reshuffled: the old `unreadable-sidecar-lines` and funding-declaration `io.err` citations are covered now, replaced by new `loadInForceLadders` and `declaredJoinsOnFile` gaps.
- `glance/verdict.ts` — the precedence-sort branch is covered; a new `referenceLabel` throw is not.
- `push/glance.ts` — suppression-list branches covered; new `?? []` default-array gaps.
- `orders/ingest.ts` — the uncovered field flipped from `timeInForce` to `orderType`.
- `fetch-prices.ts` — branch 77.5%→91.93%.
- `push/fixture-synthesis.ts` — 5→14 ranges.
- `binance-provider.ts` and `twelvedata-provider.ts` — each gained a range.
- `record-fill.ts` — 79.76%→85.19%.

## Ten rows added for files that had none

`import-orders-rung-picks.ts` plus nine more found during the completeness sweep: `fold-lines.ts`, `calendar.ts`, `unattended-report.ts`, `convexity-caption.ts`, `dca-block.ts`, `fill-path.ts`, `plans-cli.ts`, `operator-notice-cli.ts`, `binance-spot.ts`.

## Three rows that report a dishonest 0% — flagged, not fixed

This is the part a reviewer must see. `plans-cli.ts` and `operator-notice-cli.ts` are self-executing top-level scripts of exactly the shape §1 already excludes (`import-orders-cli.ts`, `record-fill-cli.ts`, `cancel-order-cli.ts`, `price-feed/cli.ts`) but they are NOT excluded in `vitest.config.ts`, so they report 0% instead of the honest "excluded, here's what guards it instead." `lib/binance-spot.ts` is a browser-only React hook, uninstrumented for the same reason §6 already gives for the `.tsx` components — except it's a `.ts` file, so it falls inside the include glob instead of being caught by the extension boundary.

All three are named in the document as not fitting the defensive/unreached-fixture/real-branch taxonomy, rather than forced into a label they don't earn. The actual fix for the first two is a `vitest.config.ts` exclude entry — a SOURCE change, deliberately out of #377's scope and not made here.

## Known drift left alone, deliberately

§2/§4/§5/§6 prose also carry stale figures — most sharply `apps/web/src/push/push-core.ts` branch coverage claimed at 88.88%, measuring 69.56%; both `event-store.ts` files; and a §6 sentence calling `convexity-caption.ts` "fully measured," which the fresh 95.65% contradicts. #377 is scoped to §7's table, so these are left as-is rather than patched incidentally. Saying so here so the next reader doesn't mistake this silence for currency — it's a follow-up, not a fix.

## Verification

No tests written, no source touched, no coverage raised. The issue is explicitly an accounting task, and the diff is one documentation file: `docs/coverage-rationale.md`.

## Addendum — review-fix pass (PR #387 review, pullrequestreview-4968807922)

The review re-measured against this branch's HEAD and found the completeness
claim above still false: two more files had no row anywhere in the document,
each hidden behind a filename collision with a different file this document
already accounts for — `packages/engine/src/reconciliations.ts` (collided
with the preferences package's row of the same bare name) and
`packages/event-store/src/gap-report.ts` (collided with `apps/web`'s
excluded shell and `push/gap-report-core.ts`). Both now have rows, labeled
to name their package so a third collision can't hide the same way. Two more
files (`cli-main.ts`, `ladder/fill-path-view.ts`) were named only in passing
elsewhere, with no figure — not an accounting by this document's own
contract — and now have real rows too. §7 holds 55 rows as of this pass, not
38.

The review also found §7's "Uncovered" column had an undocumented, silently
inconsistent derivation: some rows cited exactly what `pnpm coverage`'s text
reporter prints, others unioned line gaps with branch gaps without saying
so. The rule is now stated once in §7's preamble — cite exactly what the
text reporter prints — and every row was checked against it. `parse.ts` was
under-citing by two ranges; `dca-block.ts` and `fill-path.ts` were mixing
metrics, and `fill-path.ts` named the wrong line for its `bookAxis` gap
(`:443` executes 198 times and carries no branch; `:444` is the real one).
Three rows had a cosmetic bridging mismatch (`record-fill.ts`,
`import-orders.ts`, and this PR's own `reconciliations.ts` row above, which
was also re-derived from scratch rather than trusted — see the corrected
bullet above, which is what it should have said the first time).
`calendar.ts`'s row named one caller where `utcWeekday` has two.

Total diff after this pass: `docs/coverage-rationale.md` (+84/-36 against
`main`). Still no tests, no source, no coverage raised.

Closes #377.


AmetAlvirde (commented):
## Review — re-measured, not read

Ran `pnpm coverage` twice at this branch's HEAD (`e1d6265`) and enumerated every measured file below 100% from `coverage-final.json` rather than trusting the truncated text column. The two runs agree everywhere except one branch percentage on `packages/engine/src/parse.ts` (96.21 vs 96.18), so the numbers below are stable.

**The good news first, because it is most of the diff.** All 51 §7 rows' percentages reproduce **exactly** — every one. 44 of 51 rows' uncovered-range citations reproduce exactly too. All four spot-checked "covered now" claims hold against source. The three 0% rows' arguments check out against `vitest.config.ts`. The diff is documentation only.

What follows is what did not survive.

---

### 1. §7's completeness claim is still false — two measured files have no row anywhere (highest severity)

This is the deliverable, so it is the finding that matters. Enumerating from `coverage-final.json`: **66 files** measure below 100% on some metric. 51 have a §7 row; 13 are individually named by §§1/2/5/6. Two are named **nowhere in the document**:

| File | Measured | Uncovered |
| --- | --- | --- |
| `packages/engine/src/reconciliations.ts` | 98.02% lines, 98.46% branch | `:357-359` |
| `packages/event-store/src/gap-report.ts` | 100% lines, 98.14% branch | `:398` |

Verification that they are absent, not just missed by my eye:
- `grep -n "reconciliations" docs/coverage-rationale.md` returns **two** hits — the `plans-cli.ts` row and the *preferences* package's `reconciliations.ts` row. The engine module is never mentioned.
- `grep -n "gap-report" docs/coverage-rationale.md` returns **two** hits — `apps/web/src/push/gap-report.ts` (§1) and `push/gap-report-core.ts` (§7). `packages/event-store/src/gap-report.ts` is a different file and is never mentioned.

Both belong in §7's tables (`packages/engine/src` and `packages/event-store/src … ` respectively). Until they are there, "every other currently measured file below 100% that the earlier sections do not individually name" remains a false sentence — which is exactly what #377 exists to fix.

### 2. §7's preamble was not updated for the ten new rows

Lines 521 and 526 (outside the diff hunks, hence here rather than inline):

> Between those sections and a fresh `pnpm coverage` run, **38 measured files** sat with no row anywhere in this document
> … Not all **38 rows** are lines gaps — about a third are branch-only

The table now holds **51 rows**. The section whose completeness is the whole point of this PR miscounts its own contents by 13.

Line 531 has a related problem:

> All of the following are the same shape: a rejection/error-path branch that runs only on a malformed input, a network/provider failure, or a lock/fs race

Three of the rows the PR adds (`plans-cli.ts`, `operator-notice-cli.ts`, `lib/binance-spot.ts`) say *in the row itself* that they do not fit that taxonomy. The blanket sentence above them now contradicts them. It needs a clause, not a rewrite.

### 3. Two files below 100% are "named" by §§1/6 only in passing — judgment call, flagging not asserting

| File | Measured | Uncovered | Where it is "named" |
| --- | --- | --- | --- |
| `apps/price-feed/src/cli-main.ts` | 100% lines, 90.32% branch | `:98-100,111` | §1 line 36, only to say it is **not** excluded |
| `apps/web/src/ladder/fill-path-view.ts` | 98.52% lines, 94.24% branch | `:502,613,675-676` | §6 line 483, "the measured `ladder/fill-path-view.ts`" |

A literal reading of "the earlier sections do not individually name" lets both through — the filenames do appear. But neither mention carries a figure or accounts for a gap, and the clause cuts both ways: if a passing mention discharges the contract, `grep`-ing a module name comes back non-empty while telling the reader nothing. This is a judgment call about where the line sits, not a wrong number. Worth a decision either way; I'd add both rows.

### 4. PR-body arithmetic (not the document)

> `reconciliations.ts` — 86.66%, **13 ranges** — one more than the issue's 12. The extra is a new `:524-526`

The row cites **12** ranges and its prose correctly says "Eleven of the twelve". And `:524-526` is not new: it is #377's `:521-523` with 3 added, like every other range in that row. The document is right here; the PR body is not. Mentioned because it is the same shift that produces the inline finding on that row.

---

### On the review's own questions

- **A — are the cited ranges right?** Mostly yes, and better than the last four vintages. Every percentage in all 51 rows reproduces. Seven rows' range citations do not: `parse.ts` (a real omission), `dca-block.ts` and `fill-path.ts` (a different derivation, plus one wrong number), and `record-fill.ts` / `import-orders.ts` / `reconciliations.ts` (cosmetic). Inline on each.
- **B — is §7's completeness sentence true now?** **No** — see finding 1. Two files short, plus two borderline. The reverse check is clean: I cross-checked all 51 row names against the measured set and **no row exists for a file now at 100%** — no dead weight.
- **C — do the "covered now" claims hold?** **Yes, all of them.** Verified against source and the run: `import-orders.ts`'s `unreadable-sidecar-lines` (the refusal is at `:432`, absent from the uncovered set; `:427` is the `unreadable-sidecar` refusal the row now cites, correctly distinguished); `glance/verdict.ts`'s precedence sort (`fired.sort` at `:448`, absent — and `:621-622` really is `referenceLabel`'s invalid-calendar-date throw); `orders/ingest.ts`'s `timeInForce` (the block at `:491-498` is absent, and `:488-489` really is the `orderType` push — the flip is real); `provider.ts`'s `AbortError` (the branch at `:140-141` is absent; `:128` really is `options.fetchImpl ?? fetch`). Also spot-checked and confirmed: `fetch-prices.ts`'s `fetchImpl`/`sleepImpl` defaults covered with only `options.now` left at `:223`, and `push/glance.ts`'s three arms at `:93` / `:94` / `:108`. No false retirement found.
- **D — are the three 0% rows' arguments true?** **Yes.** `plans-cli.ts` is a bare top-level `try/catch` with top-level `await` (importing runs it), `operator-notice-cli.ts` is a module-level `writeOperatorNotice(...).then(ok, err)` (importing runs the write), and **neither appears anywhere in `vitest.config.ts`'s coverage `exclude`** — which does list `import-orders-cli.ts`, `record-fill-cli.ts`, `cancel-order-cli.ts`, `migrate-legacy-log.ts` and `apps/price-feed/src/cli.ts`. The `binance-spot.ts` claim checks out too: it is a `.ts` file holding a `useEffect`/`useState` hook with nothing executing at import, the include glob really is `["packages/*/src/**/*.ts", "apps/*/src/**/*.ts"]`, and `SummaryCard`/`SectionTable`/`FillPath`/`PriceDropPathChart` really are `.tsx` under `apps/web/src/components/`, so the extension boundary really is what separates them. One stylistic note: for `plans-cli.ts` the closest §1 sibling is `report.ts` / `spine.ts` (also top-level `try/catch` over a fold + an already-tested formatter), not the DI-wiring CLIs the row cites — the cited siblings are self-executing too, so the claim holds, but the nearer analogue would land harder.
- **E — scope?** **Clean.** One file, `docs/coverage-rationale.md`, +37/−24. No source, no tests, no coverage raised.

### On categorization and voice — judgment, not defect

All ten new rows genuinely categorize rather than restate line numbers: `fold-lines.ts` names its shape by pointing at `gap-lines.ts`'s twin; `unattended-report.ts`, `convexity-caption.ts`, `import-orders-rung-picks.ts` and `dca-block.ts` all say *real and reachable, no current fixture drives it*; `sidecar-io.ts` splits its two ranges into defensive and real; `plans.ts` and `reconciliations.ts` classify by the per-field-rejection shape and then correctly separate the one range that is not that shape (`:524-526`, defensive against its own serializer). The three "does not fit the taxonomy" rows earn it: two are an instrumentation-posture bug whose real home is §1, one is §6's blind spot leaking through an extension boundary — forcing any of them into defensive/unreached-fixture/real-branch would have been the dishonest move. Naming the `vitest.config.ts` fix and declining to make it is the right call for an accounting issue.

docs/coverage-rationale.md
```diff
@@ -537,64 +537,77 @@ individually itemized with its own paragraph. Grouped by package.
...
+| `push/gap-report-core.ts` | 100% lines, 97.61% branch | `:209` | One branch-only gap. |
+| `push/glance.ts` | 100% lines, 93.18% branch | `:93-94,108` | Branch-only gaps: `:93` is `data.closes ?? []`'s default-array arm (the fold always carries a `closes` array in every fixture), `:94` is the future-dated-close skip (`close.asOf > asOf`) never taken, and `:108` is `data.instruments ?? []`'s equivalent default-array arm. |
 | `auth/verify-rate-limit-core.ts` | 100% lines, 92% branch | `:117,171` | Branch-only gaps in the attack script's decision core. |
 
 **`packages/engine/src`**
 
 | File | Lines | Uncovered | What's there |
 | --- | --- | --- | --- |
 | `parse.ts` | 97.72% | `:263-264,403-404` | Two more per-field genesis-door rejection branches, same category as §5's `events/parse.ts` gap. |
```
AmetAlvirde: **Under-cites by two ranges — and this is a row the PR reports as audited.**

The percentage is right, which is why it looks fresh. The ranges are not.

- doc: `:263-264,403-404`
- `pnpm coverage` at `e1d6265`: `150-151,235-236,263-264,403-404`

The two missing ranges are the same category the row already describes:
- `:150-151` — the `!Array.isArray(value)` → `schemaError(path, \`Review file is missing array: ${path}\`)` rejection
- `:235-236` — `validateReserves`'s equivalent, `schemaError("reserves", "Review file is missing array: reserves")`

Measured by reading `getLineCoverage()` off `coverage/coverage-final.json`; both lines report 0 hits, in both of my runs. `packages/engine/src/parse.ts` is the one file whose branch percentage moved between my two runs (96.21 / 96.18) — but its statement percentage is identical in both, so the uncovered-line set is stable and this is not flake.

docs/coverage-rationale.md
```diff
@@ -537,64 +537,77 @@ individually itemized with its own paragraph. Grouped by package.
...
+| `push/dca-block.ts` | 98.41% | `:161,232,236-237,277-278,282-283` | `:161` is `torn ?? []`'s default-array arm (every fixture supplies a torn-acts array); `:232` is `position?.lots ?? []`'s equivalent; `:236-237` are the `currency`/`reviewFx` field-presence spread guards, one arm each unexercised. `:277-278,282-283` are the `placedQuantity`/`venueFilledFraction` omit-when-undefined spread guards — real, reachable: every current fixture supplies both fields, so the omission arm that would drop them from the view never fires. |
```
AmetAlvirde: **This row's Uncovered column does not reproduce from `pnpm coverage`.**

- doc: `:161,232,236-237,277-278,282-283`
- what the run prints: `278,283`

Every number here is real, so this is a derivation mismatch rather than an invention. Off `coverage-final.json`:

- **uncovered lines** (0 hits): `278`, `283` — these are the two `? {}` omission arms.
- **uncovered branches**: `161` (0/1), `232` (0/1), `236` (0/1), `237` (0/1), `277` (0/1), `282` (0/1) — but all six of those *lines* execute (14, 14, 14, 14, 42, 42 hits respectively).

So the row silently unions two metrics. Because `dca-block.ts` measures below 100% on lines, istanbul's text reporter takes the line path and prints only `278,283`; the branch lines never appear. §7's own preamble promises "Line numbers are the fresh run's", and the other 49 rows honour that literally — so a reader re-running `pnpm coverage` and diffing gets six false positives on this row alone.

The prose is accurate throughout: I confirmed `:161` is `torn ?? []`, `:232` is `position?.lots ?? []`, `:236`/`:237` are the `currency`/`reviewFx` spread guards, and `:277-278`/`:282-283` are the `placedQuantity`/`venueFilledFraction` omit-when-undefined guards with the omission arm unfired. Only the column is off. Either cite `278,283` and describe the branch lines in prose, or state the union convention in the preamble — but it should not differ silently from the neighbouring rows.

docs/coverage-rationale.md
```diff
@@ -537,64 +537,77 @@ individually itemized with its own paragraph. Grouped by package.
...
+| `fill-path.ts` | 97.75% | `:175-176,258,261-266,270,443-444` | Added at `d5fe02c` — no row existed before #377. `:175-176` is a rendering branch in the "filled" label (the `book === "not-recorded"` false arm, "filled at venue — partly recorded"). `:258,261,264,270` are branch-only gaps and `:262-263,265-266` are full lines, all in the same retired-placement join: `:262-263` is the declared-join-invalid fallback to `undefined` and the no-declaration price-match fallback — both real, reachable, but no current fixture has a retired/cancelled placement that fails its declared join or carries no declaration at all; `:265-266` is the resulting skip (`rungId === undefined || joined.has(rungId)`) that never fires because those fallbacks never run; `:270` is `declaresSomething ? "declared" : "price-matched"`'s `"price-matched"` arm, unreachable for the same reason. `:443-444` is one arm of the `bookAxis` three-way classification. All real, reachable branches — none defensive, none low-value. |
```
AmetAlvirde: **Same union as `dca-block.ts` above, plus one number that is wrong on every metric.**

- doc: `:175-176,258,261-266,270,443-444`
- what the run prints: `176,262-263,265-266`

**`443` is the wrong number.** Line 443 is `const bookAxis: FillBookAxis =` — it executes **198 times** and carries no branch at all. The three-way `bookAxis` classification the row describes lives on line 444 (`booked <= 0 ? "not-recorded" : booked >= consumed ? "recorded" : "partly-recorded";`), whose branch measures **2/3**. So `:443-444` should be `:444`.

For the rest, the union again: uncovered lines are `{176, 262, 263, 265, 266}`; uncovered branch lines are `{175, 258, 261, 262, 264, 270, 444}`. The row's citation is those two sets merged into ranges — which is why `258`, `261`, `264`, `270` appear even though all four lines execute (3, 2, 3, 3 hits).

Two smaller things in the same sentence. The row says "`:258,261,264,270` are branch-only gaps and `:262-263,265-266` are full lines" — but `262` is **both** (0 line hits *and* a 0/1 branch), and `444`'s branch gap is left out of the branch-only list entirely. And the row's closing "All real, reachable branches — none defensive, none low-value" is a fair taxonomy call, but it is doing more work than the rest of the row: worth one clause on why the retired-placement join is reachable in production rather than only in principle.

Credit where due: I checked the `:175-176` claim closely because it looked inverted, and it is correct — line 176 is `: "filled at venue — partly recorded"`, genuinely the false arm of `book === "not-recorded"`, exactly as written.

docs/coverage-rationale.md
```diff
@@ -537,64 +537,77 @@ individually itemized with its own paragraph. Grouped by package.
...
+| `push/gap-report-core.ts` | 100% lines, 97.61% branch | `:209` | One branch-only gap. |
+| `push/glance.ts` | 100% lines, 93.18% branch | `:93-94,108` | Branch-only gaps: `:93` is `data.closes ?? []`'s default-array arm (the fold always carries a `closes` array in every fixture), `:94` is the future-dated-close skip (`close.asOf > asOf`) never taken, and `:108` is `data.instruments ?? []`'s equivalent default-array arm. |
 | `auth/verify-rate-limit-core.ts` | 100% lines, 92% branch | `:117,171` | Branch-only gaps in the attack script's decision core. |
 
 **`packages/engine/src`**
 
 | File | Lines | Uncovered | What's there |
 | --- | --- | --- | --- |
 | `parse.ts` | 97.72% | `:263-264,403-404` | Two more per-field genesis-door rejection branches, same category as §5's `events/parse.ts` gap. |
+| `calendar.ts` | 95.12% | `:80-81` | Added at `d5fe02c` — no row existed before #377. `utcWeekday`'s invalid-calendar-date throw — defensive, its only caller (`weekdayName`) passes an already-validated `asOf`. |
```
AmetAlvirde: **"its only caller" is false — there are two.**

> `utcWeekday`'s invalid-calendar-date throw — defensive, **its only caller** (`weekdayName`) passes an already-validated `asOf`.

`utcWeekday` is called from two places in this file:
- `calendar.ts:51` — `const weekday = utcWeekday(asOf, "isWeekend");`
- `calendar.ts:74` — `return WEEKDAY_NAMES[utcWeekday(asOf, "weekdayName")] as string;`

The `caller` parameter the function takes is itself the tell that it has more than one. The defensive classification survives — both call sites pass a validated `asOf` — but the argument as written rests on a uniqueness that is not there, and "its callers" would carry it just as well. The `95.12%` and `:80-81` both reproduce exactly.

docs/coverage-rationale.md
```diff
@@ -537,64 +537,77 @@ individually itemized with its own paragraph. Grouped by package.
...
+| `reconciliations.ts` (preferences package) | 86.66% lines, 85.05% branch | `:186-190,200-201,203-207,215-218,222-224,226-227,239-241,243,277-278,329-330,350-355,524-526` | Added at `d5fe02c` — no row existed before #377. Eleven of the twelve ranges are the same per-field malformed-value rejection shape, spread across `readDeclared` (`:186-190` the `ended` variant's `effectiveAt` guard, `:200-201` the `kind` guard, `:203-207` the in-force `effectiveAt` guard, `:215-218` the `tierOrder` guard, `:222-224,226-227` the `planId` guard, `:239-241,243` the unknown-`status` default) and `readReconciliationLine` (`:277-278` the whole-line-not-an-object guard, `:329-330` the `fillKind` guard, `:350-355` the `mismatches`-must-be-an-array guard). The twelfth, `:524-526`, is different in kind: `appendReconciliation`'s round-trip self-check — a `JSON.parse` catch on the string this module's own serializer just produced, reachable only if the serializer itself emits invalid JSON — defensive against its own bug, not malformed input. |
```
AmetAlvirde: **Two of the twelve ranges do not reproduce — and the pattern suggests the row was shifted rather than measured.**

- doc: `…,215-218,**222-224,226-227**,**239-241,243**,277-278,…`
- run: `…,215-218,**222-227**,**239-243**,277-278,…`

Cosmetic in substance: lines 225 and 242 carry **no statement**, so istanbul's range builder bridges straight across them, while the doc's version breaks there. The set of uncovered statements is identical, and the doc's split is arguably the more precise description. Flagging it for two reasons.

First, it means a reader following the preamble's "re-run `pnpm coverage` for the current picture" gets a diff on this row.

Second, and more usefully: this row's twelve ranges are **byte-identical to #377's table at `7269a9f` with 3 added to every number**. `183-187`→`186-190`, `197-198`→`200-201`, … `521-523`→`524-526`, including the two break points that today's run does not produce. The PR body says these were "measured fresh at `d5fe02c`, not copied from the issue" — for the sibling rows that is demonstrably true (`sidecar-io.ts` went `:123-124,171-173,175-176` → `:152-153,200-205`, nothing like a shift), but this row is consistent with a `+3` on the issue's list. The percentages and the range *set* are right either way, so nothing here is wrong — it is worth one re-derivation to be sure.

Same note applies to the PR body's "13 ranges — one more than the issue's 12. The extra is a new `:524-526`": the row has 12, its prose says twelve, and `:524-526` is `:521-523` shifted.

docs/coverage-rationale.md
```diff
@@ -537,64 +537,77 @@ individually itemized with its own paragraph. Grouped by package.
...
...iew-abandoned` refusal and the write-door guard, both now exercised — see below). `:331-337` is `loadInForceLadders`'s `catch` when `io.loadPlans` itself throws (an unreadable `plans.jsonl`); `:339-340,342-344` is its `load-failed` status branch (a malformed `plans.jsonl`); `:348-349` is its skipped-plan-line `io.err` report, entered only when the plans file carries a skip. `:369-370` is `declaredJoinsOnFile` picking up a record that already carries an explicit `planId`+`rungId` (a re-import of an already-declared join). `:394-396` is the `unreadable-export` refusal when the CSV itself cannot be read, `:410-411` is the `no-orders` refusal for a header-valid CSV with zero resting orders, and `:427-428` is the `unreadable-sidecar` refusal — the whole orders book could not be read at all (distinct from `unreadable-sidecar-lines`, a partially-read book, which is covered now). `:585-588` is the restatement-construction guard the file's own comment calls "UNREACHABLE, AND HANDLED ANYWAY" — defensive. The `interview-abandoned` refusal (#386) and the write-door guard ahead of every append are both covered; the funding-declaration report-join `io.err` this row used to cite is covered too. |
```
AmetAlvirde: Minor, same class as the `reconciliations.ts` row below: `:339-340,342-344` does not reproduce — the run prints `339-344`. Line 341 is the continuation string `` `any order imported now joins its rung by price match` `` and carries no statement, so istanbul bridges it.

No substantive error — the uncovered statement set is the same, and I verified the prose: `:331-337` really is `loadInForceLadders`'s `catch` when `io.loadPlans` throws, `:339-344` really is the `load-failed` arm, `:348-349` really is the skipped-plan-line `io.err`, `:369-370` really is `declaredJoinsOnFile`'s explicit-`planId`+`rungId` push, and `:585-588` really does sit under the comment reading "UNREACHABLE, AND HANDLED ANYWAY" at line 578. The `unreadable-sidecar` / `unreadable-sidecar-lines` distinction is right too (`:427` vs `:432`, the latter covered).

docs/coverage-rationale.md
```diff
@@ -537,64 +537,77 @@ individually itemized with its own paragraph. Grouped by package.
...
-| `ingest-commit.ts` | 96.72% | `:101-102,129-130` | Two `git` exec-failure branches in the commit-capture path. |
+| `record-fill.ts` | 85.19% | `:303-304,544-545,579-580,602-607,615-616,743-752,761-762,765-766,787-792,799-804,817-818,820-821,846-851,858-859,873-878,883-889` | Not the "worst gap in the repo" on either measure (audit finding 3's headline overstated it): by line COUNT `events/parse.ts`'s 78 uncovered lines are more; by PERCENTAGE `apps/web/src/lib/dashboard.ts`'s 78.26% is lower. What is real: 65 lines across 16 ranges, the largest share of any single file in this table. `:743-752` is the interactive "Also record N confirmed cancellation(s) in this act?" prompt arm — an operator-facing branch, not a rejection path. The remaining ranges are earlier refusal/assembly branches (`reject(io, …)` calls and the fill act's own re-parse/re-cross-reference of the transaction it just built) plus the `ordersToAppend` cancellation-sibling assembly; `record-fill.test.ts`/`record-fill-reliable.test.ts` drive the refusals earlier in the flow but not this full set of late-stage branches. |
```
AmetAlvirde: Minor, same class: `:761-762,765-766` does not reproduce — the run prints `761-766`. Lines 763 and 764 carry no statement and istanbul bridges them.

Worth noting this row's "65 lines across 16 ranges" is consistent with its own citation and is in fact **more** accurate than the tool: the true uncovered-statement count is 65, whereas the run's bridged rendering implies 67 across 15 ranges. So if these three rows are deliberate, say so once in the preamble — the convention is defensible, it just is not stated, and right now it reads as drift.

Everything else here reproduces: `85.19%` exactly, and `:743-752` really is the `Also record ${cancelled.length} confirmed cancellation(s) in this act? [y/N]:` prompt arm.

AmetAlvirde:
## Response to review (pullrequestreview-4968807922) — commit 29357ba

Re-derived every number in this reply from `coverage-final.json` at this branch's HEAD via `istanbul-lib-coverage`'s own `toSummary()` / `getLineCoverage()` / `getBranchCoverageByLine()`, not by hand — including the exact range-bridging logic `istanbul-reports`' text reporter uses (`nodeMissing` in `istanbul-reports/lib/text/index.js`), so what §7 now cites is provably what `pnpm coverage` prints, not an approximation of it.

**1 (HIGH) — completeness.** Both files now have rows: `packages/engine/src/reconciliations.ts` (`:357-359`, 98.02% lines / 98.46% branch) and `packages/event-store/src/gap-report.ts` (`:398`, 100% lines / 98.14% branch). Both labels now say which package, so the same filename collision can't hide a third file behind either row again. Re-enumerated every file below 100% from the JSON against the finished document: **66 files measure below 100%; 55 have a §7 row; 11 are named with an actual figure in §§1/2/4/5/6.** Zero unnamed. §7's preamble states 55, not 38.

**2 (MEDIUM-HIGH) — `parse.ts`.** Row now cites all four ranges (`:150-151,235-236,263-264,403-404`), each attributed to its guard.

**3 & 4 (MEDIUM) — `dca-block.ts` / `fill-path.ts`.** Root cause fixed, not just the two rows: §7's preamble now states the derivation rule once — the Uncovered column cites exactly what the text reporter's own range-builder prints for that file (line gaps when lines < 100%, branch gaps when lines = 100% but branches aren't; never both). `dca-block.ts` now cites `:278,283`; `fill-path.ts` now cites `:176,262-263,265-266` and the row explicitly says `:443` executes 198 times and carries no branch, `:444` is the real 2/3 `bookAxis` gap. Branch-only companions that used to be silently unioned into these columns are now named in prose instead, not the range list.

**5 (MEDIUM) — stale preamble.** Rewritten: states 55 rows (not 38), explains where each of the four new rows came from, and the taxonomy sentence now carries an explicit carve-out for the three rows (`plans-cli.ts`, `operator-notice-cli.ts`, `lib/binance-spot.ts`) that say plainly they don't fit it.

**6 (LOW) — cosmetic bridging + the preferences `reconciliations.ts` row.** `record-fill.ts` (`:761-766`, was split at `:761-762,765-766`) and `import-orders.ts` (`:339-344`, was split at `:339-340,342-344`) now bridge the same way the tool does. The preferences `reconciliations.ts` row was re-derived from scratch against the fresh run rather than trusted — it was in fact the stale `+3`-shifted table, exactly as flagged. Corrected numbers: **86.66% lines, 85.05% branch, 10 ranges** (two pairs merge under the bridging rule: `:222-224,226-227`→`:222-227`, `:239-241,243`→`:239-243`), nine of the ten the per-field-rejection shape, the tenth (`:524-526`) the round-trip self-check.

**7 (LOW) — `calendar.ts`.** Row now names both callers (`isWeekend` at `:51`, `weekdayName` at `:74`); defensive categorization holds for both, as you found.

**8 (LOW/judgment) — `cli-main.ts` / `fill-path-view.ts`.** Agreed with your read: a passing mention with no figure isn't an accounting. Gave both real §7 rows rather than tightening the earlier mentions — `cli-main.ts` (100% lines, 90.32% branch, `:98-100,111`, all DI-default arms) and `fill-path-view.ts` (98.52% lines, 94.24% branch, `:502,613,675-676`).

**9 — PR body.** Corrected: `reconciliations.ts` has 12 ranges in the original commit, not 13, and `:524-526` is the issue's `:521-523` shifted by the same +3 as every other range, not new. Added an addendum documenting this review-fix pass so the body's "measured fresh" claims stay true after the corrections above (final `reconciliations.ts` figures now 86.66%/85.05%/10 ranges, per finding 6).

All fixes are in `docs/coverage-rationale.md`, commit 29357ba, pushed to this branch. No source or tests touched.

PR #391, opened 2026-08-19
test: state the test budget once, and stop two CLI shells reporting a dishonest 0%

AmetAlvirde:
Three small seams, one commit each. No issue — the edits are minimal and the
rationale lives in the commit messages and in the code comments themselves.

## 1. The test budget, stated once

Vitest's default `testTimeout` is **5000 ms** and this repo does not fit in it.
The tui and price-feed suites spawn real subprocesses (`tsx`, the price-feed
wrapper shell) and build real git repos in temp dirs. Measured: a case whose
body costs ~75 ms alone costs **1.4–2.0 s** under full-suite parallel load — a
19–27× inflation leaving roughly **2.4×** of headroom.

**30_000 is not a new number.** The repo had already converged on it eleven
times over:

- ten test files each carrying their own `vi.setConfig({ testTimeout: 30_000 })`
- CI passing `--testTimeout=30000` on the command line

The defect was *where the eleventh lived*. The flag was CI-only, so **a local
run was stricter than CI** — a loaded machine could go red locally on a case CI
would never fail. A red CI cannot reproduce is a red that gets chased.

So the budget moves into `vitest.config.ts` and comes off CI's command line.
One statement, every run reads it, and it holds as the suite grows and on
slower hardware.

The config comment carries the whole trade, including the two things it is
easy to get wrong later:

- **What it costs** — a genuinely hung test now takes 30 s to fail instead of
  5 s. That is the trade CI has been making all along.
- **What it cannot do** — `testTimeout` is a timer on the same thread and
  cannot preempt a blocking `spawnSync`. That is why the spawning tests pass
  their own `timeout` to the spawn.

## 2. The two CLI shells reporting a dishonest 0%

`apps/tui/src/plans-cli.ts` and `apps/price-feed/src/operator-notice-cli.ts`
are self-executing shells of exactly the shape §1 of the coverage rationale
already excludes — importing either one *runs the act*. Their four siblings
were excluded; these two were missed, so they reported 0%, which reads as
untested code when the truth is unmeasurable-by-design code whose every
decision lives in a measured module.

§7 already carried a row for each saying precisely this and naming the exclude
as the fix. With the exclude in place those rows are wrong twice over — the
files are no longer measured, so they are not "below 100%", and §7's contract
is that it names every *measured* file that is. Both rows move to §1 with what
guards each instead.

§7's preamble said three rows do not fit its taxonomy. `lib/binance-spot.ts`
is now the only one, and the preamble now says why an exclude is the **wrong**
fix for it: it is a browser hook caught by the `*.ts`-only include glob where
§6's excused components are `.tsx`, so it needs the RTL toolchain §6
anticipates or a logic extraction.

## 3. The ten per-file declarations the config replaced

With the budget in the config, every `vi.setConfig({ testTimeout: 30_000 })` is
a no-op. Leaving them would be worse than redundant on two counts:

- **They mislead.** Ten files declaring 30 s implies the surrounding default is
  still 5 s, which stopped being true one commit earlier.
- **They pin.** A no-op matching the global today becomes a divergence the
  moment the global moves — raise the config to 45 s and these ten silently
  stay at 30 s, re-creating inside the suite the exact split-brain commit 1
  removed from CI. That is the real argument: not that they are dead, but that
  they are dead in a way that comes back to life pointing the wrong direction.

The justification comments went with them — a comment explaining an action the
file no longer takes is not documentation. One piece was worth keeping, and it
was never really about its file: `event-store.test.ts`'s ruling that the number
is **headroom over scheduler contention, not a tolerance for slow code**, so a
case that actually reaches 30 s is a real hang to fix and not a budget to raise
again — with #178 as the precedent. That moved into the config comment, where
it governs every suite instead of one.

`vi` became unused in six files and was dropped from each; four still use it
for `vi.fn` / spies. The surviving `vi.setConfig` mention is
`migrate-legacy-log.test.ts`'s comment explaining it cannot preempt a blocking
`spawnSync` — still true, and now true of the config's value.

## Verification

- `pnpm typecheck` — clean.
- `pnpm test` — **162 files passed / 3 skipped, 2555 passed / 51 skipped**,
  30.6 s. Matches the pre-change baseline exactly. (The 3 skipped files are the
  Postgres suites gated on `NUMISMA_TEST_DATABASE_URL`.)
- **The global budget proven to reach the files that lost their `setConfig`**,
  rather than inferred from a green suite (these cases finish in ~2 s either
  way, so passing proves nothing on its own): a throwaway 6.5 s case in
  `apps/tui/src/` with no per-file `setConfig` **passes**, where the 5000 ms
  default would have failed it.
- §7's completeness contract checked **by enumeration, not assertion** —
  `coverage-final.json` walked directly: **66 below-100% files → 64**, exactly
  the two whose rows were removed; all 64 still named in the document; **0
  named nowhere**; neither shell appears in the report at all.

## What this does NOT do

- **Issue #372 stays open, deliberately.** The margin on that case widens from
  ~2.4× to ~14.7×, which makes the flake much less likely, but the underlying
  cost is untouched: `makeCaseDir` still builds a git repo the two refusal
  cases never exercise. That issue is not resolved here and must not be
  auto-closed by this PR.
- **It changes no assertion.** Nothing in any test's behavior moved.


docs/coverage-rationale.md
```diff
@@ -34,6 +34,8 @@ dishonest.
...
+| `apps/tui/src/plans-cli.ts` | The `pnpm plans` entry: a bare top-level `try/catch` that resolves the fold (`loadFoldedReview`), the `plans.jsonl` sidecar and the `reconciliations.jsonl` trail and hands all three to `formatPlansReport`, then sets `process.exitCode`. Importing it *runs the act*, same shape as the orders CLI shells above. It also inherits the event log's write-on-read quarantine maintenance, which its own header names — so importing it is not even read-only. Excluded as of the increment that added it to `vitest.config.ts`; before that it reported a dishonest 0% and §7 carried a row saying so. | `formatPlansReport` IS measured (`plans-report.test.ts`), as is the engine-side `listPlansAsOf` it renders. The three reads it wires are each measured in their own modules (`loadFoldedReview` by `packages/event-store/src/event-store.test.ts`, the sidecar and trail loaders by their own suites). The shell's own argv/exit-code wiring has no spawn test today — the same open gap the paragraph below this table names for `import-orders-cli.ts`. |
```
AmetAlvirde: **MEDIUM — this row forward-references a paragraph that does not name it, and that paragraph's count is now an undercount.**

This row ends with:

> The shell's own argv/exit-code wiring has no spawn test today — the same open gap the paragraph below this table names for `import-orders-cli.ts`.

The paragraph below the table (`:56-64`) still reads:

> The **four** orders/migration CLI shells above … The **other three** CLI shells (`import-orders-cli.ts`, `cancel-order-cli.ts`, `migrate-legacy-log.ts`) have no such spawn test today — they remain untested-by-any-suite, not merely uninstrumented — which is a real, open gap, not one this table should imply is closed by analogy.

So the pointer is dangling: the paragraph enumerates exactly three files by name and neither new arrival is one of them.

I checked whether either new shell actually has a driver, and neither does:

- `plans-cli.ts` — no suite imports or spawns it. The only reference anywhere in `apps/`/`packages/`/`ops/` is a prose mention in `apps/tui/src/plans-report.ts:3`.
- `operator-notice-cli.ts` — the wrapper harness looks like a driver and is not one. `apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts` exercises the wrapper's step 5b through a **fake** bin (`fake-bin.testkit.ts`'s `"writes-operator-notice"` behavior, armed via `operator-notice.testkit.ts:195`). The real CLI is never spawned — which is the point of that harness, but it means the shell itself is untested.

**Concrete failure scenario.** Before this PR both files sat in the report at 0%, and that 0% was itself the flag — ugly, but load-bearing. After this PR they are out of the number entirely, and the only prose that enumerates "excluded *and* untested-by-any-suite" says there are three of them and lists three other filenames. The next person deciding where to spend a slice greps §1, reads "the other three," and concludes `plans-cli.ts` and `operator-notice-cli.ts` are in the guarded majority. The exclude is supposed to trade a dishonest 0% for an honest accounting; here it trades it for silence, against a document whose opening line promises "no gap is silent."

The fix is small — "the other **five** CLI shells," with the two new names added to the list — but it should land with the exclude, not after it, because the 0% that used to do this job is gone as of this commit.

vitest.config.ts
```diff
@@ -44,6 +44,43 @@ export default defineConfig({
...
+    // WHAT IT COSTS: a genuinely hung test now takes 30 s to fail instead of 5 s.
+    // That is the trade CI has been making all along and it is the right one — a
+    // hang is a bug that will be found regardless, a false red is one that burns a
+    // session.
+    //
+    // WHAT IT DOES NOT DO: it cannot preempt a BLOCKING synchronous body.
+    // `testTimeout` is a timer on the same thread, so a `spawnSync` that never
+    // returns runs past any value set here. That is why the spawning tests pass
+    // their own `timeout` to the spawn — see `apps/tui/src/migrate-legacy-log.test.ts`.
+    testTimeout: 30_000,
```
AmetAlvirde: **INFO — the budget is stated once for test bodies; hooks are still on Vitest's 10 s default, and the comment reads as though they are not.**

The comment says this "applies it to every suite" and its **WHAT IT DOES NOT DO** paragraph names exactly one limit (a blocking `spawnSync`). There is a second one worth a clause: `testTimeout` does not govern hooks. `hookTimeout` and `teardownTimeout` are separate settings, both still at the 10 s default — confirmed in the installed copy, `node_modules/vitest/dist/chunks/coverage.DfSpMS-b.js:3922`:

```js
resolved.hookTimeout ??= resolved.browser.enabled ? 3e4 : 1e4;
```

**Not a regression** — `vi.setConfig({ testTimeout: 30_000 })` did not cover hooks either, and neither did CI's `--testTimeout=30000`, so this branch is exactly as exposed as `main`. It is the *claim* that got broader than the setting.

**Why it is worth a clause rather than nothing.** Every one of the ten files that just lost its declaration cleans up in an `afterEach` that `rm`s its temp dirs recursively, and for the git-heavy suites (`ingest-commit`, `ingest-commit-hardening`, `go-back-invariant`, `ingest-capture-wiring`) those are real `.git` trees full of small objects. Apply this comment's own 19–27× parallel-load inflation figure to that cleanup and the ceiling it runs into is 10 s, not 30 s. When it happens the failure prints `Hook timed out in 10000ms`, points at no test, and raising the number this comment tells you not to raise would not have helped — which is precisely the "a red that gets chased" cost the rest of the comment is written to prevent.

Suggest one sentence in **WHAT IT DOES NOT DO**: it governs test bodies only; `hookTimeout` / `teardownTimeout` remain at 10 s, and a cleanup hook that crosses that fails independently of this budget.

docs/coverage-rationale.md
```diff
@@ -558,11 +560,17 @@ branch that runs only on a malformed input, a network/provider failure, or a
...
 just not yet each individually itemized with its own paragraph. Grouped by
-package. Three rows below say plainly that they do NOT fit even that:
-`plans-cli.ts`, `operator-notice-cli.ts` and `lib/binance-spot.ts` are an
-instrumentation-posture problem — a genuine 0% that `vitest.config.ts`
-should exclude and does not — not a branch shape, and are named as such in
-their own rows rather than folded into this sentence.
+package. One row below says plainly that it does NOT fit even that:
+`lib/binance-spot.ts` is an instrumentation-posture problem — a genuine 0%
+— not a branch shape, and is named as such in its own row rather than folded
+into this sentence. It is the LAST of what used to be three. `plans-cli.ts`
```
AmetAlvirde: **LOW — while rewriting this preamble, the row count a few lines above it (`:537-538`, **"55 rows, as of this fix."**) went stale.**

This edit correctly narrows "three rows below" to one. But the sibling claim in the same preamble was not adjusted: §7 now has **53** rows, not 55.

Counted mechanically — table rows (`^| ` + backtick) between the `## 7.` heading and the `None of the above is claimed unreachable` paragraph:

- `main`: 55
- this branch: 53

Exactly the two rows this commit deletes, so the section is internally consistent; only the stated total is not. Line 537 is untouched by the diff, which is how it slipped past.

Worth fixing because §7's contract is that it is checkable by enumeration — the PR body verifies the file set that way, and a reader doing the same check on the row count gets a mismatch with no way to tell a deliberate relocation from a silent drop. The lineage sentence immediately above ("opened at 38 rows … ten more at `d5fe02c` … this fix adds two … plus two") is the natural place to record the subtraction.

AmetAlvirde (commented):
## Review — 4 findings, none blocking the code change

The three seams themselves verify clean. Everything below is the doc/ledger half of the increment: this PR moves two files *out of the number*, and three separate places that enumerate "what is excluded" or "what is untested" were not moved with them. Ranked most-severe first.

---

### 1. MEDIUM — §1's untested-by-any-suite ledger still says "three", and the new row points at it as if it said five

`docs/coverage-rationale.md:56-64` (the **"'Excluded from instrumentation' is not 'untestable.'"** paragraph) and the new row at `docs/coverage-rationale.md:37`.

Detail in the inline comment on `:37`.

### 2. LOW — §7's row count is stale: the preamble says 55, the section now has 53

`docs/coverage-rationale.md:537-538` — **"55 rows, as of this fix."**

Counted mechanically (`^| ` rows between the `## 7.` heading and `None of the above is claimed unreachable`): `main` = **55**, this branch = **53** — exactly the two rows this PR deletes. Line 537 is untouched by the diff, which is why it slipped.

This matters more than an off-by-two normally would because §7's entire value proposition is that its count is checkable by enumeration — the PR body itself verifies the file set "by enumeration, not assertion." A reader who counts finds 53 against a stated 55 and cannot tell whether two rows were deliberately relocated or silently dropped. The lineage sentence right above it ("opened at 38 rows … ten more at `d5fe02c` … this fix adds two … plus two") is the natural place to append the subtraction.

### 3. LOW — two package READMEs enumerate the exclusion set and were not updated

- `apps/price-feed/README.md:161` — "`src/cli.ts` is excluded from the coverage number (…see `vitest.config.ts`…)". As of this branch two files in that package are excluded; `operator-notice-cli.ts` appears nowhere in that README.
- `apps/tui/README.md:80-102` and `:141-145` — the two "**Excluded from the coverage number**" blocks list the four orders/migration shells and then `report.ts` / `spine.ts` / `spine-reset.ts`. `plans-cli.ts` appears nowhere in the file.

Both statements were **true on `main`** and are false on this branch, so this is staleness the PR introduces rather than staleness it inherits. Both READMEs point at `docs/coverage-rationale.md` as the authority, which softens it — but each also makes its own definite claim about which files in *its* package are out of the number.

### 4. INFO — the config comment over-claims: the budget covers test bodies, not hooks

`vitest.config.ts:46-83`. Detail in the inline comment. Not a regression — `vi.setConfig({ testTimeout })` and `--testTimeout` never covered hooks either — just one clause missing from the "WHAT IT DOES NOT DO" paragraph.

---

## Verified clean (checked, nothing found)

The four things most likely to make this PR wrong all hold:

- **The global budget genuinely reaches every file that lost a declaration.** There is exactly one vitest config in the tree (`vitest.config.ts`); no `projects` / `workspace` key; `apps/web/vite.config.ts` declares no `test` block; and no workspace `package.json` has a `test` script of its own — all testing runs through root `vitest run`. So the ten files, including the price-feed suites that never had a per-file declaration, all read the config value.
- **Nothing in CI bypasses the root config.** `.github/workflows/ci.yml` is the only workflow file and contains exactly one vitest invocation (`:104`, `pnpm test`). Nothing under `ops/` or in a git hook invokes vitest. `test:wrapper` also runs from the root.
- **No deleted declaration was doing more than it looked.** All ten set `testTimeout` only — none set `hookTimeout`, `retry`, or anything else — so the config is a faithful replacement. Confirmed no `vi.setConfig` call survives anywhere in `apps/`, `packages/`, `ops/` (the one remaining hit is prose inside a comment at `migrate-legacy-log.test.ts:173`, which is still accurate). Per-test inline timeouts in the gated integration suites (`60_000` / `180_000`) still override the global, as intended.
- **All six dropped `vi` imports are genuinely unused.** Zero `vi.` call sites in any of the six; the four retained imports each use `vi.spyOn` / `vi.restoreAllMocks`. `pnpm typecheck` — clean.
- **The two exclude globs match exactly the two named files and nothing else.** Fresh `pnpm coverage` on this branch: 117 files measured, **64** below 100% (down from 66), neither `plans-cli.ts` nor `operator-notice-cli.ts` present in `coverage-final.json` at all, and every one of the 64 is named somewhere in the document. §7's *completeness* claim is accurate — it is only the *count* (finding 2) that is stale.
- Full run reproduced green: `pnpm coverage` exit 0, 162 files.

docs/coverage-rationale.md
AmetAlvirde: Fixed in f1ef070. §1's paragraph now opens on six CLI shells (the four orders/migration entries plus these two) and names the five without a spawn test. The row's dangling forward reference is gone — it now states directly that nothing imports or spawns `plans-cli.ts`, and the `operator-notice-cli.ts` row says its apparent driver is not one: `run-daily-fetch.test.ts` drives step 5b through the fake bin deliberately, because the wrapper is what is under test. A following paragraph carries why this accounting has to be explicit — the dishonest 0% used to be the flag, and both files are invisible to a grep for their own filename.

docs/coverage-rationale.md
AmetAlvirde: Fixed in f1ef070. Recounted mechanically: 55 on main, 53 on this branch, and the two missing are exactly the deleted rows. The preamble now reads "That took it to 55… 53 rows, as of that relocation", with the reasoning that the exclude did the very thing both rows asked for — so they were relocated, not resolved. Also re-verified the sibling "18 branch-only" claim still holds against the 53.

vitest.config.ts
AmetAlvirde: Fixed in 192ba5e — the claim, not the behavior. The scope line now says the budget applies to every suite's test bodies, and WHAT IT DOES NOT DO gained a hooks clause: `hookTimeout`/`teardownTimeout` are separate and still 10 s, which is the ceiling the recursive-`rm` `afterEach`es actually meet under the same inflation. `hookTimeout` was deliberately not added — that is a behavior change and a separate decision.

PR #392, opened 2026-08-19
fix(price-feed): give the wrapper harness's case dirs an end

### LINKED ISSUE #390: fix(price-feed): the wrapper harness never deletes its case dirs
## What is wrong

`makeCaseDir` in `apps/price-feed/src/wrapper-harness/case-dir.testkit.ts:69`
mints a fresh `mkdtemp` directory per case and **nothing ever removes it**.
Every wrapper-harness case leaks a directory into the OS temp dir, permanently.

Measured on 2026-08-19: **6,725** `numisma-wrapper-case-*` directories in
`$TMPDIR`, accumulating since Aug 17. They were the dominant contributor to a
1.9 G temp dir; deleting them reclaimed ~900 MB.

## What this is NOT

**Not a flake amplifier**, and specifically not part of #372. Measured directly:
`mkdtempSync` takes **0.057 ms** with 5,281 siblings present versus **0.061 ms**
in an empty directory on APFS. Directory-count has no effect on creation cost,
and temp-dir collision is structurally impossible anyway (`mkdtempSync` is the
atomic unique-directory syscall, and `assertIsolated` refuses any `NUMISMA_*`
path outside the created dir).

This is disk hygiene. It is filed because unbounded growth on a dev machine is a
real cost, not because any test is at risk.

## Fix

Remove the case dir when the case is done — `afterEach`, or a returned disposer,
or `onTestFinished`. Whichever shape is chosen must keep the dir alive long
enough to be inspected when a case **fails**, since these cases assert on files
the wrapper wrote and a post-mortem needs them. An opt-out env var for exactly
that (keep-on-failure, or keep-always while debugging) is worth having.

## Constraint

`case-dir.testkit.ts` is shared — `run-daily-fetch.test.ts` and the
`schedule-window` suites both use it. Any change to its lifecycle has to hold
for every consumer.

AmetAlvirde:
## What this fixes

`makeCaseDir` minted an `mkdtempSync` directory per wrapper-harness case and
nothing ever removed it. 6,725 `numisma-wrapper-case-*` directories had piled up
in `$TMPDIR` since Aug 17, the dominant contributor to a 1.9 G temp dir. That is
what closes #390.

## The shape

The seam is a pure function, `disposeCaseDir(caseDir, { failed })`, and it is
what the new tests exercise directly:

- deletes the directory when the case passed;
- **keeps it when the case failed** — these cases assert on files the wrapper
  wrote, and a post-mortem with the evidence deleted is a mystery;
- keeps it always under `NUMISMA_HARNESS_KEEP_CASE_DIRS`, for debugging;
- refuses, by throwing, any path that is not a direct child of the temp dir
  carrying the `numisma-wrapper-case-` prefix. The refusal runs before the keep
  rules, so the opt-out cannot mask a caller passing the wrong path at a
  recursive delete.

`makeCaseDir` registers that disposal itself through `onTestFinished`. None of
the roughly forty call sites changed. The alternative — a disposer each caller
registers — is one a future case forgets, and the testkit is shared by
`run-daily-fetch.test.ts` and `schedule-window.test.ts` alike.

The delete carries `maxRetries: 3` because the timeout cases deliberately leave
a stray `sleep` alive inside the case dir.

## The measurement rider

Case 4's report line now states each series against the budget it is judged
against, rather than as bare milliseconds, with the worst run's headroom spelled
out:

```
case 4 · reached the step (ms, bound 2000): 398, 425, ... — worst 425 of 2000, 1575 to spare
```

That is what issue 372's comment asked anyone chasing that flake to measure:
per-step latency against its budget, not file duration. The flake itself is
untouched, so 372 stays open as the recorded observation it was filed as. No
assertion and no budget value moved.

## Verified

- `pnpm typecheck` clean.
- `pnpm test`: 2609 passed, 19 skipped. The arming trigger fires on a
  wrapper-harness edit, so this run included the armed cases.
- `pnpm test:wrapper`: 83 passed.
- Leak count 14 before and 14 after both runs.
- Keep-on-failure proven empirically as well as by unit test: a throwaway probe
  with one passing and one deliberately failing case left the failing case's
  directory intact with its full contents, and removed the passing one's.

### DIFF
--- a/apps/price-feed/src/wrapper-harness/case-dir-disposal.test.ts
+++ b/apps/price-feed/src/wrapper-harness/case-dir-disposal.test.ts
@@ -11,12 +11,18 @@
  * real case dir, and no assertion depends on a wrapper having run.
  */
 
-import { existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { chmodSync, existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterAll, afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
-import { CASE_DIR_PREFIX, disposeCaseDir, KEEP_CASE_DIRS_ENV } from "./case-dir.testkit.js";
+import { CONTRACT_MARK_CONFIG } from "./mark-window.testkit.js";
+import {
+  CASE_DIR_PREFIX,
+  disposeCaseDir,
+  KEEP_CASE_DIRS_ENV,
+  makeCaseDir,
+} from "./case-dir.testkit.js";
 
 /** Dirs this file made, removed whatever the assertions did or did not do to them. */
 const minted: string[] = [];
@@ -31,10 +37,32 @@ function mintCaseDir(): string {
   return dir;
 }
 
+/**
+ * THE OPT-OUT IS RESET BEFORE EACH CASE, NOT ONLY AFTER ONE. This file documents a variable
+ * a human exports while debugging, so a human who is doing exactly that runs the suite with
+ * it already in the environment — and an `afterEach`-only reset leaves the FIRST case in
+ * this file inheriting it, which turns the disposal cases red for a reason that has nothing
+ * to do with the code. The ambient value is put back at the end rather than deleted, because
+ * this process does not own the shell it was started from.
+ */
+const ambientKeep = process.env[KEEP_CASE_DIRS_ENV];
+
+beforeEach(() => {
+  delete process.env[KEEP_CASE_DIRS_ENV];
+});
+
 afterEach(() => {
   delete process.env[KEEP_CASE_DIRS_ENV];
   while (minted.length > 0) {
     rmSync(minted.pop() as string, { recursive: true, force: true });
+  }
+});
+
+afterAll(() => {
+  if (ambientKeep === undefined) {
+    delete process.env[KEEP_CASE_DIRS_ENV];
+  } else {
+    process.env[KEEP_CASE_DIRS_ENV] = ambientKeep;
   }
 });
 
@@ -101,6 +129,38 @@ describe("disposeCaseDir", () => {
     }
   });
 
+  /**
+   * CLEANUP DOES NOT GET A VOTE ON THE VERDICT. `rmSync` throws once it exhausts its
+   * retries, and this runs from `onTestFinished`, so a throw here would redden a case that
+   * had already passed — the timeout cases end with the wrapper's strays still writing
+   * inside the case dir, which is exactly how that race is lost. The refusal above still
+   * throws, because that one means the caller is wrong; this one means the filesystem was
+   * busy, which is not a fact about the code under test.
+   *
+   * The unremovable directory is authored, not waited for: a case dir the process may
+   * traverse but not write makes the recursive unlink fail with `EACCES` on the first try,
+   * every time, with no race to lose. Root ignores those bits, so the case states that it
+   * is not describing a root run rather than passing vacuously in one.
+   */
+  it.skipIf(process.getuid?.() === 0)(
+    "warns and returns when the removal fails, so a green case stays green",
+    () => {
+      const dir = mintCaseDir();
+      chmodSync(dir, 0o500);
+      const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
+
+      try {
+        expect(() => disposeCaseDir(dir, { failed: false })).not.toThrow();
+        expect(existsSync(dir), "the removal was supposed to have failed").toBe(true);
+        expect(warn, "a directory left on disk without a word is the silent leak").toHaveBeenCalledTimes(1);
+        expect(String(warn.mock.calls[0]?.[0]), "the warning does not name the directory").toContain(dir);
+      } finally {
+        warn.mockRestore();
+        chmodSync(dir, 0o700);
+      }
+    },
+  );
+
   // ── THE PATH GUARD ────────────────────────────────────────────────────────────────
   //
   // A recursive delete that can be pointed anywhere is not acceptable in a testkit, so it
@@ -116,15 +176,28 @@ describe("disposeCaseDir", () => {
       expect(existsSync(stranger), "the guard threw and deleted the directory anyway").toBe(true);
     });
 
-    it("refuses a correctly-prefixed path that is not in the temp dir", () => {
-      const outsideRoot = mkdtempSync(join(tmpdir(), CASE_DIR_PREFIX));
-      minted.push(outsideRoot);
+    it("refuses a case dir's own child, prefixed or not, because that is not a case dir", () => {
+      const parent = mkdtempSync(join(tmpdir(), CASE_DIR_PREFIX));
+      minted.push(parent);
       // Prefix-correct in its own basename, but one level deeper than a case dir ever is.
-      const impostor = join(outsideRoot, `${CASE_DIR_PREFIX}deeper`);
+      const impostor = join(parent, `${CASE_DIR_PREFIX}deeper`);
       mkdirSync(impostor, { recursive: true });
 
       expect(() => disposeCaseDir(impostor, { failed: false })).toThrow(/refused to remove/);
       expect(existsSync(impostor)).toBe(true);
+    });
+
+    // The case above is still INSIDE the temp tree, and the three literal paths below either
+    // miss the prefix or are relative. This is the one that hands the guard an absolute,
+    // correctly-prefixed, existing directory whose parent is not the temp dir at all — the
+    // shape a caller would produce by building a case dir path against the wrong root.
+    it("refuses a correctly-prefixed directory that lives outside the temp tree", () => {
+      const stranger = join(process.cwd(), `${CASE_DIR_PREFIX}outside-the-temp-tree`);
+      mkdirSync(stranger, { recursive: true });
+      minted.push(stranger);
+
+      expect(() => disposeCaseDir(stranger, { failed: false })).toThrow(/refused to remove/);
+      expect(existsSync(stranger), "the guard threw and deleted the directory anyway").toBe(true);
     });
 
     for (const path of ["/", "/tmp", "relative/numisma-wrapper-case-x", ""]) {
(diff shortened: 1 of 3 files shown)


AmetAlvirde (commented):
Review of the disposer. The design decisions hold up: the pure `disposeCaseDir` seam is the right thing to test, self-registration at the mint is right for the reason the PR gives, keep-on-failure is correct, and I checked the `failed` derivation against vitest 3.2.6 empirically — `pass`/`fail`/timeout all resolve as the PR assumes, `onTestFinished` runs *after* `afterEach`, and its callbacks are LIFO so every existing per-case `onTestFinished` logger runs before disposal (and none of them touch the case dir anyway). I also walked all ~34 `makeCaseDir` call sites across `run-daily-fetch.test.ts` and `schedule-window.test.ts`: every one executes inside an `it` body, including the two behind `schedule-window`'s `runWrapper` helper, so nothing calls `onTestFinished` outside a test context. The cross-test `TermPathObservation` values are eagerly materialized, so nothing reads a case dir after its owning test ends. `assertIsolated` only guards the nine named wrapper vars, so `NUMISMA_HARNESS_KEEP_CASE_DIRS` cannot trip the isolation contract. The #372 rider is faithful: each series is printed against the constant that actually judges it (`REACH_STEP_BOUND_MS` is the bound `waitForFakeToReach` throws at, `ceilingMs` is the value in the `toBeLessThan`, `SETTLE_DEADLINE_MS` is what is handed to `launchWrapper`), and no assertion or budget value moved.

Five findings below, most severe first. Two are worth acting on before merge.

apps/price-feed/src/wrapper-harness/case-dir-disposal.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  const dir = mkdtempSync(join(tmpdir(), CASE_DIR_PREFIX));
+  minted.push(dir);
+  // A file and a nested directory, so "it was removed" means the RECURSIVE removal
+  // happened and not merely an `rmdir` of something that was already empty.
+  mkdirSync(join(dir, "logs"), { recursive: true });
+  writeFileSync(join(dir, "logs", "run.log"), "authored fixture\n", "utf8");
+  return dir;
+}
+
+afterEach(() => {
```
AmetAlvirde: **1 — the documented opt-out turns this file red.** The env reset is `afterEach` only, so the *first* test to execute in this file inherits whatever the shell exported. A developer who follows the documented debugging workflow gets a red suite on an unrelated file.

Verified on this commit:

```
$ NUMISMA_HARNESS_KEEP_CASE_DIRS=1 npx vitest run apps/price-feed/src/wrapper-harness/case-dir-disposal.test.ts
 FAIL  … > disposeCaseDir > removes a passing case's directory, contents and all
 AssertionError: a passing case left its directory behind: expected true to be false
 Tests  1 failed | 21 passed (22)
```

The rest of the file is protected by this hook having already run once. The tests that assert the *off* grammar set the variable explicitly, so they are fine either way; it is the tests that assume it unset that need the reset to happen **before** them. A `beforeEach` that deletes the key (keeping the `afterEach` for the dir teardown) closes it.

apps/price-feed/src/wrapper-harness/case-dir.testkit.ts
```diff
@@ -65,9 +144,24 @@ export interface CaseOptions {
...
+ * each must be judged by the result of the test that minted it. It also runs AFTER the
+ * result state is settled, so `pass` is a fact by the time the callback reads it — and
+ * anything that is not `pass` is treated as a failure, so the doubtful direction keeps the
+ * evidence.
  */
 export function makeCaseDir(options: CaseOptions): CaseDirs {
-  const caseDir = mkdtempSync(join(tmpdir(), "numisma-wrapper-case-"));
+  const caseDir = mkdtempSync(join(tmpdir(), CASE_DIR_PREFIX));
+  onTestFinished((context) => {
+    disposeCaseDir(caseDir, { failed: context.task.result?.state !== "pass" });
```
AmetAlvirde: **2 — nothing pins the wiring, so the fix for #390 is not regression-proof.** `case-dir-disposal.test.ts` imports `CASE_DIR_PREFIX`, `disposeCaseDir` and `KEEP_CASE_DIRS_ENV` — never `makeCaseDir`. Delete these three lines and every one of the 22 new tests stays green, `pnpm typecheck` stays clean, and the leak is back exactly as #390 described it, silently.

That is the one seam the PR argues hardest for ("a disposer the CALLER has to register is one the forty-first call site forgets"), and it is the only part of the change with no test. A case that calls `makeCaseDir` inside a trivial passing `it` and asserts, in a later `onTestFinished` or via a captured path, that the directory is gone would pin it — the LIFO ordering I confirmed means a `onTestFinished` registered *before* `makeCaseDir` runs *after* the disposal, which gives exactly the observation point needed.

apps/price-feed/src/wrapper-harness/case-dir.testkit.ts
```diff
@@ -25,15 +25,94 @@
...
+    throw new Error(
+      `wrapper harness refused to remove ${path}: only a direct child of ${tmpdir()} named ` +
+        `\`${CASE_DIR_PREFIX}*\` is a case dir, and this is a RECURSIVE delete`,
+    );
+  }
+
+  if (options.failed || keepAlways()) {
+    return;
+  }
+  rmSync(path, { recursive: true, force: true, maxRetries: 3, retryDelay: 50 });
```
AmetAlvirde: **3 — losing the retry race turns a passing case red.** `rmSync` throws when it exhausts `maxRetries`, and a throw inside `onTestFinished` fails an otherwise-green test. Verified:

```js
it("green body, throwing disposer", () => {
  onTestFinished(() => { throw new Error("disposer boom"); });
  expect(1).toBe(1);
});
// → Tests  1 failed (1)
```

The scenario is the one the retries are there for, taken one step further. Cases 2, 3 and 7 deliberately leave the wrapper's strays alive inside the case dir at the moment the case ends — case 3's premise is specifically a TERM-deaf grandchild holding `tee` alive on the log file under `logDir`. A `tee` still appending while the recursive walk is unlinking is an `ENOTEMPTY`; three retries at 50 ms is 150 ms of tolerance against a process the case never waits for. When that budget is exceeded, disk hygiene fails a case that passed, in the armed suite, non-deterministically — a new flake in the family #372 is already tracking.

The guard's throw above is deliberate and should stay a throw: that one means the caller is wrong. This one means the filesystem was busy, which is not a fact about the code under test. Wrapping just the `rmSync` in a `try`/`catch` that warns keeps the failure visible without letting cleanup vote on the verdict.

apps/price-feed/src/wrapper-harness/case-dir.testkit.ts
```diff
@@ -65,9 +144,24 @@ export interface CaseOptions {
...
+ * `onTestFinished` rather than an `afterEach`, because a case may mint several dirs and
+ * each must be judged by the result of the test that minted it. It also runs AFTER the
+ * result state is settled, so `pass` is a fact by the time the callback reads it — and
+ * anything that is not `pass` is treated as a failure, so the doubtful direction keeps the
+ * evidence.
  */
 export function makeCaseDir(options: CaseOptions): CaseDirs {
-  const caseDir = mkdtempSync(join(tmpdir(), "numisma-wrapper-case-"));
+  const caseDir = mkdtempSync(join(tmpdir(), CASE_DIR_PREFIX));
+  onTestFinished((context) => {
```
AmetAlvirde: **4 — `skip` is not `failed`, and treating it as such leaks (latent).** Confirmed against 3.2.6: a runtime `ctx.skip()` leaves `context.task.result.state === "skip"`, which `!== "pass"` classifies as failed, so the dir is kept permanently. A skipped case has no evidence to preserve — the doubtful-direction argument in the doc comment does not cover it — so this is #390 reinstated for that one case.

Not reachable today: the only skips in the two consumers are `describe.runIf(decision.run)` and `it.skipIf(!onMac)`, both collection-time, so the body never runs and `makeCaseDir` is never called. It becomes live the moment any case skips itself mid-body after minting a dir. `state === "fail"` says what is actually meant here, and `fail` is also what a timeout and an `afterEach` throw produce (both verified), so nothing else changes.

apps/price-feed/src/wrapper-harness/case-dir-disposal.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  describe("the path guard", () => {
+    it("refuses a temp-dir sibling that does not carry the harness prefix", () => {
+      const stranger = mkdtempSync(join(tmpdir(), "not-a-numisma-case-"));
+      minted.push(stranger);
+
+      expect(() => disposeCaseDir(stranger, { failed: false })).toThrow(/refused to remove/);
+      expect(existsSync(stranger), "the guard threw and deleted the directory anyway").toBe(true);
+    });
+
+    it("refuses a correctly-prefixed path that is not in the temp dir", () => {
```
AmetAlvirde: **5 — the test name says something the test does not do.** This path *is* in the temp dir; it is one level too deep. The case is worth keeping (it exercises the `roots.has(dirname(path))` clause), but the name promises the coverage that is actually missing: no test hands the guard an absolute, correctly-prefixed path outside the temp tree entirely — say `<repo>/numisma-wrapper-case-x`. `/` and `/tmp` fail on the prefix clause, and `relative/numisma-wrapper-case-x` never reaches the root check because `isAbsolute` rejects it first. Rename this to what it pins ("a case dir's own child, not a case dir") and add the outside-the-temp-tree path as its own case.

apps/price-feed/src/wrapper-harness/case-dir-disposal.test.ts
AmetAlvirde: Fixed in c2eabfa. A `beforeEach` now clears the variable, the `afterEach` stays, and an `afterAll` restores the ambient value captured at module load. Proven by emptying the `beforeEach` and re-running with the variable exported: 1 failed / 25 passed.

apps/price-feed/src/wrapper-harness/case-dir.testkit.ts
AmetAlvirde: Fixed in c2eabfa, and this was the finding worth the most. A new describe block pins the wiring with four cases: one mints a real case dir and stashes the path, the next asserts it is gone, then the same pair for a case that skips itself. Proven by deleting the three registration lines: 2 failed / 24 passed.

apps/price-feed/src/wrapper-harness/case-dir.testkit.ts
AmetAlvirde: Fixed in c2eabfa. Only the `rmSync` is wrapped; the path guard still throws, because a misdirected recursive delete has to break the build. A failed delete now warns with the path and returns, so cleanup never votes on the verdict. Pinned deterministically by chmod 0o500 on a case dir, which fails with EACCES rather than waiting on a race, and skipped under uid 0.

apps/price-feed/src/wrapper-harness/case-dir.testkit.ts
AmetAlvirde: Fixed in c2eabfa. The dir is kept only when the state is `fail` or missing; `pass`, `skip` and `todo` all delete. Unknown-keeps is deliberate and now documented: an unreadable verdict should preserve evidence. Proven by restoring `state !== "pass"`, which left the skipped case's dir behind.

apps/price-feed/src/wrapper-harness/case-dir-disposal.test.ts
AmetAlvirde: Fixed in c2eabfa. Renamed to `refuses a case dir's own child, prefixed or not, because that is not a case dir`, and the missing case is now covered by `refuses a correctly-prefixed directory that lives outside the temp tree`, which builds a real one under the cwd and cleans it up.

PR #393, opened 2026-08-19
test(price-feed): race case 4's wait against the watchdog, not a stopwatch

### LINKED ISSUE #394: Flake (recorded, with a reproduction): the child-reap mutation control reds under parallel load when its orphan outlives its own poll hop
**Filed as a recorded observation with a reproduction, not a request to chase it.** It has a
mechanism and a way to trigger it on demand, so the next lane that sees this red has
something better than an hour of re-deciding.

## The observation

`apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts`

> `the child-reap mutation — assertion 3 is seen RED before it is trusted > makes the healthy case go RED on a stray process left in the run's pgid`

```
AssertionError: the child-reap mutation left NOTHING in the pgid — assertion 3 is no
longer guarding anything: expected 0 to be greater than 0
```

**Reproduced 3 times in 5 attempts** under sustained two-lane load, and **never once in 5
quiet full-suite runs** (60 armed cases each). The trigger is below, and it is reliable enough to work with.

## How to reproduce

Run the armed suite with a second full suite restarted continuously alongside it, so the
whole armed run is contended rather than just its opening minutes:

```sh
( while [[ ! -f /tmp/.load-stop ]]; do NUMISMA_WRAPPER_TEST=never pnpm test >/dev/null 2>&1; done ) &
pnpm test:wrapper; touch /tmp/.load-stop
```

One `pnpm test` alongside is **not** enough — it finishes in ~65 s against the armed run's
~555 s, so it is long gone before this case runs. That mistake is why the first attempt at
this measurement showed nothing.

## The mechanism

`CASE_OPTIONS.maxRunSeconds` is 30, so the wrapper's watchdog is above the clamp and polls
in **`sleep 5` hops**. The mutation removes the child reap, so when the run finishes its
EXIT trap SIGKILLs the watchdog subshell and **orphans that hop's `sleep` into the run's
process group** — which is exactly the residue assertion 3 exists to catch, and exactly what
this control asserts is present.

How much of the hop is left when residue is sampled depends on **where in the 5-second cycle
the run happened to finish**. A healthy fake run normally completes early in the first hop
and leaves ~4.5 s of orphan, which is an enormous margin and why this is invisible on a quiet
machine. Under load the run stretches toward the hop boundary, and a run that finishes near
it leaves an orphan with milliseconds to live — gone before `pgidResidue` is read.

**This is a race on the PHASE of a 5 s cycle, not a monotone budget**, and the distinction
matters for whoever fixes it: there is no headroom number to read off a report line and no
threshold to widen. The failure is not "the machine got slower than N", it is "the run landed
on the wrong side of a boundary". It also means the red is honest in one narrow sense — the
control genuinely did not observe what it claims to observe — so it must not be silenced by
loosening the assertion.

## Related, and deliberately NOT the same thing

Same class as #372 names — a wrapper-harness case with a fixed wall-clock assumption that
only holds on an unloaded machine — but a **different mechanism**, which is why it is filed
separately rather than as a comment there. #372's case-4 sighting was a monotone budget being
eaten (measured: 388-678 ms quiet, 1473 ms contended, against a flat 2000 ms bound) and was
answered in #393 by making that wait race the watchdog's calling card instead of a clock.
Nothing in that change touches this control, and this one still reds in runs where case 4 now
passes with 1846 ms to spare.

## What would settle it

Not scheduled, listed so whoever picks it up does not start cold:

- Shorten the watchdog's poll for this case so the orphan's remaining life is not a 5-second
  lottery — `maxRunSeconds` below the 5 s clamp makes the hop equal the ceiling.
- Or sample `pgidResidue` on a bounded poll that starts the instant the shell exits, rather
  than reading it once, so a short-lived orphan is still seen.
- Or have the mutation orphan something that does not expire on its own, making the control's
  subject independent of the watchdog's poll phase entirely.

AmetAlvirde:
## What this changes

Case 4's `waitForFakeToReach` enforced a flat **2000 ms** before it signalled the run. That
number was never the thing the case is about — it was a **proxy** for case 4's actual
premise, *the signal must arrive before the watchdog fires*, picked as a safety strip under
the timeout cases' 3 s ceiling.

#372 recorded the proxy going red under a full parallel suite, with the message

> the run never reached the step this signal is aimed at

which was **false**. The run reached the step. The machine was busy.

## The measurement that decided it

Twelve runs per suite run, read off case 4's own report line, on 2026-08-19:

| condition | worst reach per run | against the old 2000 ms bound |
|---|---|---|
| quiet machine, 5 full suite runs (60 samples) | 416, 466, 544, 472, 678 | 1322 ms to spare |
| **two lanes, 3 full suite runs (36 samples)** | **1473, 1460, 1282** | **527 ms to spare — 74% consumed** |

The contended band reproduces the series #372's comment recorded from Lane D (1234, 702, 840,
872) almost exactly, and Lane D's red is that same series crossing. **Scheduler contention
alone ate three quarters of the bound**, which is not a margin the issue could be closed on.

Method note, because the first attempt was wrong and someone will repeat it: a *single*
concurrent `pnpm test` proves nothing here. It finishes in ~65 s against the armed run's
~555 s, so it is long gone before case 4 runs, and the reach barely moves (worst 509 ms). The
load generator has to **restart itself** until the armed run finishes.

## The shape

The watchdog announces itself — it touches its calling card **before** it signals the group
(`ops/price-feed/run-daily-fetch.sh`, the watchdog subshell). So the losing condition is
directly observable, and the wait no longer races a clock at all. Three exits, each meaning
exactly one thing:

- **the sentinel appears** — the run is in the step; the elapsed time is returned for the
  report line;
- **the calling card appears** — the watchdog beat the harness, so this run is case 2 wearing
  case 4's name and there is nothing left to signal;
- **the backstop expires** — neither happened, which is a wrapper that is not running and not
  being bounded by its own watchdog.

`REACH_STEP_BACKSTOP_MS` is derived from `ceiling + grace + 2 s` and sits **above** everything
the watchdog does. The card itself lands *at* the ceiling, before the TERM and so before the
grace, but the wrapper writes it best-effort (`|| true`), so on a failed write the last event
the watchdog still guarantees is the SIGKILL at ceiling + grace. Reaching the backstop
therefore means a watchdog that never armed at all, which is the only thing it is meant to
catch. A watchdog that genuinely got there first still fails — and fails *saying so*, instead
of claiming the run never reached a step it plainly reached.

**What this does not buy, corrected after review.** An earlier revision of this description
said "a slow machine now costs this case latency and never a red". That is false, and the
review was right to call it. Case 4 still reds past the ceiling — at the card check, and
slightly earlier at `durationMs < ceilingMs`, which is measured from `spawn` rather than from
the watchdog's own `SECONDS=0` and is now the **binding** wall-clock bound of the case. What
the change actually buys is headroom: the tightest bound went from a flat 2000 ms to roughly
the ceiling minus the signal-to-exit tail, against a measured contended worst of 1541 ms, and
the red at the far end names the watchdog instead of accusing the run of never arriving. Both
the `waitForFakeToReach` comment and that assertion's message now say this.

**No assertion moved and no budget was raised.** The one behavioural change beyond the wait is
the report line, which now judges the reach against the ceiling it actually races instead of
the retired strip.

The `watchdogCardPresent` check that already followed the wait is kept, narrowed to the one
instant the wait cannot cover: the wait only reads the card while the sentinel is *absent*, so
a run whose sentinel and card land in the same 25 ms poll gap would come back "reached" and be
signalled anyway.

## Verified

- `pnpm typecheck` clean.
- `pnpm test:wrapper`: **88 passed**, 1 skipped.
- **Both new exits proven to fire, rather than assumed reachable.** Waiting on a sentinel the
  fake never writes failed in 3.1 s with `the watchdog left its calling card after 3036ms,
  before the fake recorded ... so it is a second copy of case 2 rather than an external stop`.
  The same probe with the watchdog switched off reached the backstop and said `the run reached
  neither its step nor its own ceiling`. Both probes were thrown away; the file is byte
  identical to the verified state.
- **Under the same two-lane load that ate 74% of the old bound**: worst reach 1154 and 1541 of
  3000, leaving 1846 ms and 1459 ms to spare.

**CI does not verify any of this, and the green check should not be read as if it did.** An
earlier revision of this description claimed it would, on the grounds that this branch edits
`apps/price-feed/src/wrapper-harness/**` and so trips `TRIGGER_PATH_SET`. That is true and
irrelevant: the arming trigger evaluates its **platform gate first**, `HARNESS_PLATFORM` is
`darwin`, and CI is `ubuntu-latest`. Every armed case is skipped there — which is why the CI
`test` job finishes in ~1m44s against ~555s locally. The evidence for this change is the local
runs listed above, on darwin, and nothing else.

## What this does NOT do

**#372 is not closed by this.** It was filed over two sightings and this answers one of them.
The `schedule-window.test.ts` sighting did not reproduce in any of the 8 full-suite runs here,
quiet or contended, and its own bound (a 60 s `spawnSync` cap on a file that costs 3.3-4.0 s
under load) is nowhere near thin. That half stays recorded, and the disposition comment on
#372 says so explicitly.

A **second, unrelated flake** surfaced while measuring. The child-reap mutation control went
red in 3 of 5 contended runs, including one where case 4 passed with 1846 ms to spare. It has
a different mechanism (a race on the phase of the watchdog's 5 s poll, rather than a monotone
budget), so it was filed separately as #394. **The second commit on this branch answers it**,
described below.

---

## Second commit: the child-reap control stops being a lottery (#394)

`test(price-feed): repeat the child-reap control against the poll phase, not once`

### The mechanism

The control asserts that the un-reaped watchdog child is still in the run's pgid when the
settle window closes. That child is a `sleep` of **one poll hop that was already running**
when the shell exited, so how much of it is left is decided by where in the watchdog's
repeating 5 s cycle the run happened to finish. Quiet, a fake run finishes early in a hop and
the stray has ~4.5 s left. Loaded, the run's own duration wanders across the cycle, and a run
that finished near a boundary leaves a stray with milliseconds to live, gone before residue is
read.

This is a **phase**, not a budget. A longer settle window makes the control blinder; a shorter
one makes assertion 3's own reading unrepresentative. There is no threshold to widen.

### The fix, and the candidate that measurement killed

The control now **repeats up to the same 12-run floor the tee-deafness control uses**, stopping
at the first attempt that observes the residue. Same racy-phenomenon argument, same file, same
pattern.

Rejected on measurement rather than taste: **#394's own first suggestion**, dropping
`maxRunSeconds` under the 5 s clamp so the watchdog sleeps exactly one hop and the orphan's
life becomes `ceiling − run duration`, a monotone budget. It was built and measured, and it is
**worse**:

| variant, same load, same session | contended reds |
|---|---|
| unchanged control | 3 of 6 |
| one-hop 5 s ceiling, 3000 ms budget | **8 of 12** |

The mutated run took 3606, 3651, 3603, 3921, 3795, 4223, 4264 and 3860 ms against that 3000 ms
budget. Under sustained load a healthy run already spends **85% of a 5 s ceiling**, so the
ceiling is not spare room to spend, it is the case's own margin.

### What did not move

**Nothing accepts an absence as a pass.** The loop still demands that assertion 3's own field,
read through assertion 3's own window, came back occupied at least once. `toBeGreaterThan(0)`
is untouched, the healthy comparison run is untouched, and both sides still take the same
`CASE_OPTIONS`.

**Corrected after review:** an earlier revision of this branch asserted on every attempt that
`residueAtExit` contained a `sleep`, and this description called that reading "race-free". It
is not. `residueAtExit` is sampled after the `exit` event, after `duringRun` is awaited, and
after a `ps -A -ww` has forked and returned, so an *empty* reading is the same phase draw as
the settle window's over a shorter interval. Asserting on it re-created #372's misdiagnosis one
control over, and it threw inside the loop rather than spending an attempt. Each attempt is now
classified into four outcomes — `SEEN`, `expired`, `gone-at-sample`, `NO-ORPHAN` — of which
only the unambiguous one (`NO-ORPHAN`: the pgid occupied by something that is not the timer)
fails fast. `gone-at-sample` costs an attempt and is named in the loop-level message.

`SETTLE_DEADLINE_MS`'s comment claimed the stray had "most of a 5-second poll" left. That was
the quiet-machine assumption #394 disproved, so it now states what actually decides the
remainder and why the control repeats.

### Verified

- `pnpm typecheck` clean.
- `pnpm test:wrapper`: **88 passed**, 1 skipped, 498 s.
- **Under sustained two-lane load, 12 contended repetitions of this control: 12 passed.** The
  repetition earns its keep rather than getting lucky: **14 expiries across 25 attempts**, a
  56% per-attempt miss rate, worst case 4 attempts of the 12 available.
- The **pre-fix red captured under the same harness** minutes earlier, so the before/after is a
  comparison: 3 of 6 contended runs failed with `expected 0 to be greater than 0`, 0 of 3 quiet.

The CI caveat above applies here too: every armed case is skipped on `ubuntu-latest`, so the
green check is not evidence for any of this.

**This closes #394.** #372 is still not closed by this PR, for the reasons above.

---

## Review response (`pullrequestreview-4976862184`)

Two commits: `dd575dc` (case 4's bounds) and `3a69835` (the child-reap control's outcomes).
Findings 1-5 fixed, 6 answered with a comment rather than a rename, 7 left alone as
pre-existing.

**Verified after the fixes, on darwin:**

- `pnpm typecheck` clean.
- `pnpm test:wrapper`: **88 passed, 1 skipped, 493 s.**
- **10 contended repetitions of the child-reap control** under a continuously restarted
  `NUMISMA_WRAPPER_TEST=never pnpm test`: **10 passed, 21 attempts spent** — `expired ×2, SEEN`
  / `expired ×3, SEEN` / `SEEN` / `SEEN` / `SEEN` / `expired ×2, SEEN` / `SEEN` / `expired,
  SEEN` / `SEEN` / `expired, SEEN`. The load was real: mutated runs stretched to 3.3-4.5 s
  against 0.6-0.9 s quiet.
- **Both new outcomes proven to fire rather than assumed reachable**, with throwaway probes
  reverted afterwards. Pointing the loop at the *unmutated* wrapper spent all twelve attempts
  as `gone-at-sample` and failed with the loop-level message naming that count, with no
  fail-fast. Synthesizing a pgid holding a non-`sleep` process failed on attempt 1 with `the
  child-reap mutation left the run's pgid occupied by something that is not the watchdog's
  timer` — and the outcome series printed that attempt, which the old ordering would have
  dropped.

Note on the contended series: no attempt came back `gone-at-sample` in those 10 repetitions,
so the shape the old assertion would have thrown on did not occur in this session. Its
handling is proven by the probe above, not by the load run.

### DIFF
--- a/apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
+++ b/apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
@@ -695,12 +695,32 @@ function watchdogCardPresent(logDir: string): boolean {
 }
 
 /**
- * How long case 4 will wait for the run to actually be INSIDE the fetch step before it
- * signals. Bounded well under the timeout cases' 3s ceiling on purpose: the whole claim of
- * case 4 is that its signal arrives before the watchdog's, so a wait that could run into the
- * ceiling would be a wait that could silently turn case 4 into a second copy of case 2.
+ * THE ANTI-HANG BACKSTOP ON CASE 4'S WAIT — deliberately NOT a budget, and the distinction
+ * is the whole point of this constant's existence.
+ *
+ * It used to be one: a flat 2000 ms, chosen as a safety strip under the timeout cases' 3 s
+ * ceiling, on the reasoning that a wait which ran into the ceiling could silently turn case
+ * 4 into a second copy of case 2. The reasoning was right; the mechanism was a PROXY for it,
+ * and the proxy is what went red. #372 recorded this exact wait failing under a full
+ * parallel suite with the message "the run never reached the step this signal is aimed at" —
+ * a statement that was simply FALSE. The run reached the step. The machine was busy.
+ *
+ * Measured on 2026-08-19, twelve runs per suite run: on a quiet machine the reach costs
+ * 388-678 ms across 60 samples, but with a second suite running concurrently — the board's
+ * ordinary condition, one worktree per lane — the worst run of each set was 1473, 1460 and
+ * 1282 ms. That is 74% of the old bound consumed by scheduler contention alone, and #372's
+ * original sighting is the same series crossing it.
+ *
+ * So the wait no longer races a clock at all. It races {@link watchdogCardPresent}, which is
+ * the thing case 4's premise is ACTUALLY about, and this number is what remains: a bound on
+ * a run that is neither reaching its step nor being timed out by its own watchdog, which is
+ * a broken wrapper rather than a busy machine. It sits above ceiling + grace so the card
+ * always wins the race on a merely-slow run, and any run that reaches it has outlived the
+ * watchdog that was supposed to bound it.
  */
-const REACH_STEP_BOUND_MS = 2_000;
+const REACH_STEP_BACKSTOP_MS =
+  (TIMEOUT_CASE_OPTIONS.maxRunSeconds + TIMEOUT_CASE_OPTIONS.watchdogGraceSeconds) * 1_000 +
+  2_000;
 
 /**
  * Block until the fake has recorded `command`, so a signal aimed at a step lands while the
@@ -708,22 +728,48 @@ const REACH_STEP_BOUND_MS = 2_000;
  * heartbeat is half of what case 4 asserts, and `LAST_STEP` is set just before the call the
  * sentinel proves happened.
  *
- * FAIL-CLOSED AND BOUNDED. A sentinel that never arrives means the run never reached the
- * step this signal is aimed at, which is a failure of the case; waiting for it with
- * unbounded patience would be indistinguishable from a run that hung before it got there.
+ * FAIL-CLOSED, AND IT RACES THE WATCHDOG RATHER THAN A CLOCK. What case 4 needs is not "the
+ * step arrived inside N milliseconds" but "the step arrived before the watchdog fired", and
+ * those two are only the same statement on an unloaded machine. The watchdog announces
+ * itself: it touches its calling card BEFORE it signals the group
+ * (`run-daily-fetch.sh`, the watchdog subshell), so the losing condition is directly
+ * observable and needs no proxy. Waiting on the card instead of a stopwatch means a slow
+ * machine costs this case latency and never a red, while a watchdog that genuinely got there
+ * first still fails — and fails saying so, instead of claiming the run never reached a step
+ * it plainly reached. See {@link REACH_STEP_BACKSTOP_MS} for the measurements that retired
+ * the stopwatch.
+ *
+ * THREE EXITS, EACH MEANING EXACTLY ONE THING:
+ *   - the sentinel appears — the run is in the step, and the elapsed time is returned for
+ *     the report line;
+ *   - the calling card appears — the watchdog beat the harness, so this run is case 2
+ *     wearing case 4's name and there is nothing left to signal;
+ *   - the backstop expires — neither happened, which is a wrapper that is not running and
+ *     not being bounded by its own watchdog.
  */
 async function waitForFakeToReach(
   caseDir: string,
   command: string,
   label: string,
+  logDir: string,
 ): Promise<number> {
   const sentinel = join(caseDir, "sentinels", sentinelNameFor(command));
   const startedAt = Date.now();
   while (!existsSync(sentinel)) {
-    if (Date.now() - startedAt >= REACH_STEP_BOUND_MS) {
+    // CHECKED BEFORE THE BACKSTOP, because it is the informative half: a run the watchdog
+    // has already claimed should say THAT, not report a generic expiry.
+    if (watchdogCardPresent(logDir)) {
       throw new Error(
-        `${label}: the fake never recorded \`${command}\` within ${REACH_STEP_BOUND_MS}ms — ` +
-          "the run never reached the step this signal is aimed at",
+        `${label}: the watchdog left its calling card after ${Date.now() - startedAt}ms, ` +
+          `before the fake recorded \`${command}\` — the watchdog beat the harness to this ` +
+          "run, so it is a second copy of case 2 rather than an external stop",
+      );
+    }
+    if (Date.now() - startedAt >= REACH_STEP_BACKSTOP_MS) {
+      throw new Error(
+        `${label}: the fake never recorded \`${command}\` within ` +
+          `${REACH_STEP_BACKSTOP_MS}ms, and the watchdog never fired either — the run ` +
+          "reached neither its step nor its own ceiling",
       );
     }
     await sleep(25);
@@ -2291,8 +2337,16 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
       // headroom named. A bare list of milliseconds needs the reader to remember which
       // constant each one is racing, and #372 records the cost of getting that wrong: the
       // file's 475s wall clock looked like the signal and was noise, while the number that
-      // actually moved was one step running at ~60% of its own 2000ms budget. Headroom is
-      // the quantity that goes to zero before a budget flips red, so it is the one printed.
+      // actually moved was one step running at ~60% of the flat 2000ms budget the reach
+      // used to carry. Headroom is the quantity that goes to zero before a bound flips red,
+      // so it is the one printed.
+      //
+      // That budget is gone — the reach now races the watchdog's calling card rather than a
+      // stopwatch (see `waitForFakeToReach`), so its series is judged against the ceiling
+      // like the duration beside it. The series is still worth printing, and arguably worth
+      // more: it is now the only place the contention this case actually runs under is
+      // visible at all, and the measurements in `REACH_STEP_BACKSTOP_MS` came from reading
+      // exactly this line across quiet and contended runs.
       onTestFinished(() => {
         const against = (series: readonly number[], bound: number): string => {
           if (series.length === 0) {
@@ -2302,8 +2356,12 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
           return `${series.join(", ")} — worst ${worst} of ${bound}, ${bound - worst} to spare`;
         };
         console.log(
-          `[wrapper harness] case 4 · reached the step (ms, bound ${REACH_STEP_BOUND_MS}): ` +
-            `${against(reaches, REACH_STEP_BOUND_MS)} · ran for (ms, ceiling ${ceilingMs}): ` +
+          // JUDGED AGAINST THE CEILING, because that is the bound the wait now races: the
+          // watchdog's card lands there, and the backstop above it only catches a wrapper
+          // that never ran. Reporting these against the retired 2000ms strip would print
+          // headroom this case no longer spends.
+          `[wrapper harness] case 4 · reached the step (ms, ceiling ${ceilingMs}): ` +
+            `${against(reaches, ceilingMs)} · ran for (ms, ceiling ${ceilingMs}): ` +
             `${against(durations, ceilingMs)} · settled (ms, deadline ${SETTLE_DEADLINE_MS}): ` +
             `${against(settles, SETTLE_DEADLINE_MS)}`,
         );
@@ -2331,10 +2389,17 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
           // test, never as a wedged worker.
           maxWaitMs: 60_000,
           duringRun: async (pgid) => {
-            reaches.push(await waitForFakeToReach(dirs.caseDir, "prices:fetch", label));
+            reaches.push(await waitForFakeToReach(dirs.caseDir, "prices:fetch", label, dirs.logDir));
             // BEFORE THE CEILING, ASSERTED RATHER THAN ASSUMED. If the watchdog got there
             // first this is case 2 wearing case 4's name — a run that would still exit 124
             // and still pass an exit-code-only test of "something stopped it".
+            //
+            // THE WAIT ABOVE ALREADY RACES THIS CARD, so what is left here is the last
+            // instant it cannot cover: the wait only reads the card while the sentinel is
+            // ABSENT, so a run whose sentinel and card land in the same poll gap would come
+            // back "reached" and be signalled anyway. Re-reading it immediately before the
+            // kill closes that gap. Cheap, and the alternative is signalling a run the
+            // watchdog has already claimed.
             expect(
               watchdogCardPresent(dirs.logDir),
               `${label}: the watchdog fired BEFORE the harness could signal — this run is a ` +
--- a/apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
+++ b/apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
@@ -591,9 +591,18 @@ const RUNS_PER_CASE = decision.run ? resolveRunCount(process.env.NUMISMA_WRAPPER
 
 /**
  * The settle window for a case with no timeout in it. Short ON PURPOSE, and the shortness
- * is load-bearing: the child-reap mutation leaves a stray `sleep` with most of a
- * 5-second watchdog poll still to run, so a generous window here would let the mutated run
- * pass and the control would prove nothing.
+ * is load-bearing in the child-reap control: the mutation leaves a stray `sleep` with PART
+ * of a watchdog poll hop still to run, so a generous window here would outlast the stray on
+ * every run and the control would prove nothing.
+ *
+ * HOW MUCH of the hop is left is not a property of this number. It is decided by where in
+ * the watchdog's repeating poll cycle the run happened to finish, which is ambient — a
+ * phase, not a budget, and no value here makes it one. That is why the control repeats
+ * instead of trusting a single reading; #394 is the red that taught it to.
+ *
+ * As a bound on an ordinary case this is generous rather than tight. A healthy run reaps the
+ * watchdog and its child from inside the EXIT trap, so its pgid is already empty at the
+ * instant the shell exits and none of this window is ever spent.
  */
 const SETTLE_DEADLINE_MS = 2_000;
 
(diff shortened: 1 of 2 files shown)


AmetAlvirde (commented):
Reviewed at `1a2b849`, both commits, read against `ops/price-feed/run-daily-fetch.sh` and `launcher.testkit.ts` rather than against the PR body. `pnpm typecheck` is clean. I did not run `pnpm test:wrapper`.

I confirmed the CI caveat independently rather than taking it: `decideArming` evaluates the platform gate first (`arming-trigger.testkit.ts:131`, `HARNESS_PLATFORM = "darwin"`), and `ci.yml` runs `ubuntu-latest`, so every armed case is skipped there. The green check is not evidence for anything in this diff.

**One blocking finding, three worth fixing, two notes.** Ranked, inline where a line owns them.

## Blocking

1. `run-daily-fetch.test.ts:2670-2679`: the new per-attempt `residueAtExit` assertion re-introduces the phase race, fails fast on it, and misdiagnoses it. Detail inline.

## Worth fixing before merge, not blocking

2. `:2401` / `:2437-2442`: case 4's wall-clock tripwire moved onto `durationMs < ceilingMs`; it did not disappear, and that assertion is measured from a different zero than the ceiling it names. Detail inline.
3. `:2690`: `attempts.push` runs after the assertions, so the attempt that throws is missing from the outcome log the comment at `:2636-2638` calls "the whole diagnosis".
4. `:2626`: the compound-probability claim. `0.56^12` is about one in a thousand, not four; four in a thousand needs a 63% miss rate. And independence is asserted rather than shown.

## Notes

5. `:730`: the `+ grace` term in `REACH_STEP_BACKSTOP_MS` is inert, and the comment names it as load-bearing.
6. `RUNS_PER_CASE` means two different things in this file now. Everywhere else it is "repeat N times so a 1-in-12 failure is seen". In this control and in the tee-deafness control it is "retry up to N times until one succeeds", so `NUMISMA_WRAPPER_TEST_RUNS=50` for a soak makes these two controls strictly harder to fail rather than more thorough. The tee-deafness control set the precedent; this commit doubles the surface. Worth one sentence in `repetition.testkit.ts` so the knob's two meanings are written down somewhere.
7. Minor: the 600s test timeout is the file's house number, but this loop's own budget is 12 x `maxWaitMs` of 60s plus a healthy run, so a systemic wedge reports as a bare vitest timeout instead of the launcher's "exceeded the harness cap" message. Same shape already exists at `:2810`, so this is consistency, not a regression.

## Verified sound, listed so the next reader does not re-derive it

- **The `watchdogCardPresent` check kept after the wait (`:2406-2418`) is not dead code.** The wait reads the card only while the sentinel is absent, so a card that lands in the same 25ms gap as the sentinel is invisible to it. The re-read covers exactly that instant, and the comment describing it is accurate.
- **Card-before-backstop ordering is right.** The backstop sits above the card's arrival, so testing it first could never change an outcome, and testing the card first is what makes the message specific instead of generic.
- **The cases 2 and 4 pairing survives.** Both still take `TIMEOUT_CASE_OPTIONS`, the shared 3s ceiling and 2s grace are unchanged, no assertion moved in either case, and the S-pair comparison below them is untouched by this diff.
- **The final `expect(seen)` message is now earned rather than assumed.** Because every attempt asserts the orphan existed at exit, "orphaned a `sleep` on every one of N runs and it had expired" is something the loop has actually established. That is a real improvement over the old single-shot message.
- **The control still does not exercise assertion 3 itself.** It reads `record.pgidResidue` directly, so deleting step (3) from `assertTriple` (`:918-923`) leaves this control green while the thing it is named after is gone. This is pre-existing, unchanged by this diff, and I raise it only because the `describe` name promises otherwise. Not a merge concern for this PR.

Net: both commits move the suite in the right direction and the measurement work behind them is real. The blocking finding is a five-line change, which is why I am calling it blocking rather than filing it as a follow-up: it is the same defect class the branch exists to remove, one control over.

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
```diff
@@ -2533,33 +2607,99 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
...
+          // sleeping out its hop when the shell exits whatever the phase is, so this reading
+          // is the one the race cannot reach — and a red here says the mutation has stopped
+          // biting, which no amount of repeating would fix and which the loop below would
+          // otherwise report as an expiry.
+          expect(
+            record.residueAtExit.join(" | "),
+            `attempt ${attempt}: the child-reap mutation orphaned NOTHING into the pgid at the ` +
+              `instant of exit (${record.residueAtExit.length} process(es) there) — the reap it ` +
+              "removes is no longer what keeps the group clean",
+          ).toMatch(/sleep/);
```
AmetAlvirde: **BLOCKING. This assertion re-introduces the phase race it is documented as immune to, ends the test rather than spending an attempt on it, and blames the wrapper when it loses.**

The comment above calls this reading "THE HALF THAT OWES NOTHING TO THE PHASE" and "race-free". It is neither.

`residueAtExit` is not sampled at the instant of exit. `launcher.testkit.ts:255-256` samples it after the child's `exit` event, after `await during`, and after an `execFileSync("ps", ["-A", "-ww", ...])` has forked, run and returned. I measured `ps -A -ww -o pid=,pgid=,command=` at 50ms wall on an idle machine here, before node's spawn overhead and before whatever the sustained two-lane load in #394's reproduction adds. So what this assertion actually observes is "the orphan was still alive some tens to low hundreds of milliseconds after the shell exited", and the orphan's remaining life is the same uniform draw across the 5s hop that #394 is entirely about. At a 60ms sampling window that is roughly 1.2% per attempt; at 200ms, roughly 4%. The loop stops at the first success, so under the load regime that reproduced #394 it runs about two or three attempts, putting the per-suite-run exposure in the low single-digit percent. #372 was filed on one sighting.

Three things make it worse than that number alone:

1. **It throws inside the loop.** It does not consume one of the twelve attempts and get recorded as a miss; it ends the test. The repetition this commit adds cannot absorb it, which inverts the commit's own argument.
2. **The message is a confident false diagnosis.** "the reap it removes is no longer what keeps the group clean" sends the next reader to `run-daily-fetch.sh`, on a machine that was merely busy. That is the exact #372 failure mode the first commit on this branch exists to delete, re-created one control over.
3. **The likely losing shape is not an empty pgid.** The orphan holds the inherited write end of the log pipe, so `tee` outlives it by a moment. The realistic loss is `residueAtExit == ["<pid> tee ..."]`, which prints `orphaned NOTHING into the pgid at the instant of exit (1 process(es) there)`, a message that contradicts itself inside its own parenthesis.

The motive is right and I would not drop the assertion: without it, a wrapper that stopped orphaning fails with "expired every time", which is the opposite misdiagnosis. But the two losing shapes are distinguishable, and separating them keeps every property the comment claims:

- `residueAtExit` non-empty with no `sleep` in it: the mutation stopped biting. Unambiguous, so fail on attempt 1 exactly as promised.
- `residueAtExit` empty: ambiguous, because the phase can produce it. Record it as an attempt outcome distinct from `expired`, and let the loop-level assertion fail only if all twelve came back that way, naming that shape in its message.

For grounding: the file's two existing `residueAtExit` assertions are safe precisely because their subjects cannot expire on their own. `:1237` asserts the watchdog is still serving its grace, and `:1386` asserts a child that ignores TERM is still there. This is the first use whose subject is a self-expiring timer, and that is what breaks the idiom.

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
```diff
@@ -2331,10 +2398,17 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
           // test, never as a wedged worker.
           maxWaitMs: 60_000,
           duringRun: async (pgid) => {
-            reaches.push(await waitForFakeToReach(dirs.caseDir, "prices:fetch", label));
+            reaches.push(await waitForFakeToReach(dirs.caseDir, "prices:fetch", label, dirs.logDir));
```
AmetAlvirde: **The wall-clock tripwire moved onto `:2437-2442`. It did not disappear, and the assertion it moved onto is measured from a different clock than the ceiling it names.**

`durationMs` starts at `launcher.testkit.ts:179`, immediately before `spawn`. The wrapper's ceiling starts at `SECONDS=0` inside the watchdog subshell, which forks well into the script, after the log dir, the `tee`, the stamp, the prior-heartbeat read and the operator notice. So `durationMs` is strictly larger than the elapsed time the watchdog is measuring, by the whole arming offset, and `expect(record.durationMs).toBeLessThan(ceilingMs)` is strictly tighter than the wrapper's own ceiling.

That matters now in a way it did not before this commit. There is a window where the reach is large enough that `durationMs` crosses 3000ms while the card has not appeared yet, because the card lands at `arm + 3000` on the harness's clock. In that window the wait returns "reached", the post-wait card check passes, the signal goes out, the run exits, and `:2442` reds with

> the run outlived its own 3000ms ceiling, so the ending it recorded may be the watchdog's rather than the harness's

while two assertions in the same run say the opposite: `record.watchdogFired` is false and the log carries no `WATCHDOG: run exceeded`. That message is false on its face in the only scenario that produces it, and it is a wall-clock bound going red on a busy machine. It is the #372 shape. Before this commit the window was unreachable, because the 2000ms wait failed first.

So the honest description of the case-4 change is: the effective bound on the reach went from a flat 2000ms to roughly the 3000ms ceiling minus the signal-to-exit tail, and the failure message on the path that now binds is worse than the one that was retired. That is still a real improvement, roughly 800ms more headroom against a measured contended worst of 1541ms, and the card race is the right mechanism. But the PR body's "a slow machine now costs this case latency and never a red", repeated at `:750-752`, is false in two ways: the card check reds at the ceiling, and `:2442` reds slightly before it.

Two things would settle it, either one:
- state in the comment that `durationMs < ceilingMs` is now the binding bound for this case, and that it is measured from spawn rather than from arming, so it is deliberately conservative;
- or give `:2442` a message that does not assert the watchdog fired when the assertions beside it prove it did not.

Related and smaller: the report line at `:2372` prints the reach against 3000 with headroom to spare, while the duration on the same line is racing the same 3000 and contains the reach. Reading either number as free headroom double-counts. The reach's own deadline is also `arm + 3000` rather than 3000, so printing it against the bare ceiling is conservative in the other direction. Both are defensible; neither is stated.

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
```diff
@@ -2533,33 +2607,99 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
...
+          // AND THE HALF ASSERTION 3 ACTUALLY TAKES: the same field, after the same window.
+          if (record.pgidResidue.length > 0) {
+            expect(
+              record.pgidResidue.join(" | "),
+              `attempt ${attempt}: something outlived the settle window, but it is not the ` +
+                "watchdog's orphaned timer",
+            ).toMatch(/sleep/);
+            seen += 1;
+          }
+          attempts.push(
```
AmetAlvirde: `attempts.push` runs after both assertions above it, so the attempt that throws is the one attempt missing from the log.

The comment at `:2636-2638` says the series "is the whole diagnosis if this control ever fails", that `"expired on all 12"` and `"the mutation orphaned nothing on attempt 1"` are different problems "and only the series tells them apart", and that it "must survive a throw in the loop". The hook does survive the throw, but on an attempt-1 failure it prints `child-reap mutation outcomes: ` with nothing after it, and the `run=` and `settle=` numbers for the only attempt anyone wants to see are gone. The comment at `:2626-2627` makes the stronger claim outright: "every attempt is on the record either way".

Moving the `push` above the `exitCode` assertion fixes it, with the outcome string computed from whatever is known at that point.

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
```diff
@@ -695,35 +704,81 @@ function watchdogCardPresent(logDir: string): boolean {
...
+ *
+ * So the wait no longer races a clock at all. It races {@link watchdogCardPresent}, which is
+ * the thing case 4's premise is ACTUALLY about, and this number is what remains: a bound on
+ * a run that is neither reaching its step nor being timed out by its own watchdog, which is
+ * a broken wrapper rather than a busy machine. It sits above ceiling + grace so the card
+ * always wins the race on a merely-slow run, and any run that reaches it has outlived the
+ * watchdog that was supposed to bound it.
  */
-const REACH_STEP_BOUND_MS = 2_000;
+const REACH_STEP_BACKSTOP_MS =
```
AmetAlvirde: The `+ watchdogGraceSeconds` term is inert, and the comment at `:723-728` names it as the reason the derivation works.

The watchdog touches the calling card at `run-daily-fetch.sh:849`, immediately **before** `kill -TERM` and therefore before the grace `sleep` at `:851`. So the event this backstop has to sit above is the ceiling, at 3000ms, not ceiling plus grace at 5000ms. The stated justification, "It sits above ceiling + grace so the card always wins the race on a merely-slow run", describes a card that lands after the grace, which is not what the wrapper does.

Harmless in effect: the extra 2000ms is padding on a backstop that only fires when the wrapper never armed at all, and padding there costs a slower red on an already-broken wrapper. Compare `:1148`, where the same `(ceiling + grace) * 1000` expression is correct, because case 7 is observing a run alive past the escalation. Same arithmetic, different reason, and only one of the two reasons is true here.

If the intent was "above everything the watchdog does", say that. If it was "above the card", the grace term should go.

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
```diff
@@ -2533,33 +2607,99 @@ describe.runIf(decision.run)("wrapper harness — the real wrapper, armed", () =
...
+        // seconds left, which is why a single reading looked solid for months. Under load the
+        // run's own duration wanders across the cycle, and a run that finished near a hop
+        // boundary leaves a stray with milliseconds to live, gone before the settle window
+        // closes. Measured at 3 contended runs in 5 when #394 was filed, and 3 in 6 again on
+        // the machine this repetition was added on.
+        //
+        // There is nothing to widen here: the phase is uniform, so a longer window makes the
+        // control blinder and a shorter one makes assertion 3's own reading unrepresentative.
+        // What DOES collapse it is asking more than once. At the measured miss rate a single
+        // reading is a coin flip; twelve independent ones miss together about four times in a
```
AmetAlvirde: Two problems with this paragraph, in a file whose style is to derive its numbers.

**The arithmetic.** At the 56% miss rate the PR body measured (14 expiries in 25 attempts), twelve misses together is `0.56^12 = 9.5e-4`, about one in a thousand. Four in a thousand needs a per-attempt miss rate of 63%. At the 3-in-6 rate the comment cites two paragraphs up it is 2.4 in ten thousand. The stated figure errs toward caution, so nothing downstream breaks, but it does not follow from either number in the same comment.

**Independence.** "twelve independent ones" is asserted, and the mechanism argues against it. The comment itself establishes that the miss is a deterministic function of where the run's duration falls in the 5s cycle, and run duration is set by ambient load, which persists across the roughly one minute the loop takes. A load regime that parks durations in the part of the cycle leaving under 2000ms of orphan makes all twelve attempts miss together, not `p^12`. The PR's own measurement (worst case 4 attempts of 12) is consistent with independence and does not establish it; a dozen loop invocations under one load regime is not a sample of regimes.

That does not change the conclusion that repeating is the right answer. It changes what can be claimed for it: the fix collapses the failure rate under the regimes measured, and the residual is bounded by how correlated the phase is under a regime nobody has measured. Worth saying, given this file says everything else.

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
AmetAlvirde: Taken, in `3a69835`. The assertion is gone and each attempt is classified into four outcomes instead: `SEEN`, `expired`, `gone-at-sample`, `NO-ORPHAN`. Only `NO-ORPHAN` — `residueAtExit` non-empty with no `sleep` in it — fails fast, and its message now says the pgid was occupied by something that is not the timer rather than that nothing was orphaned, so the self-contradicting parenthesis you found is gone too. `gone-at-sample` costs an attempt and is counted; the loop-level message reports how many attempts came back that way and what twelve of them would mean. The comment no longer claims anything is race-free and instead spells out where `residueAtExit` is actually sampled and why an empty reading proves nothing.

Both new outcomes were proven to fire rather than assumed reachable, with throwaway probes reverted afterwards. Pointing the loop at the unmutated wrapper spent all twelve attempts as `gone-at-sample` and failed with the loop-level message naming that count, no fail-fast. Synthesizing a pgid holding a non-`sleep` process failed on attempt 1 with the new message. Under 10 contended repetitions of the real control (continuously restarted `pnpm test` alongside) it went 10/10 across 21 attempts; no attempt came back `gone-at-sample` in that session, so the load run does not prove that path — the probe does.

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
AmetAlvirde: Both, in `dd575dc`. The assertion stays as it is; the message no longer asserts the watchdog. It now reports what was measured — `the run took Nms from spawn to exit, past the 3000ms ceiling` — states that it includes the arming offset the watchdog does not measure, points out that the assertions above already establish the watchdog did not fire, and sends the reader to the report line for which of the two it was. The comment above it says this is the binding wall-clock bound of the case now that the strip is gone, where each clock starts, and that being tighter than the wrapper is deliberate.

The `never a red` claim is deleted from both the `waitForFakeToReach` comment and the PR body, replaced with the honest version: the tightest bound went from 2000 ms to roughly the ceiling minus the signal-to-exit tail, and the far-end red now names the watchdog. Your smaller point about the report line is stated too: the duration contains the reach, the two are printed against the same ceiling from two different zeros, and neither is spare budget.

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
AmetAlvirde: Fixed the ordering, in `3a69835`. `attempts.push` now runs immediately after `launchWrapper` returns, before the `exitCode` assertion, with the outcome computed from the completed record. The comment says why the push has to sit there. Proven with a probe that failed on attempt 1: the series printed `NO-ORPHAN(run=604ms, settle=2118ms)` alongside the throw, where the old ordering would have printed nothing.

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
AmetAlvirde: Kept the grace term and rewrote the derivation, in `dd575dc`. You are right that the card lands at the ceiling, before the TERM. The reason the grace term still belongs is a different one: the card write is best-effort, `: > "$WATCHDOG_FIRED_FILE" 2>/dev/null || true`, so on a failed write the wait never sees a card and the last event the watchdog still guarantees is the SIGKILL at ceiling + grace. The comment now says that, names the line the card is written on relative to the TERM, and points at `timeoutSettleDeadlineMs` and case 7 as the place the same arithmetic is derived from the escalation itself.

apps/price-feed/src/wrapper-harness/run-daily-fetch.test.ts
AmetAlvirde: Both corrected, in `3a69835`. The number is now `0.56^12`, about one in a thousand, with the old four-in-a-thousand figure explicitly retracted in the comment (it needs a 63% miss rate) and the 3-in-6 rate quoted at 2.4 in ten thousand.

Independence is dropped rather than defended. The comment now says `p^12` is a floor, not a rate, gives your mechanism for why — the phase follows run duration, duration follows ambient load, and load persists across the minute the loop takes, so a bad regime makes all twelve miss together — and states that the measurement behind this branch is consistent with independence without establishing it. What the repetition buys is stated as collapsing the failure rate under the regimes measured, with the residual bounded by an unsampled correlation.

AmetAlvirde:
Findings 6 and 7 from the summary body, decided rather than fixed.

**6, `RUNS_PER_CASE`'s two meanings.** Agreed, and written down rather than renamed — `3a69835` adds a paragraph to `repetition.testkit.ts` naming the inversion: every ordinary case reads the count as "repeat N times" and a soak makes it harder to pass, while the two mutation controls read it as "attempt up to N and stop at the first observed failure" and a soak makes them *easier*. `NUMISMA_WRAPPER_TEST_RUNS=50` soaks twelve cases and simultaneously relaxes two controls.

I did not split the knob. The floor is the right floor for both readings, "how many attempts may be spent looking" is a real bound and not an accident, and a second env var would have to be documented, validated and refused-below-floor for a problem that is entirely one of documentation. The note tells whoever runs a soak to discount those two greens. If a soak ever becomes routine, split it then, with the reason already recorded.

**7, the 600 s timeout under a 12 × 60 s budget.** Left alone. It is pre-existing, identical in shape to the `tee`-deafness control, and changing it here would put a number in this file that disagrees with every other case for reasons this PR did not measure. The failure mode you describe is real — a systemic wedge reports as a bare vitest timeout rather than the launcher's "exceeded the harness cap" message — but it is one message worse on a path that is already a hard red, and fixing it properly means deciding the house number for all of them. Worth its own issue, not a change smuggled into a review response.

PR #395, opened 2026-08-20
fix(tui): give an unanswered question a value no operator can type

### LINKED ISSUE #388: fix(tui): make an unanswered question distinguishable from an empty one
## What is wrong

`""` is overloaded in the TUI prompt channel. It means both **"the operator
answered nothing"** and **"the operator was never able to answer"** (Ctrl-D,
no terminal). No question can tell those apart from the string it receives.

#370 fixed the *symptom* — a rejected `rl.question` no longer leaks a readline
internal to the operator — and #386 stopped the *worst consequence*: the import
flow now refuses at the write door when `channel.aborted()` is true, so an
abandoned interview cannot write joins nobody declared.

But the underlying overload is untouched, and the guard is positional: it lives
in `apps/tui/src/import-orders.ts`, not in the questions. Any new flow that asks
ratifying questions and then writes inherits none of it.

## Why this is filed rather than deferred again

Three increments have now named this and walked away, and both construction
sites (`import-orders-cli.ts`, `record-fill-cli.ts`) say in their own docstrings
that the sentinel is what they actually want. The deferral has been silent each
time, so the next agent to read those docstrings re-proposes it from scratch.

This issue exists so the deferral is *recorded*. Closing it as `wontfix` is a
perfectly good outcome — but then the docstrings must say "deliberately not
built, see this issue" so the fourth proposal doesn't happen.

## The shape of the fix

Make an unanswered question return something no operator can type, and make the
questions — not the flows downstream of them — reject it.

- `ask` resolves a sentinel (or a discriminated result) instead of `""` when the
  question was abandoned or there was no terminal.
- Every ratifying question rejects that value in its own voice.
- The write-door predicate in `import-orders.ts` becomes redundant and can go.

## The constraint that killed the first attempt

`d5fe02c`'s review proved that a naive sentinel is a wider change than it looks:
**every** ratifying question learns to reject a magic string, which is three-plus
call sites and a new forever-invariant. That is why #386 shipped the latch
instead. Any implementation here has to carry that cost deliberately, not
discover it halfway.

## Must not regress

`apps/tui/src/import-orders.test.ts` — *"still lets a Ctrl-D at the FIRST
question refuse as no-reserve-declared"*. That case pins #370's actual fix: a
Ctrl-D on question one must still let the domain refuse in its own voice rather
than print a readline internal.

## Context

- `apps/tui/src/prompt-channel.ts` — clause 4 and the `aborted()` latch
- Landed in PR #386 (`d5fe02c`); see also #370

AmetAlvirde:
Closes #388.

`""` was overloaded in the TUI prompt channel: it meant both "the operator
answered nothing" and "the operator was never able to answer" (Ctrl-D, or no
terminal on stdin). No question could tell those apart from the string it
received, so a Ctrl-D mid-interview raced through whatever was left of the
interview, ratifying every default on the way.

`ask` now resolves `string | typeof UNANSWERED`, and the questions refuse it.

## The three rulings this carries

**The sentinel is a symbol in a union, not a magic string.** A reserved string
flows through `isAffirmative("\0abandoned")` and is silently ratified, which is
the same defect one layer down plus an invariant humans have to remember
forever. `Answer = string | typeof UNANSWERED` makes every call site a type
error instead: `.trim()` does not exist on `string | symbol`, and a helper typed
`(answer: string)` will not take one. The compiler enumerated all 24 sites once
and enforces them from here. It is not a discriminated record either, because a
record breaks all nine test harnesses for no extra safety, while a fake `ask`
typed `Promise<string>` stays assignable under a union.

**The 15 questions that already refused a blank keep their reason codes.**
`no-reserve-declared`, `unknown-rung`, `bad-timestamp`, `incomplete-decision`,
`ambiguous-tier`, `bad-quantity`, `unknown-instrument`, `abandoned`. They fire
on `UNANSWERED` as well, through the same arm and the same words. The ten
default and ratifying sites are the behaviour change, and they refuse through
their flow's existing vocabulary rather than ten new reason codes.

**The write-door latch stays, demoted.** `PromptChannel.aborted`,
`OrdersImportIo.promptAbandoned` and the `interview-abandoned` check before
`appendOrders` are redundant now that the questions refuse for themselves. They
are kept one increment longer so a regression in the new per-question refusals
meets a guarantee that is still proven rather than a gap. Removing them is its
own change.

## What an operator sees differently

| question | was | now |
| --- | --- | --- |
| import: `Override the funding reserve per order? [y/N]` | declined the pass, the batch answer stood for every rung | `interview-abandoned`, naming the question |
| import: per-order `[batch]` | took the batch reserve for that rung | `interview-abandoned`, naming the rung |
| import: `Accept all? [Y/n]` | accepted every proposed rung join | `interview-abandoned` |
| import: `pick [hint]` loop | kept the proposal, and re-asked a terminal that cannot answer | `interview-abandoned`, no re-ask |
| fill: `Filled quantity [n]` | recorded the rung's whole remainder | `abandoned` |
| fill: per-rung `[r]/[t]/[g]` | claimed the rung resting untouched, a positive claim monotonicity reasons over | `abandoned` |
| fill: `Also record N cancellation(s)?` | recorded none and continued the act | `abandoned` |
| fill: `Cash debited [n]` | debited the proposed figure | `abandoned` |
| fill: `Append this lot to '<id>'? [Y/n]` | attached the lot, since silence meant yes | `abandoned` |
| fill: `Tempo [n]` | took the reserve's tempo | `abandoned` |

**Read the fill rows as intermediate state, not as durable writes.** Every
writer in `record-fill.ts` sits behind `Write BOTH? [y/N]`, on which
`isAffirmative("")` was already false, so an abandoned fill act reached that
gate and abandoned there with nothing on disk. What changes is where the refusal
lands: on the question the operator actually abandoned, rather than six answers
later at a gate that has by then been handed a quantity, a tempo, a cash figure
and a Position nobody named. The import flow is the one that could reach a
durable write, and it was the write-door latch that stopped it.

The two `[y/N]` confirmations, `Confirm these derived verdicts?` and
`Write BOTH?`, reach the same outcome as before. Only the message changed.

## Verification

`pnpm verify` green at the tip: 163 test files passed, 3 skipped; 2634 tests
passed, 20 skipped; `smoke:startup` exit 0. The TUI package went from 610 to 630
tests.

The must-not-regress case is green verbatim:
`import-orders.test.ts` `"still lets a Ctrl-D at the FIRST question refuse as
no-reserve-declared"`, which pins #370's actual fix. Both subprocess shell tests
that assert the no-terminal path end to end are unchanged in behaviour.

## Two things a reviewer should look at first

`record-fill.ts` now has both `typedOrNothing` and `abandon`. The whole second
ruling lives in which of the two each question calls, and getting one wrong is
silent. `typedOrNothing` reads the sentinel as `""` at exactly the questions
that already refuse a blank, which is what keeps the pinned refusal messages
identical.

`declareFunding` and `declareRungPicks` grew discriminated returns
(`{declared|not-declared|abandoned}` and `{picked|abandoned}`). There was no
other channel for "abandoned" versus "declared nothing", and the record-fill
leaves already use that shape.

## Known gap, not a regression

`abandoned` prints nothing today, which is the existing contract (`Write BOTH?
n` is silent too). The eight newly-abandoning sites therefore end the run with
exit 1 and no message. Giving `abandoned` an operator-facing channel is its own
increment.

### DIFF
--- a/apps/tui/src/import-orders.ts
+++ b/apps/tui/src/import-orders.ts
@@ -57,6 +57,7 @@ import {
 } from "./import-orders-report.js";
 import { renderUnattributedRefusal } from "./import-orders-unattributed-refusal.js";
 import { plural } from "./plural.js";
+import type { Answer } from "./prompt-channel.js";
 import { renderSkipMessage } from "./skip-message.js";
 
 /** Everything this flow touches that is not a pure function, in one injectable bag. */
@@ -108,27 +109,32 @@ export interface OrdersImportIo {
    */
   plansPath: string;
   loadPlans: (path: string) => Promise<LoadedPlans>;
-  /** Ask the operator one question; the answer is returned trimmed by the caller. */
-  ask: (question: string) => Promise<string>;
+  /**
+   * Ask the operator one question. Resolves what they typed, or `UNANSWERED` when there
+   * was no terminal to ask or the question was abandoned — see `prompt-channel.ts`. Every
+   * question in this flow narrows before it reads the answer as text, and every one of
+   * them refuses the sentinel: the batch reserve and the per-order override in
+   * `declareFunding`, both prompts in `declareRungPicks`.
+   */
+  ask: (question: string) => Promise<Answer>;
   /**
    * WHETHER THE TERMINAL CONDUCTING THIS INTERVIEW WAS ABANDONED — optional, and absent
    * means "it was not". Every caller that has no terminal to abandon (every fixture, and
    * any future non-interactive driver) is unaffected by omitting it.
    *
-   * WHY THE FLOW NEEDS A SECOND CHANNEL AT ALL. `ask` can only answer with a string, and
-   * the shell's prompt channel resolves `""` when a question is abandoned by Ctrl-D — it
-   * must, because that is what lets the FIRST question's blank come back as this flow's
-   * own `no-reserve-declared` refusal instead of a readline internal on the operator's
-   * screen (#370). But `""` is a RATIFICATION at three of this flow's questions: the
-   * funding override, and the rung-pick prompt at both of its sites. A Ctrl-D pressed
-   * after the funding declaration therefore used to race through them and land
-   * `planId`/`rungId` joins nobody declared, in an append-only file. The string cannot
-   * carry the difference, so the fact travels beside it.
+   * IT IS THE SECOND DOOR NOW (#388). It was the only one: `ask` used to answer `""` for
+   * an abandoned question, `""` is a RATIFICATION at three of this flow's questions — the
+   * funding override and the rung-pick prompt at both of its sites — so a Ctrl-D pressed
+   * after the funding declaration raced through them and landed `planId`/`rungId` joins
+   * nobody declared. The string could not carry the difference, so the fact travelled
+   * beside it. `ask` now answers `UNANSWERED`, which those three questions refuse where
+   * they stand, and the refusal names the question the operator walked away from rather
+   * than arriving at the door.
    *
-   * CHECKED AT THE WRITE DOOR, WHICH IS WHY THE GUARANTEE IS POSITIONAL-INDEPENDENT: if
-   * the terminal was abandoned, nothing is written — regardless of which question caught
-   * the Ctrl-D, including the last one. A latch that only changed later ANSWERS would
-   * still leak the interview whose final question was the abandoned one.
+   * WHAT THE LATCH STILL BUYS is a guarantee that holds without depending on any one
+   * question being written correctly: if the terminal was abandoned, nothing is written —
+   * regardless of which question caught the Ctrl-D. It is kept for exactly that reason
+   * while the per-question refusals are new. Removing it is its own change (#388).
    */
   promptAbandoned?: () => boolean;
   out: (message: string) => void;
@@ -143,10 +149,18 @@ export type OrdersImportRejection =
   | "unreadable-sidecar-lines"
   | "no-reserve-declared"
   /**
-   * THE OPERATOR ABANDONED THE INTERVIEW AT THE TERMINAL — Ctrl-D — and the answers
-   * behind that point are blanks nobody typed. See {@link OrdersImportIo.promptAbandoned}
-   * for why the flow cannot see this in the answers themselves, and `prompt-channel.ts`
-   * for the latch that carries it.
+   * THE OPERATOR ABANDONED THE INTERVIEW AT THE TERMINAL — Ctrl-D — or there was no
+   * terminal to conduct it on, which is the same fact to every question here.
+   *
+   * RAISED AT THE QUESTION THAT WENT UNANSWERED, and the message names it: `declareFunding`
+   * and `declareRungPicks` each return an `abandoned` arm carrying the question, and this
+   * flow turns either into this one reason. ONE reason rather than one per question, on
+   * purpose — the operator's next move is identical whichever prompt they walked away from
+   * (run the import again), and a reason token exists to distinguish next moves.
+   *
+   * ALSO RAISED AT THE WRITE DOOR, from {@link OrdersImportIo.promptAbandoned}. That check
+   * is redundant with the per-question refusals and is kept one increment longer as a
+   * belt on a guarantee that does not depend on any single question (#388).
    */
   | "interview-abandoned"
   /**
@@ -309,6 +323,25 @@ function reject(
   // be able to mistake a refusal for a quiet no-op.
   io.err(`REFUSED — ${message}\nNothing was written to ${io.ordersPath}.`);
   return { status: "rejected", reason, message };
+}
+
+/**
+ * The refusal an unanswered question earns, wherever in the interview it was put.
+ *
+ * ONE REASON, MANY QUESTIONS. The token is the flow's existing `interview-abandoned`; the
+ * MESSAGE names the prompt, because that is the half the operator needs and the half a
+ * reason code is bad at carrying. Nine per-question tokens would inflate the union without
+ * changing anybody's next move, which is to run the import again.
+ */
+function rejectAbandonedInterview(io: OrdersImportIo, question: string): OrdersImportOutcome {
+  return reject(
+    io,
+    "interview-abandoned",
+    `the interview stopped at ${question} — either the terminal was abandoned (Ctrl-D) or ` +
+      `there was no terminal to conduct it on, so that question was never answered. This ` +
+      `import attributes every rung to a declaration the operator made, and there is no ` +
+      `such declaration to attribute to. Nothing is written. Run the import again`,
+  );
 }
 
 /**
@@ -612,8 +645,13 @@ export async function importBitgetOpenOrders(
     // `io.ask` ALONE, not the bag: that module reads the prompt channel and nothing else,
     // and its signature says so — see its header.
     const declaration = await declareFunding(io.ask, admitted);
-    if (declaration === undefined) {
+    if (declaration.status === "not-declared") {
       return reject(io, "no-reserve-declared", "no funding reserve was declared for this batch");
+    }
+    // THE REFUSAL LANDS ON THE QUESTION THAT WAS WALKED AWAY FROM (#388), not thirty lines
+    // later at the write door. Nothing has been built, let alone written.
+    if (declaration.status === "abandoned") {
+      return rejectAbandonedInterview(io, declaration.question);
     }
 
     // THE DECLARED RUNG JOIN (#286), prompted after the funding declaration and before
@@ -624,13 +662,22 @@ export async function importBitgetOpenOrders(
     ladders = await loadInForceLadders(io, observedAt.slice(0, 10) as IsoDate);
     // `io.ask` ALONE again: the pick-list reads the prompt channel and nothing else —
     // the book on file arrives as VALUES, read here where the file already is.
-    const rungPicks = await declareRungPicks(
+    const picked = await declareRungPicks(
       io.ask,
       admitted,
       ladders,
       declaredJoinsOnFile(existingRecords),
     );
-    records.push(...buildOrderPlacedRecords(admitted, { ...declaration, rungPicks }));
+    if (picked.status === "abandoned") {
+      return rejectAbandonedInterview(io, picked.question);
+    }
+    records.push(
+      ...buildOrderPlacedRecords(admitted, {
+        fundingReserveId: declaration.fundingReserveId,
+        overrides: declaration.overrides,
+        rungPicks: picked.picks,
+      }),
+    );
 
     // `O1`. Coverage is checked over the WHOLE resting book — what is already on file plus
     // this batch — because a reserve funds every claim against it, not one import's slice.
--- a/apps/tui/src/import-orders.ts
+++ b/apps/tui/src/import-orders.ts
@@ -330,7 +330,7 @@ function reject(
  *
  * ONE REASON, MANY QUESTIONS. The token is the flow's existing `interview-abandoned`; the
  * MESSAGE names the prompt, because that is the half the operator needs and the half a
- * reason code is bad at carrying. Nine per-question tokens would inflate the union without
+ * reason code is bad at carrying. Five per-question tokens would inflate the union without
  * changing anybody's next move, which is to run the import again.
  */
 function rejectAbandonedInterview(io: OrdersImportIo, question: string): OrdersImportOutcome {
@@ -742,11 +742,17 @@ export async function importBitgetOpenOrders(
   const fresh: OrderRecord[] = [...records, ...observations].filter(
     (record) => !currentOnFile.has(appendKey(record)),
   );
-  // THE WRITE DOOR, AND THE LAST THING CHECKED BEFORE IT. An abandoned terminal makes
-  // every answer after the abandonment a blank nobody typed, and three of this flow's
-  // questions read a blank as consent — so the refusal cannot live at any one question,
-  // it has to live here, where every path that writes passes. Positional-independent: it
-  // does not matter which question caught the Ctrl-D, including the last one.
+  // THE WRITE DOOR, AND THE SECOND DOOR ONLY (#388). It is no longer what stops an
+  // abandoned interview writing: `ask` resolves `UNANSWERED`, every question of this flow
+  // refuses it where it stands, and the refusal names the question the operator walked
+  // away from. No blank is involved and no question reads one as consent.
+  //
+  // WHAT IT STILL BUYS is a guarantee that does not depend on any one question being
+  // written correctly: if the terminal was abandoned, nothing is written, whichever
+  // question caught the Ctrl-D. A regression in the new per-question refusals meets this
+  // rather than a gap. In production it is unreachable — reaching it means an `ask`
+  // returned the sentinel, and every path that could have is already gone — which is
+  // exactly what a second door should look like. Removing it is its own change.
   //
   // A REFUSAL, NOT A THROW, and in the established vocabulary — the operator gets the
   // same `REFUSED — ... Nothing was written` contract as `write-failed`, and the second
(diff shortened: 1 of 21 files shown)


apps/tui/src/record-fill.ts
```diff
@@ -209,6 +221,44 @@ function reject(
...
+ * record the fill again. The MESSAGE names the question — the half a reason token carries
+ * badly — which is why nine per-question tokens would have bought nothing.
+ *
+ * `isAffirmative` KEEPS ITS `string` PARAMETER and every caller narrows before it. Widening
+ * the helper so `isAffirmative(UNANSWERED)` returned false would compile and would answer
+ * NO to `Write BOTH?` — right by luck at that gate, and wrong at `Append this lot?`, whose
+ * `[Y/n]` phrasing makes silence mean yes. A helper that decides what silence means is the
+ * defect #388 removes, one layer down.
+ */
+function abandon(question: string): RecordFillOutcome {
```
AmetAlvirde: **1. [medium] The eight new `abandoned` messages are never shown. A mid-interview Ctrl-D in `orders:fill` now exits 1 with zero output.**

`abandon()` builds a message that names the question, and unlike `reject()` twenty lines up it never touches `io.err`. `record-fill-cli.ts:155` only sets `process.exitCode = 1` for a non-`recorded` outcome. The no-terminal notice does not cover the gap either: it fires from `prompt-channel.ts` only when `isTTY` is false.

Scenario: `pnpm orders:fill` at a real terminal. Answer `Which rung filled?` and `Fill timestamp`, then press Ctrl-D at `Cash debited [4000]:`. `resolveFunding` returns `{status:"abandoned", message:"nobody answered the cash debited for this fill, and 4000 is a proposal, not an answer"}`, `record-fill.ts:914` returns it verbatim, the shell prints nothing and exits 1. The operator sees a bare prompt, a dropped shell and an exit code, with no way to tell whether the fill landed.

The PR body discloses the silence as an existing contract, and it is: `n` at `Write BOTH?` was silent before. What the diff changes is the size of the hole. Two silent sites became eight, and six of them gained a hand-authored, question-naming message that nothing can read. The reportable half is the drift: `prompt-channel.ts` clause 2 says "Two sentences, two layers: the shell names WHY there is no answer, the flow names WHAT IT DID about it", and on this path the operator gets zero sentences. `import-orders.ts` does not have the problem, because `rejectAbandonedInterview` goes through `reject`, which writes to `io.err`.

The one-line version is `err` in `abandon()`. Whether that belongs in this increment or the follow-up is your call, but clause 2's claim should not stand unqualified while it is false for `orders:fill`.

apps/tui/src/import-orders.ts
```diff
@@ -108,27 +109,32 @@ export interface OrdersImportIo {
...
-   * CHECKED AT THE WRITE DOOR, WHICH IS WHY THE GUARANTEE IS POSITIONAL-INDEPENDENT: if
-   * the terminal was abandoned, nothing is written — regardless of which question caught
-   * the Ctrl-D, including the last one. A latch that only changed later ANSWERS would
-   * still leak the interview whose final question was the abandoned one.
+   * WHAT THE LATCH STILL BUYS is a guarantee that holds without depending on any one
+   * question being written correctly: if the terminal was abandoned, nothing is written —
+   * regardless of which question caught the Ctrl-D. It is kept for exactly that reason
+   * while the per-question refusals are new. Removing it is its own change (#388).
    */
   promptAbandoned?: () => boolean;
```
AmetAlvirde: **2. [low] This docstring is now correct and the write door's own comment is not. They contradict each other in one file.**

This member's prose says "IT IS THE SECOND DOOR NOW (#388)". The comment at `import-orders.ts:745` still says the opposite:

> An abandoned terminal makes every answer after the abandonment a blank nobody typed, and three of this flow's questions read a blank as consent, so the refusal cannot live at any one question, it has to live here, where every path that writes passes.

Both halves are false as of this commit. The channel resolves `UNANSWERED`, not a blank, and all five import questions refuse it where they stand. The operator-facing message at `:757` repeats it: "the answers after that point were blanks nobody typed, and three of this import's questions read a blank as consent".

Reachability: I could not construct a production path to that message. `prompt.aborted()` is true only after an `rl.question` rejected, which means some `ask` returned `UNANSWERED`, and all five import questions return before the write door on it. So today only `import-orders.test.ts`'s stubbed `promptAbandoned` reaches it, and the test asserts the reason token rather than the words. That is precisely why ruling 3 keeps the latch: it is meant to stay honest for one more increment, and its comment and its message are the parts of it that went stale while everything around them was rewritten. Region `:745-763` is the only untouched block in a file that is otherwise +91.

apps/tui/src/record-fill-ladder-target.ts
```diff
@@ -118,11 +149,19 @@ export async function authorLadderTarget(
...
   // the fill, and nowhere else.
   out("First fill on this ladder — opening the Position.\n");
-  const positionId = (await ask("  Position id: ")).trim();
+  const idAnswer = await ask("  Position id: ");
+  const positionId = idAnswer === UNANSWERED ? "" : idAnswer.trim();
   if (!positionId) {
     return { status: "abandoned", message: "no position id was given" };
   }
-  const tempoAnswer = (await ask(`  Tempo [${reserve.tempo}]: `)).trim();
+  const tempoReply = await ask(`  Tempo [${reserve.tempo}]: `);
```
AmetAlvirde: **3. [low] `Tempo [n]` is a tenth behaviour-changed site, and it is absent from every summary.**

This question advertises `reserve.tempo` as its default, so before the change an unanswered one wrote the reserve's tempo onto a `PositionOpened` that carries it for the life of the Position. It abandons now, which is right.

It is documented in this module's header ("`Tempo [n]` gets the same treatment for the same reason") and covered by "abandons on an unanswered tempo rather than writing the reserve's own". `record-fill-cli.ts` counts it correctly: eleven that keep their refusal plus eight that changed equals nineteen.

What omits it is the narrative. The PR body's table has nine rows and no tempo row; the commit message says "nine of the twenty-four questions"; the body's note that "the two `[y/N]` confirmations reach the same outcome as before" accounts for the other unlisted sites but not this one. The old `record-fill-cli.ts` docstring said "SIX are silent ratifications" for the fill act and listed tempo among them, so the correct total is ten, not nine. Worth fixing in the body before merge, since a reader auditing the change against the table will not check this line.

apps/tui/src/record-fill-ladder-target.ts
```diff
@@ -118,11 +149,19 @@ export async function authorLadderTarget(
   // never heard of a Tempo, so the decision context is authored here, at the moment of
   // the fill, and nowhere else.
   out("First fill on this ladder — opening the Position.\n");
-  const positionId = (await ask("  Position id: ")).trim();
+  const idAnswer = await ask("  Position id: ");
+  const positionId = idAnswer === UNANSWERED ? "" : idAnswer.trim();
```
AmetAlvirde: **5. [low, cleanup] Possible Duplicated Code: this inlines the helper defined seventy lines above it in the same file.**

`askDecision` at `:83` defines `const text = (answer: Answer): string => (answer === UNANSWERED ? "" : answer.trim())`, and this line writes the same expression by hand. Hoisting `text` to module scope and calling it here is a two-line change with no argument attached to it.

The wider version is a judgment call and I am not asking for it in this PR: the same collapse now appears five times in four files, as `typedOrNothing` in `record-fill.ts:258` and `import-orders-funding-declaration.ts:70` (byte-identical), as `text` here, and inline at this line and `record-fill-funding.ts:166`. Unlike `isAffirmative`, whose duplication both leaf modules argue for on the grounds that a shared home is a decision rather than a move, this one already has a home: `prompt-channel.ts`, beside `UNANSWERED` and `Answer`. The counter-argument the code makes is real, that keeping it local is a guard rail because calling it at a defaulting question is the silent defect this PR fixes. That argument covers the cross-file copies. It does not cover two copies inside one file.

apps/tui/src/prompt-channel.test.ts
```diff
@@ -257,10 +258,13 @@ describe("on a terminal where the question is aborted the channel ends the ANSWE
   // the issue names: the domain's refusal, `REFUSED — no funding reserve was declared for
   // this batch`, was never reached. Resolving with `""` is the same answer the
```
AmetAlvirde: **6. [low] Stale comment. The case below it was updated, this sentence was not.**

> Resolving with `""` is the same answer the no-terminal path gives, and it lets the domain refuse in its own voice.

The case it introduces is now `RESOLVES with UNANSWERED when the question rejects with Node's measured Ctrl-D abort`, and its inline comment says the same thing correctly. The sentence above still names the value the channel no longer resolves. Same fix as the describe title on `:196` got.

apps/tui/src/skip-message-shells.test.ts
```diff
@@ -241,7 +244,9 @@ const SHELLS: Shell[] = [
             plans: [],
             skipped: [],
           }),
-          ask: async () => "",
+          // THIS FLOW MUST REFUSE BEFORE IT ASKS ANYTHING, so the honest answer is the one
+        // no question can use: `UNANSWERED` (#388).
```
AmetAlvirde: **7. [low, cleanup] Indentation break.**

The comment's second line and `ask:` sit at 8 spaces inside an object whose other members are at 10 (`plansPath`, `loadPlans`, `out`, `err` around them). The sibling edit at `:207-209` in the same file got it right, so this is a hand-formatting slip in the second of the two.

Worth calling out because the repo carries no prettier, biome or eslint config, so nothing will pick it up later.

AmetAlvirde (commented):
## Review of `37ee5a6`

Base `main`. Typecheck clean (`tsc --noEmit -p apps/tui/tsconfig.json`, exit 0) and the TUI suite green at the tip in a detached worktree: 45 files, 630 tests, 0 failures.

### What holds

**The three rulings are honoured.** The sentinel is a symbol in a union, and all 24 `ask` sites narrow before reading text: 3 in `import-orders-funding-declaration.ts`, 2 in `import-orders-rung-picks.ts`, 2 in `record-fill-funding.ts`, 8 in `record-fill-ladder-target.ts`, 9 in `record-fill.ts`. `isAffirmative` (three copies) and `isNegative` (one) all keep their `string` parameter, so the silent-ratification defect one layer down does not exist here.

**Every numeric parse still tests the blank before it parses.** `record-fill.ts:373` puts `rawQuantity === ""` first in the `||` chain, ahead of `Number.isFinite`; `record-fill.ts:681` and `record-fill-funding.ts:80` branch on `=== ""` before `Number(...)`; `record-fill.ts:645` gates on `/^\d+$/`; `import-orders-rung-picks.ts:210` handles `reply === ""` before `Number(reply) - 1`. No path pushes a narrowed-but-empty string into a silent 0.

**`typedOrNothing` vs `abandon`, checked one call site at a time.** All nine `record-fill.ts` sites are on the right helper. Collapsing to `""` at the rung pick (644), the timestamp (654), a touched rung's quantity (367) and the instrument id (879), which is where `unknown-rung`, `bad-timestamp`, `bad-quantity` and `unknown-instrument` already fired on a blank. Checking the sentinel directly at filled quantity (670), the per-rung book question (355), the verdict confirmation (807), the cancellation confirmation (818) and `Write BOTH?` (986). No defaulting question reaches for `typedOrNothing`.

**The discriminated returns are fully handled.** `import-orders.ts:648` `not-declared` to `no-reserve-declared`, `:653` and `:671` `abandoned` to `interview-abandoned`. "Declared nothing" and "abandoned" cannot be conflated: they come back as different arms and take different reason codes. `declareRungPicks` returning `picked` with an empty map when `choices.length === 0` asks nothing, so an empty map never stands in for an abandonment.

**The `for(;;)` cannot spin.** `import-orders-rung-picks.ts:204` returns `abandoned` on the sentinel instead of looping or falling into the blank's keep-the-proposal branch. Pinned by "abandons the override loop rather than keeping the proposal or spinning", which asserts exactly two questions.

**Must-not-regress is intact.** `import-orders.test.ts` "still lets a Ctrl-D at the FIRST question refuse as `no-reserve-declared`" passes, its driving answer changed from `""` to `UNANSWERED` (necessary, since that is what the shell now hands the flow) and its assertions unchanged. Both subprocess shells (`import-orders-cli.test.ts`, `record-fill-cli.test.ts`) are comment-only diffs and green. The clause-3 property survives: `prompt-channel.test.ts` still drives nine questions on a non-terminal and asserts `harness.errors` equals exactly one notice.

**Test fidelity, no tautologies.** The five flipped fakes all sit under cases that assert a positive outcome (`status: "recorded"` in `record-fill-reliable.test.ts:242` and `record-fill-reconcile.test.ts`, or a specific `rejected` reason). An accidental fall-through to `UNANSWERED` would produce `abandoned` and fail, so exhaustion is a loud failure rather than a silent default. The abandonment cases are driven the way the real channel behaves: one real answer, then the sentinel forever.

### Findings

Seven, most severe first. Six are inline. The seventh has no single location:

**4. The PR body's "before" column describes intermediate state, not outcomes: no abandoned fill act could ever have reached a write.** Every writer in `record-fill.ts` sits behind `Write BOTH? [y/N]` at line 986 (`writeLogImage` 997, `appendOrders` 1005, `reconcileRecordedFill` 1046). An abort closes the interface, so every question after it rejected too, `isAffirmative("")` was false at that gate, and the act abandoned with both files untouched. That includes `Append this lot to '<id>'? [Y/n]`, which the PR body calls "the worst of the nine and the only one no latch ever covered". The lot was attached in memory; the act still refused four questions later and wrote nothing. `record-fill-cli.ts`'s own docstring states this correctly ("WHAT USED TO STOP THIS ACT WRITING WAS ITS FINAL GATE, AND THAT WAS A PROPERTY OF THE PHRASING"), so the code's prose and the PR's prose disagree. The change is still the right one, since it removes the dependency on one gate's phrasing, but the five fill-side rows read as if durable writes were happening and they were not. The import side is different: `#386`'s latch is what covered it, which the body does say.

Severity: 1 medium, 6 low.

AmetAlvirde:
Findings 3 and 4 are in the PR body rather than the code, so they are fixed here rather than in the branch.

**Finding 3, the undercount.** The table now carries the tenth site, `record-fill-ladder-target.ts`'s `Tempo [n]`, and the ruling text says ten. The commit message still says nine; it stays as written, because this repo does not rewrite history to fix prose.

**Finding 4, the overstated "before" column.** Correct, and the correction is now in the body. Every writer in `record-fill.ts` sits behind `Write BOTH? [y/N]`, so no abandoned fill act ever reached a durable write, including `Append this lot?`. The body no longer claims that gate was uncovered. What the change actually buys on the fill side is where the refusal lands: on the abandoned question, rather than six answers later at a gate already holding a quantity, a tempo, a cash figure and a Position nobody named. The import flow is the one that could reach a real write, and the latch is what stopped it.

apps/tui/src/record-fill.ts
AmetAlvirde: Fixed in f2b2e0c, and wider than the one-line version. Every `abandoned` exit now goes through a new `abandonWith(io, message)` that writes `reject()`'s exact contract — `REFUSED — <message>` plus `Nothing was written to <both paths>`. That includes the two pre-existing `[y/N]` sites where the operator typed a deliberate `n` (they already had message strings, so nothing was invented), and the two delegated flows: `authorLadderTarget` and `resolveFunding` return an abandonment message with no `io` to speak it, so `record-fill.ts` speaks it at the pass-through.

Tests: the `writes NOTHING` assertions are unchanged; each gained an assertion on the new line. Two cases added to the sentinel table at answers 7 (`Tempo [n]`) and 13 (`Cash debited [n]`) so the delegated pass-through is pinned end-to-end, not only in the leaf suites. `record-fill-cli.ts`'s header now states the exit code is the second half of the answer and never the whole of it.

apps/tui/src/record-fill-ladder-target.ts
AmetAlvirde: Fixed the in-code half in ba918f7 and 4ca9491; the commit message stays wrong by policy (this repo never rewrites history) and the repo owner is fixing the PR body.

The two in-code miscounts were `abandon()`'s "nine per-question tokens would have bought nothing", now ten, and `rejectAbandonedInterview`'s identical sentence in `import-orders.ts`, which said nine for an interview of five questions — now five. `record-fill-cli.ts`'s "nineteen `ask` sites, not the nine `record-fill.ts` holds literally" is left alone: that nine is a literal count of `io.ask` calls in the file and it is correct.

apps/tui/src/record-fill-ladder-target.ts
AmetAlvirde: Hoisted to module scope in ba918f7 and the position id calls it.

Not moved to `prompt-channel.ts`, and the new docstring says why: every question in this file that reaches for `text` already refuses a blank, while `Append this lot? [Y/n]` and `Tempo [n]` twenty lines below must not, so a shared home would put it within reach of the two questions it is unsound for. The cross-file `typedOrNothing` copies are left as they are.

PR #397, opened 2026-08-20
docs: reconcile the docs with the implementation since v0.8.0

AmetAlvirde:
Reconciles the documentation with the implementation across the 220 commits
since `v0.8.0` (9dee673, 2026-08-08), which is where the last sweep stopped.
39 files, six lanes, partitioned so no two lanes shared a file.

Every claim was re-verified against current source rather than against commit
messages. Where a doc cited `file:line`, the line was re-opened; dead citations
are retargeted or replaced by a named block.

## What was actually wrong

The drift that would have cost an operator something:

- **`docs/hosted-cutover-runbook.md` read `push` exit `0` as the success signal.**
  A good run can exit 1 with the row landed, because the discard channel owns
  that code. Step 8 and the soak condition both rested on it.
- **`docs/price-feed-ops.md` asserted in three places that a lost day is
  unrecoverable.** True before `--as-of` shipped in #373, false since. The flag
  itself was undocumented: the price-feed README still said "No argv flags".
- **`docs/durable-log-ops.md` listed five durable ledger files** against the
  seven `durable-log-guards.test.ts` already pins. `local-data.md`, ADR-006 and
  the ADR index carried the same undercount.
- **`docs/coverage-rationale.md` §1 claimed five CLI shells are untested by any
  suite.** Three have gained spawn suites. The honest count is two. #396
  re-derived §§2/4/5/6 and #387 handled §7, so §1 was never re-checked.

The rest is inventory catching up: `pnpm plans` had no entry-point row despite
shipping in #267, `ops/testkit/` was invisible in the codebase map, the web
README counted five routes against seven and three payload branches against
four, `domain-model.md` had no discard channel, no plans and no fill path, and
`product.md` still called the hosted surface hypothetical after it shipped.

## ADRs

Citation and dead-mechanism fixes only. No ruling is rewritten, and where a
decision has genuinely been overtaken it is marked in place and pointed at its
successor rather than edited. ADR-002 gains a dated amendment; ADR-011,
ADR-017 and ADR-021 verified clean and untouched.

## Deliberately not in this PR

- **`docs/coverage-rationale.md` §§2 through 7.** They were re-derived against a
  fresh coverage run within the last two days, and this sweep had no run behind
  it. Only §1's tree-falsifiable prose is touched; no percentage is restated.
- **Four code findings**, filed as issues rather than smuggled into a docs PR:
  the wrapper's step-3 backstop and strict post-check name five durable files
  while `TRACKED_FILES` names seven, so `reconciliations.jsonl` is never
  captured by the daily run and the post-check still reports clean; the plist
  hardcodes an operator path that `data-dir.ts` forbids by name; ADR-005's
  mark-instant consequence no longer describes `--as-of` recovery; and
  `invalidLotFields` accepts `entryFx` on a cash lot that ADR-002 says cannot
  carry one.

## Verification

Docs and comments, with one exception. Four of the five non-markdown files in
the diff (`plans-cli.ts`, `plans-report.ts`, `run-daily-fetch.sh`,
`vitest.config.ts`) are comment-only, confirmed by filtering the diff for
non-comment lines. No version string, no config value and no executable line
changed. No real fund data was added.

The exception is `apps/tui/src/plans-report.test.ts`. Its runbook guard banned
the substring `stderr` anywhere in `docs/plans-authoring-runbook.md`, so the new
paragraph naming the fold's discard lines turned it red. The doc is the correct
half: `plans-cli.ts:65-67` writes those lines to `process.stderr` outside
`report.exitCode`, and the paragraph tells the operator exactly that. Making the
prose vaguer to dodge a substring match would trade real information for a green
check, so the assertion was narrowed instead. It now matches an instruction verb
bound to `stderr` inside one sentence, which is the property the guard's own doc
comment already claimed, and it was seen red against a probe line before it was
trusted.


AmetAlvirde (commented):
Docs review of the sweep, verified against `main` at 2d105ea (CalVer) merged with this branch.

I re-derived the load-bearing claims rather than the prose: every `file:line` citation rewritten here, the seven-file durable floor against `TRACKED_FILES`, the four spawn suites, the two driverless shells, the seven web routes, schema versions 4 through 6, the four `ProjectionReport` branches, `--as-of`'s parser and exit contract, the nine `NUMISMA_*` wrapper variables with 78/124/2700/30, `MAX_NOTICE_LOST_DAYS` 10, `MAX_WINDOW_DAYS` 400, 21 ADRs, 166 test files, the CI gating, and the `pnpm` script table. Those all hold.

Three findings below. Also confirmed: the four non-markdown files change comment lines only (no diff line survives stripping comment markers), no real fund figure is added anywhere in the diff, the two real NAV values are gone from the tree, every relative link in every changed markdown file resolves, and no doc states a stale version after the CalVer merge.

Not merging, not approving. Comments only.

README.md
```diff
@@ -28,11 +28,11 @@ six members are versioned together at the monorepo version.
...
+| [`@numisma/event-store`](./packages/event-store/README.md) | Node-compatible                            | The durable log's **read** path — path resolution, genesis load, log load with quarantine, `loadFoldedReview` — plus the gap-report, job-heartbeat, and operator-notice sidecars. Consumed by both the TUI and the web push. Exposes `.` and the `./testkit` subpath export.                                                                                 |
+| [`@numisma/preferences`](./packages/preferences/README.md) | Node-compatible                            | Sidecar file IO for `preferences.jsonl` (profit-split policy), `orders.jsonl` (resting claims), `plans.jsonl` (per-position DCA plans), and `reconciliations.jsonl` (the record of what a reader showed the operator). Append-only, validated on load, appended through one lock + temp + rename shell.                                                                                                                        |
```
AmetAlvirde: **`preferences.jsonl` does not go through that shell.**

`packages/preferences/src/sidecar-io.ts` is imported by exactly three modules: `orders.ts:34`, `plans.ts:62`, `reconciliations.ts:104`. The only preferences writer, `seedDefaultPreferences`, calls `mkdir` and then a bare `appendFile` (`packages/preferences/src/preferences.ts:416-417`). No lock, no temp sibling, no rename.

The line this replaces said "cross-process locked for orders", which was accurate and narrow, so this is a regression rather than a stale claim catching up.

The same sweep already says it correctly elsewhere. `docs/local-data.md`'s `*.jsonl.lock` row names `orders.jsonl`, `plans.jsonl` and `reconciliations.jsonl` only, and `packages/preferences/README.md` still says the preferences append "is inline; there is no general `appendPreference` entry point". Three places in one sweep, two of them wrong.

Consequence: someone adding a second preferences writer, or reasoning about two processes seeding the sidecar at once, reads this table and believes a cross-process lock protects the file. It does not, and `preferences.jsonl.lock` is never created on disk.

docs/codebase-map.md
```diff
@@ -27,11 +27,11 @@ placeholder convention itself, are documented in
...
+| `packages/engine` | `@numisma/engine` | Node, IO-free | The pure fund domain: parse → validate → fold → compose → format, plus the pure half of the price model, the whole Orders model, the pure halves of the plans and reconciliations sidecars, and the one data-dir predicate (`normalizeDataDirOverride`) every runtime plane routes through. No file, network, or terminal IO. | `src/index.ts` (curated barrel, no `export *`), `src/events/types.ts` |
+| `packages/event-store` | `@numisma/event-store` | Node | The durable log's **read** path — path resolution, genesis load, log load with quarantine, `loadFoldedReview` — plus the gap-report, heartbeat, and operator-notice sidecars. | `README.md`, then `src/` |
+| `packages/preferences` | `@numisma/preferences` | Node | Sidecar file IO for `preferences.jsonl`, `orders.jsonl`, `plans.jsonl`, and `reconciliations.jsonl`. Append-only, validated on load, appended through one cross-process lock + temp + rename shell (`src/sidecar-io.ts`). | `README.md` |
```
AmetAlvirde: Same defect as the `@numisma/preferences` row in the root `README.md`, and worse here because this one cites `src/sidecar-io.ts` by name.

`preferences.jsonl` never reaches that module. `sidecar-io.ts` is imported only by `orders.ts`, `plans.ts` and `reconciliations.ts`; `seedDefaultPreferences` writes with `mkdir` + `appendFile` (`packages/preferences/src/preferences.ts:416-417`).

A reader who follows the pointer to `sidecar-io.ts` looking for the preferences append finds four files claimed and three served, with nothing in that file explaining the gap.

docs/coverage-rationale.md
```diff
@@ -69,21 +69,26 @@ excluded because v8 cannot report a spawned subprocess's coverage back to this
...
+codes; they stay excluded only because v8 cannot see into the subprocess.
+
+TWO shells are genuinely driverless at HEAD: `plans-cli.ts` and
+`operator-notice-cli.ts`. That is the count to read off this section. Both joined
+this list by being excluded rather than by being tested, so this paragraph is the
+only thing saying so, and it has to say it precisely. Until they were excluded,
+both reported a dishonest 0%, and that 0% was itself the flag; an exclude trades
+a dishonest number for an honest accounting, and this is the accounting. Neither
+has a driver. Nothing in the tree imports or spawns `plans-cli.ts`: the filename
+appears in the `pnpm plans` script, in `apps/tui/README.md`, in the
```
AmetAlvirde: **The `plans-cli.ts` mention list is incomplete, and it is offered as a grep oracle.**

`vitest.config.ts` names the filename twice: once in the exclusion comment ("`plans-cli.ts` is a bare top-level `try/catch` that resolves...") and once as the `coverage.exclude` entry at `vitest.config.ts:61`. A `grep -rn plans-cli` over the tree returns five sites, not the four this sentence enumerates, so the paragraph two lines down ("A `grep` for either filename therefore returns hits and still returns no driver") sets the reader up to find a hit the list did not predict and wonder what it is.

The table row at line 46 carries the same list, and there the omission is self-contradictory inside one cell: three sentences earlier it says "Excluded as of the increment that added it to `vitest.config.ts`".

The conclusion is right. `plans-cli.ts` genuinely has no driver, and I confirmed nothing imports or spawns it. Only the enumeration is short.

context/adr/ADR-013-order-a-claim-on-capital-recorded-beside-the-log.md
```diff
@@ -98,13 +98,15 @@ the venue shows, and it stays in the sidecar until it becomes one of those.
...
-  any it cannot parse** (`:152`, and again over inbox candidates at `:178`).
-  launchd runs it daily from the working checkout. It **never reads
-  `orders.jsonl`** — verified by grep: the string `orders` does not appear in
-  that file at all. Adding lines to a file the daily check does not open cannot
-  break the daily check.
+  `apps/price-feed/src/rejection-check.ts` reads `data/events.jsonl` and
+  **throws on anything it cannot read** — at the time of this decision by
+  splitting the file and running `parseEvent` line by line; today through
+  `loadEventLog` + `assertLogFullyLoaded` (`:140-141`), with the same policy
+  applied again to inbox candidates by `walkPendingInbox` (`:164`, throwing at
```
AmetAlvirde: `apps/price-feed/src/rejection-check.ts:170` is `if (walk.invalid) {`. The `throw new Error(` is at `:173`.

The two citations beside it are exact (`:140-141` is the `loadEventLog` / `assertLogFullyLoaded` pair, `:164` is the `walkPendingInbox` call), and the rest of the bullet checks out: `grep -c orders` over that file returns 0, so the "verified by grep, then and now" claim holds.

Minor, but this PR's own standard is that a rewritten `file:line` was re-opened, and `:170` names a guard rather than the throw.

PR #401, opened 2026-08-21
fix(ops): stage the reconciliation trail in the daily wrapper

### LINKED ISSUE #398: the daily wrapper stages five durable files against a seven-file contract
`TRACKED_FILES` (`apps/tui/src/ingest-commit.ts:45`) names seven durable files.
The daily wrapper names five, in two places:

- `ops/price-feed/run-daily-fetch.sh:952` stages `events.jsonl genesis.json
  preferences.jsonl orders.jsonl plans.jsonl` explicitly, after a `git add -u`
  that only touches already-tracked paths.
- The strict post-check at `:1000` runs `git status --porcelain` over the same
  five.

`reconciliations.jsonl` is in neither list, and it is currently **untracked** in
accumulus. `git add -u` cannot pick up an untracked path, so the daily run never
captures it, and the post-check then prints `post-check OK: durable log
committed clean` over a durable file sitting uncommitted beside the log.

That is the exact failure ADR-006's amendment names: it says the four places
must be edited in order (accumulus allowlist, `TRACKED_FILES`, the wrapper's
explicit-add loop, the strict post-check) *because* the reverse order produces a
green check over a file git is discarding. The allowlist and `TRACKED_FILES`
halves landed; the two wrapper halves did not.

**Not data loss today.** The accumulus `.gitignore` does allowlist
`/data/reconciliations.jsonl`, so the file is not ignored, and the TUI's own
`ingestInbox` capture carries all seven. A fill that ingests through the TUI is
committed. The gap is the wrapper backstop, which exists precisely for the runs
where the in-process capture did not happen, and the false-clean post-check that
hides it.

Also at `:1034`, a comment still says "the allowlist versions six files".

### What a fix needs

A guard, not just an edit. `apps/tui/src/durable-log-guards.test.ts` already
asserts both ends over `TRACKED_FILES`; the wrapper's two lists are a third and
fourth end with nothing pinning them. Fixing the shell without pinning it to
`TRACKED_FILES` reproduces this issue the next time the list grows.

Found during the docs sweep in #397, which documents the seven-file contract but
deliberately changes no executable line.

AmetAlvirde:
`TRACKED_FILES` (`apps/tui/src/ingest-commit.ts:45`) names seven durable files. The
daily wrapper named five, in two hand-kept literals: the step-3 explicit-add loop and
the step-4 strict post-check. `reconciliations.jsonl` was in neither.

The consequence is the one ADR-006's amendment orders these edits to prevent. `git add
-u` restages tracked modifications only, so it cannot pick up an untracked path; the
trail is written by `orders:fill` and does not exist in accumulus yet, so the first fill
recorded without the in-process capture would leave it uncommitted — and step 4 would
then print `post-check OK: durable log committed clean` beside it. A green check over a
file git is discarding.

Nothing has been lost. The accumulus allowlist does version `/data/reconciliations.jsonl`,
the TUI's own `ingestInbox` capture carries all seven, and no fill has been recorded
since the trail landed. The gap was the wrapper backstop, which exists precisely for the
runs where the in-process capture did not happen.

## What changed

**One list, two consumers.** The two literals collapse into a single
`DURABLE_STRICT_FILES` array declared above step 3 — the seven files minus
`head-digest.json`, which stays in the lenient `--ignored` arm because `git add` of an
ignored path aborts the run under `set -e`. Step 3 stages from it and step 4 checks it,
so the two can no longer disagree about what "the durable log" means.

**A guard, not just an edit.** `apps/tui/src/durable-log-guards.test.ts` already asserts
two ends of this contract (the accumulus allowlist and `TRACKED_FILES`); the wrapper was
a third statement of the same list with nothing pinning it. A new end-3 case parses the
array out of the shell and asserts it equals `TRACKED_FILES` minus the breadcrumb, with a
second case asserting that exclusion on purpose so it does not read as an oversight. The
parser throws rather than returning `[]` when it cannot find the array — a silent empty
list would pass vacuously against a wrapper that had stopped staging anything.

Without the guard, the eighth durable file repeats this issue exactly.

**Three stale counts** in the wrapper's prose said five or six where seven is now right.

## Verification

- Both mutations seen RED before being trusted: dropping `reconciliations.jsonl` from the
  array (the pre-fix state) fails end 3, and renaming the array fails both cases closed
  with the parser's own error rather than passing empty.
- `apps/tui` typecheck clean; the 8 durable-log guards pass.
- The full `@numisma/price-feed` suite passes, 287 tests including the 15-file wrapper
  harness that runs the real wrapper armed — the step-3/step-4 rewrite is exercised
  there, not just read.

## Not in scope

`head-digest.json` is currently *tracked* in accumulus (`git ls-files data/` lists it),
while the wrapper's lenient arm exists because it "may be intentionally ignored". Not a
defect, but the comment describes a state that is not the live one. Left alone here.

Found during the docs sweep in #397, which documented the seven-file contract and
deliberately changed no executable line.

Closes #398


AmetAlvirde:
## Review: PR #401

Read against `main`, grounded in the live accumulus checkout (`~/Dev/accumulus/.gitignore` does allowlist `!/data/reconciliations.jsonl`, and `git check-ignore --no-index` from inside `data/` returns not-ignored, so the new explicit `git add` will not abort under `set -e`). The core fix is right: the two literals were five files against a seven-file `TRACKED_FILES`, and step 4 would have printed `post-check OK` over an uncommitted trail.

Nothing here blocks the fix. One finding is about how durable the fix is; the rest are accuracy defects in newly added prose.

### 1. Medium: the new guard pins the declaration, not the two consumers

`apps/tui/src/durable-log-guards.test.ts:193-201`

Both end-3 cases go through `wrapperStrictFiles()` (`:174-181`), whose only input is the `^DURABLE_STRICT_FILES=\(...\)` declaration at `ops/price-feed/run-daily-fetch.sh:943`. Neither case inspects `run-daily-fetch.sh:973` (the add loop) or `:1023` (the strict post-check), so nothing in the test tree asserts that either consumer still reads the array.

Failure scenario: someone editing step 4 re-inlines the pathspec, `LOG_DIRTY="$(git -C "$DATA_DIR" status --porcelain -- events.jsonl genesis.json preferences.jsonl orders.jsonl plans.jsonl)"`, and leaves `DURABLE_STRICT_FILES` declared and still used by step 3. All 8 durable-log guards stay green, and #398 is back in its exact original shape: the add loop stages six, the post-check inspects five, and the eighth durable file goes uncommitted under a `post-check OK`.

The wrapper comment at `:925` states the invariant as "ONE LIST, TWO CONSUMERS" and the test comment claims collapsing the literals "is what makes this a single assertion instead of two", but the single assertion covers the list only. The structure holds the invariant; the guard does not. Cheapest close: assert the array reference appears in both consumer positions, or assert the wrapper contains no other bare `*.jsonl` pathspec run of length > 1.

### 2. Low: a stale count introduced by the commit that removes three stale counts

`ops/price-feed/run-daily-fetch.sh:931-932`

"which is the precise failure ADR-006's amendment orders these four edits to prevent". ADR-006's ordered sequence is allowlist, `TRACKED_FILES`, explicit-add loop, strict post-check. This PR collapses the last two into one array, so there are now three edit sites, not four. A reader adding the eighth durable file counts four places and goes looking for a fourth that no longer exists.

### 3. Low: "the one list below" points the wrong way

`ops/price-feed/run-daily-fetch.sh:964-965`

"so also add them explicitly, from the one list below". `DURABLE_STRICT_FILES` is declared at `:943`, twenty-two lines *above* this comment. A reader following the pointer scans down through step 3's body and finds only `"${DURABLE_STRICT_FILES[@]}"`.

### 4. Low: the "single-line" constraint is stated but not enforced, and the throw message misnames its own trigger

`ops/price-feed/run-daily-fetch.sh:937` and `apps/tui/src/durable-log-guards.test.ts:177-180`

`[^)]*` matches newlines, so a multi-line array literal parses correctly and `.trim().split(/\s+/)` yields the right names. The wrapper's instruction to "keep it on one line" therefore constrains a maintainer for no reason, and the thrown message, "no single-line `DURABLE_STRICT_FILES=(...)` literal", describes a condition that cannot produce it. The real trigger is a renamed, deleted, or non-line-start declaration. Either drop the one-line claim from both places, or tighten the regex to `[^)\n]*` so the message becomes true.

### 5. Low: the guard now has three ends, and five places still say two

Outside the diff, but this change is what makes them wrong:

- `apps/tui/src/ingest-commit.ts:41` ("asserts both ends over this exact array")
- `docs/durable-log-ops.md:13`
- `docs/local-data.md:114`
- `docs/plans-authoring-runbook.md:57`
- `context/adr/ADR-006-private-sibling-data-repo-commit-per-ingest.md:231` and `context/adr/INDEX.md:10`, which both state the four-edit order from finding 2

A follow-up `update-docs` pass rather than a change here, but the shell end is the one that actually failed, so leaving the prose at "both ends" understates what the guard now covers.

### Informational: a new way for the evening run to die before the derived surfaces

`ops/price-feed/run-daily-fetch.sh:973-976`

`reconciliations.jsonl` is currently absent and untracked in accumulus, so step 3's `[[ -e ]]` guard skips it and step 4's pathspec matches nothing (`git status --porcelain` with an unmatched pathspec exits 0 with no output, verified). Once the first fill writes it, the explicit `git add` runs. On any accumulus checkout whose `.gitignore` predates the `!/data/reconciliations.jsonl` line, that `git add` exits 1 on an ignored path and `set -e` kills the run at `LAST_STEP="commit"`, skipping steps 5, 5b and 6 (`gap-report`, `operator-notice`, `backfill`). The loud failure is the right stance and the live checkout is correct, so this is a note about the blast radius on a stale clone, not a defect in the change.

### Verified and not a finding

- Sorted `toEqual` comparison holds: `TRACKED_FILES` minus `head-digest.json` is exactly the six names in the array (parsed the real file to confirm).
- Bash 3.2 under the plist's `/bin/bash`: the array is a non-empty literal, so `"${arr[@]}"` under `set -u` is safe.
- No other copy of the durable-file list exists in the repo. The pre-fix greps for `events.jsonl genesis.json` and `preferences.jsonl orders.jsonl` hit only the two wrapper lines this PR collapses.
- No doc cites `run-daily-fetch.sh` by line number, so the 23 added lines above step 3 rot nothing.
- The second end-3 case is implied by the first, but the PR body argues for it as an intent marker, which is a fair call.

PR #402, opened 2026-08-21
fix(ops): render the plist's data dir instead of committing a home path

### LINKED ISSUE #399: the launchd plist template publishes the operator's home path
`ops/price-feed/com.numisma.pricefeed.daily.plist:119` commits a literal:

```xml
<key>NUMISMA_DATA_DIR</key>
<string>/Users/amet/Dev/accumulus/data</string>
```

The same file carries five `__PLACEHOLDER__` values for other machine-specific
settings, so the template already has a mechanism for this and uses it
everywhere else.

**The plist's own comment defends the literal**, and it is right about the
runtime constraint: launchd does not expand `~`, so the *rendered* plist must
carry an absolute path. That argument covers the rendered file on the operator's
machine. It does not cover the template committed to a public repo, which is
what the other five placeholders exist to keep clean.

The reason the codebase gives for banning this literal elsewhere applies here
too. `packages/engine/src/data-dir.ts:113` says: "Never a `/Users/...` literal,
in this file or in any test that pins it, ADR-006 forbids the literal, and a
committed one would leak the operator's real directory layout into a public
repo." That rule is scoped to the engine and its tests, so the plist is not in
breach of it as written. The leak it describes is the same one, and this file is
just as committed and just as public.

### Suggested shape

Render `__DATA_DIR__` like the other five, and add the substitution to the
existing render step in `ops/price-feed/launchagent-reinstall.md`. Note that the
plist needs a manual re-render for this to reach the running job, since the
wrapper installs via `git pull` but the plist does not.

Low severity: a directory layout, not a credential, and already in public
history, so this is fix-forward only with nothing to rotate.

Found during the docs sweep in #397.

AmetAlvirde:
`com.numisma.pricefeed.daily.plist` committed `<string>/Users/amet/Dev/accumulus/data</string>`
while carrying `__PLACEHOLDER__` values for every other machine-specific setting. The
plist's own comment defended the literal, and it was right about the runtime constraint —
launchd does not expand `~`, so the *rendered* file must be absolute. That covers the
rendered copy on the operator's machine and says nothing about the template, which is
what the other placeholders exist to keep clean.

Low severity, and fix-forward only: a directory layout, not a credential, already in
public history, nothing to rotate.

## What changed

**`__DATA_DIR__` renders like the other two.** The substitution joins `__REPO_DIR__` and
`__HOME__` in the reinstall runbook's `sed`.

**Three guards on the template**, in `schedule-window.test.ts` because that file already
reads the plist and already states the rule for what belongs there — a property with an
oracle somewhere else. Here the oracle is `packages/engine/src/data-dir.ts`: *"a committed
one would leak the operator's real directory layout into a public repo."* That rule is
scoped to the engine, so the plist was never in breach as written; the leak is the same
leak.

- no home literal in any value launchd reads
- the data dir goes through `__DATA_DIR__` — a control on the first, since simply deleting
  the key would also satisfy it, and the job would then take the wrapper default silently
  rather than by decision
- every placeholder the template uses has a substitution in the runbook — #398's
  four-edits-in-order failure in its other form, where adding a placeholder without its
  render line produces an install that *looks* rendered

The literal check strips XML comments first. The scoping is the claim: the hazard is a
value, which is what gets installed and what a reader of a public repo learns the layout
from. The header's illustrative `/Users/you/.asdf/shims` example names nobody.

## The part that goes past the issue

Moving off the literal creates a failure mode the literal did not have, and leaving it
open would have made this a worse trade than the leak. The wrapper reads
`${NUMISMA_DATA_DIR-...}` **unset-only**, so an unrendered `__DATA_DIR__` is set,
non-blank and not absolute — it survives verbatim as a CWD-relative path, which is the
split-brain arm ADR-006 forbids.

That gap predates this PR. `normalizeDataDirOverride` has THROWN on a relative override
since #369, while the wrapper refused only blank — so the wrapper's comment claiming it is
"byte-identical to what `resolveDataDir` does in TS" was true for blank and false for
relative. A relative value got through, `HEARTBEAT_FILE` was built from it, and the first
`pnpm` step threw a minute later.

The wrapper now refuses non-absolute with the same exit 78 (EX_CONFIG) beside the blank
check, before any side effect, and names the reinstall runbook when the value looks like
an unrendered placeholder — the only place that can say what actually went wrong, since
everything downstream reports a bad data dir rather than a bad plist.

## Docs

`docs/price-feed-ops.md` stated the opposite in three places ("The plist's
`NUMISMA_DATA_DIR` is **not** a placeholder"), and both it and the runbook enumerated the
placeholders as a pair. Reconciled, including the new refusal in the install steps and a
trap-2 note on spotting a left-over `__DATA_DIR__`.

## Not scrubbed, deliberately

`ops/price-feed/launchagent-reinstall.md` keeps its `/Users/amet` literals. It is an
operator runbook whose `sed` and `launchctl` lines are meant to be pasted; scrubbing them
leaves commands nobody can run. The guard is scoped to the template — the artifact that
gets installed — not the runbook, which gets read.

## Verification

- All three template guards seen RED before being trusted: restoring the literal fails two
  of them, deleting the key fails the control, and removing the runbook's `sed` line fails
  the third.
- Both wrapper cases seen RED with the new refusal removed.
- Full workspace typecheck clean; 292 `@numisma/price-feed` tests pass (287 + 5), including
  the wrapper harness running the real wrapper armed. `plutil -lint` OK.

## Operator action

The plist needs a **manual re-render** to reach the running job — the wrapper installs via
`git pull`, the plist does not. Until then the installed copy keeps its already-resolved
literal and the job is unaffected. Note trap 1 still applies: the render destroys
`NUMISMA_PATH_PREPEND`, which must be re-added by hand.

Found during the docs sweep in #397.

Closes #399

### DIFF
--- a/apps/price-feed/src/schedule-window.test.ts
+++ b/apps/price-feed/src/schedule-window.test.ts
@@ -519,3 +519,61 @@ describe("the launchd fetch window", () => {
     expect(ceiling).toBeGreaterThan(20 * 60);
   });
 });
+// The template is a TEMPLATE, and this is the assertion that keeps it one (#399).
+//
+// It sits in this file for the reason stated in the header: it has an ORACLE SOMEWHERE
+// ELSE. `packages/engine/src/data-dir.ts` states the rule — "Never a `/Users/...`
+// literal, in this file or in any test that pins it, ADR-006 forbids the literal, and a
+// committed one would leak the operator's real directory layout into a public repo" —
+// and scopes it to the engine. The plist was outside that scope and carried
+// `/Users/amet/Dev/accumulus/data` for exactly as long as nothing looked. The leak is
+// the same leak; this file is just as public.
+//
+// The plist's own defence of the literal was right about launchd and wrong about which
+// file it was defending: launchd does not expand `~`, so the RENDERED copy must be
+// absolute — which says nothing about the committed template, and is what the other two
+// placeholders already exist to handle.
+//
+// SCOPED TO THE PLIST, deliberately. `ops/price-feed/launchagent-reinstall.md` is full of
+// `/Users/amet` and stays that way: it is an operator runbook whose `sed` and `launchctl`
+// lines are meant to be pasted, and scrubbing them would leave commands nobody can run.
+// The template is the artifact that gets installed; the runbook is the one that gets read.
+describe("the launchd template carries placeholders, not the operator's paths", () => {
+  const template = readFileSync(PLIST_PATH, "utf8");
+
+  it("commits no home-directory literal in any value launchd will read", () => {
+    // XML comments are stripped first, and that scoping is the claim: the hazard is a
+    // VALUE, which is what gets installed and what a reader of a public repo learns the
+    // operator's layout from. The header's illustrative
+    // `<string>/Users/you/.asdf/shims:…</string>` inside a comment is a worked example of
+    // an OPTIONAL key, names nobody, and is the kind of thing this guard must not forbid.
+    const values = template.replace(/<!--[\s\S]*?-->/g, "");
+    const literals = values.split("\n").filter((line) => /\/Users\//.test(line));
+
+    expect(literals).toEqual([]);
+  });
+
+  it("renders the data dir through __DATA_DIR__, like the repo dir and the log paths", () => {
+    // Control on the assertion above: a template that had simply DROPPED the key would
+    // also carry no literal, and the job would then fall back to the wrapper's default
+    // silently rather than by decision.
+    expect(template).toMatch(/<key>NUMISMA_DATA_DIR<\/key>\s*<string>__DATA_DIR__<\/string>/);
+  });
+
+  it("names every placeholder it uses in the render step of the reinstall runbook", () => {
+    // The four-edits-in-order failure (#398) in its other form: adding a placeholder to
+    // the template without adding its substitution to the runbook produces an install
+    // that looks rendered and is not. The wrapper refuses such a value at startup, but
+    // this catches it a step earlier, at the edit.
+    const runbook = readFileSync(
+      fileURLToPath(new URL("ops/price-feed/launchagent-reinstall.md", REPO_ROOT)),
+      "utf8",
+    );
+    const placeholders = [...new Set(template.match(/__[A-Z_]+__/g) ?? [])].filter(
+      (name) => name !== "__PLACEHOLDERS__",
+    );
+
+    expect(placeholders.length).toBeGreaterThan(0);
+    expect(placeholders.filter((name) => !runbook.includes(`s|${name}|`))).toEqual([]);
+  });
+});
--- a/apps/price-feed/src/schedule-window.test.ts
+++ b/apps/price-feed/src/schedule-window.test.ts
@@ -569,11 +569,30 @@ describe("the launchd template carries placeholders, not the operator's paths", 
       fileURLToPath(new URL("ops/price-feed/launchagent-reinstall.md", REPO_ROOT)),
       "utf8",
     );
-    const placeholders = [...new Set(template.match(/__[A-Z_]+__/g) ?? [])].filter(
-      (name) => name !== "__PLACEHOLDERS__",
-    );
+    // Comments are stripped first, the same scoping the literal guard above uses and for
+    // the same reason: a placeholder only needs rendering if launchd will READ it. The
+    // header's prose about `__PLACEHOLDERS__` is placeholder-SHAPED and gets installed
+    // verbatim as a comment, so scanning the raw template forced a hardcoded exemption
+    // for it, and the next header note that names an optional placeholder would have
+    // needed another one. Strip the comments and the exemption is not needed at all.
+    const values = template.replace(/<!--[\s\S]*?-->/g, "");
+    const placeholders = [...new Set(values.match(/__[A-Z_]+__/g) ?? [])];
+
+    // And the search is scoped to the `sed` invocation itself, not to the whole runbook.
+    // `runbook.includes("s|__X__|")` matched the token in prose or in a commented-out
+    // sample line, so a placeholder that the runbook merely MENTIONS passed a test whose
+    // subject is whether the render step substitutes it.
+    const lines = runbook.split("\n");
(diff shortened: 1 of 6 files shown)


AmetAlvirde:
## Review: #402

Read the diff against `main` plus `packages/engine/src/data-dir.ts`, the full wrapper
config block, and the untouched parts of `docs/price-feed-ops.md`. Verified the three new
template guards by replaying their exact predicates over the branch's plist and runbook
(all three pass, and the literal scan finds nothing), and the new refusal by running its
`[[ ]]` test in bash. The plist parses and carries no `/Users/` outside comments. Nothing
here blocks the plist change itself; the findings are all on the wrapper's new refusal and
on docs it invalidates.

### 1. The wrapper now refuses `~/…`, which the engine is written to accept

At `ops/price-feed/run-daily-fetch.sh:135`, `[[ "$DATA_DIR" != /* ]]` rejects every
non-absolute value, and `~/Dev/accumulus/data` is one of them (confirmed in bash). The
engine's contract table does the opposite: `packages/engine/src/data-dir.ts:129` expands
`~` and `~/…` against `homedir()`, and its refusal at :134 reads `must be an absolute path
or start with "~/"`. The comment at :81 says why that arm exists, and it names this exact
input: *"a value that ORIGINATED as `NUMISMA_DATA_DIR` in a launchd plist, which cannot
expand `~` itself"*. So the row the engine added for the plist is the row the wrapper now
blocks before the engine ever runs.

Failure scenario: operator writes `<string>~/Dev/accumulus/data</string>` into the
installed plist's `EnvironmentVariables` rather than rendering an absolute path. Today the
run reaches `pnpm prices:fetch` and the engine resolves it to the real ledger. After this
PR the whole evening exits 78 at startup, and the message at :141 tells the operator to
"give it an absolute path" while the engine, one layer down, would have told them
`or start with "~/"`. Two planes, two answers, for one value.

Same shape, smaller: the engine `trim()`s before testing (`data-dir.ts:126`), the wrapper
does not, so `" /Users/x/data"` with a leading space is accepted by the engine and refused
by the wrapper (also confirmed in bash).

This may be the behavior you want, since the wrapper's own `HEARTBEAT_FILE` and
`git -C "$DATA_DIR"` cannot expand a tilde either. But then it is a deliberate,
documented *narrowing*, not the parity the comment claims. As written, the block still
says at `run-daily-fetch.sh:76` that it is *"byte-identical to what `resolveDataDir` does
in TS"*, and the PR body frames the change as closing that gap. It closes the relative row
and opens the tilde row. Nothing tests the tilde input on either side of the boundary
(`grep` for tilde cases in `run-daily-fetch.test.ts` finds none), so nothing will catch the
drift the next time either side moves.

### 2. Docs now describe wrapper behavior the wrapper no longer has

`docs/price-feed-ops.md:930-935`, the "`~/`-prefixed value" bullet, ends with *"The
wrapper writes the real notice; your shell reads an empty path and prints nothing,
forever."* After this PR the wrapper writes nothing at all: it exits 78 before
`HEARTBEAT_FILE=`, before `mkdir -p "$LOG_DIR"`, before step 5b's `pnpm operator-notice`.
The divergence that bullet describes stops existing, and the section it sits in is the one
that tells the operator what a quiet channel means.

Two smaller staleness hits in the same file, both untouched by the diff:

- `:61-62` still calls a set-but-blank value *"the one failure with no heartbeat"*. There
  are now two exit-78 refusals with no heartbeat, and this is the section an operator
  reads when a scheduled run looks like it did nothing.
- `:936-941`, the "relative value" bullet, still frames the guard as engine-only (*"The
  snippet has no such guard"*). The wrapper now refuses it first, which is the more useful
  fact for someone triaging a red job.

The PR body says the docs stated the opposite in three places and were reconciled. The
three places at `:25`, `:197` and `:219` are indeed fixed. These are a fourth region, and
they went stale in the other direction, because of the wrapper change rather than the
plist change.

### 3. Cleanup: the placeholder guard scans comments, and needs an exemption because of it

`apps/price-feed/src/schedule-window.test.ts:572` matches `__[A-Z_]+__` over the raw
`template`, so it picks up placeholder-shaped tokens inside XML comments and has to
hard-filter `__PLACEHOLDERS__` to pass. Its sibling two tests up already solved this at
:550 by stripping comments first. I ran both scans over the branch's plist: the raw scan
yields `__PLACEHOLDERS__`, `__REPO_DIR__`, `__HOME__`, `__DATA_DIR__`; the comment-stripped
scan yields exactly the three real ones. Reusing `values` drops the hardcoded exemption
and stops a future header note that mentions, say, an optional `__PATH_PREPEND__` from
failing a test about rendering.

Weaker in the other direction: the `runbook.includes("s|" + name + "|")` check matches the token
anywhere in the runbook, including inside prose or a commented-out sample line, so a
placeholder that is *mentioned* in the runbook but missing from the actual `sed`
invocation still passes. Scoping the search to the `sed -e` block would make the guard
say what its name says.

### Checked and clean

- All three template guards pass on the branch content, and the literal scan is correctly
  scoped: the header's `/Users/you/.asdf/shims` example survives comment-stripping.
- The plist parses as XML, carries no `--` inside a comment, and holds no `/Users/` in any
  value.
- The refusal sits above `HEARTBEAT_FILE=`, `mkdir -p "$LOG_DIR"` and the `exec … tee`
  redirect, so "before any side effect" holds. The harness assertion that carries that
  claim is the `Library` one, and it would genuinely go red without the fix.
- `${VAR-default}` keeps its no-colon spelling, so the unset arm still takes the default
  and the new check cannot swallow it.
- The runbook's `sed` also rewrites `__DATA_DIR__` inside the plist's comment prose, so the
  installed copy's comment reads as if it documented a literal. Cosmetic only, and `plutil
  -p` ignores comments, so trap 2 is unaffected.

### Pre-existing, adjacent

`docs/price-feed-ops.md:893` gives the wrapper's form as `${NUMISMA_DATA_DIR:-$HOME/Dev/accumulus/data}`,
with the colon, contradicting `:63` in the same file and the wrapper itself. Not this PR's
doing, but it sits in the four-readers table that this PR's absolute requirement also
belongs in, so it is cheap to fix while you are here.

PR #411, opened 2026-08-22
The component layer's structural pass: render-test harness, shared primitives, compound conversion

AmetAlvirde:
The component layer's structural pass, all seven slices of the spec, one commit each.
`apps/web` gets its first render-test toolchain, five shared primitives, and a fill path
that coordinates selection through context instead of threading two prop shapes down
from a parent.

No product behavior changes. No route change, no wire change, no schema change, no CSS
change: `git diff --stat main..HEAD -- apps/web/src/styles.css` is empty.

## The commits

| Commit | Slice |
| --- | --- |
| `9af0ad0` | S1 — buys the harness, asserts the chart's a11y invariant (T7), writes ADR-022 |
| `2ff02b6` | S2 — `Absent` gets one home; five private copies deleted |
| `16bb772` | S3 — the `Card` compound, adopted by the glance and summary cards |
| `e811243` | S4 — the composition table and the DCA card onto `Card` |
| `3c951fc` | S5 — the snapshot notices and the crumb, one home each |
| `b2ca9f7` | S6 — the fill path's internals onto the shared primitives |
| `2757e0c` | S7 — the fill path becomes the one tier-2 compound |

## Where jsdom attaches

Per-file `// @vitest-environment jsdom` docblocks, and nothing else. `vitest.config.ts`
gains no `environment` key, no `projects`, no `setupFiles`; its gitignore-derived
`exclude` and its 30 s budget are each still stated exactly once. A missing docblock
fails red at the first `render()`; a projects split would have failed toward green when
a render test matched no include. `apps/web/src/jsdom-docblock-guard.test.ts` holds the
claim no runtime check can make.

The buy is three dev dependencies: `jsdom` at the root, `@testing-library/react` and
`@testing-library/user-event` under `apps/web`. No `jest-dom`. No react plugin in the
test path — vitest's esbuild honors the web app's `"jsx": "react-jsx"` unaided.

## T7 is closed, and it has teeth

The chart mounted for real under jsdom on the first attempt, with `ResizeObserver` and
`matchMedia` stubs and no mock. The S1 fallback never fired. Demonstrated by mutation
rather than asserted in prose:

- Deleting the `.sr-only` caption element → assertion 3 red.
- Putting a `<button>` under `.fp-chart` → assertion 2 red.
- Removing a docblock from any `*.test.tsx` → the guard red.

One caveat, recorded in the test header, `docs/coverage-rationale.md` §6 and ADR-019's
residual-risk section: under `@tanstack/charts` 0.11.0 the adapter renders
`tabindex="-1"` on its `<svg>` regardless of `tabIndex`, `focus` or `keyboard` —
verified by flipping each and re-running green. The assertion therefore pins the
rendered outcome, not the four neutralization props.

## Two deviations from the slice briefs, both deliberate

**`glance/row-view.test.ts`'s two regexes were re-spelled, not retired** (S2). That file
is a source scan asserting the literal `<Absent reason="no-fund-value" />` is present and
`<Absent reason="unexpected-absence" />` is not. The spelling cannot survive an extraction
whose point is that the primitive takes resolved words instead of an enum. Both arms
survive against the new `whyAbsent(...)` spelling, with the reason written inline. Same
claim, nothing retired.

**`<section className="card fp-selected" aria-live="polite">` stays a raw section**
(S6, S7). `Card`'s Seam C signature is `{ className, children }`, so routing this element
through the primitive would drop `aria-live` off the one element that has to announce.
Widening `Card` is the thing the spec fixed the signature to prevent. The reason is
written at the call site and a test pins the attribute's location.

## Scan regexes

None retired. Nothing in `rung-state-seam.test.ts` or `route-move.test.ts` claims what
the new render tests claim, so D5's one-for-one rule never fired. The
`venueAxis === "filled"` census is untouched and still permits the predicate in exactly
`ladder/fill-path-view.ts` (once) and `push/fixture-synthesis.ts` (twice).
`route-move.test.ts`'s assertions are byte-identical to `main` across all seven commits;
only its header prose changed.

## The spec's success signals, checked

1. `pnpm verify` green at every slice, and at HEAD: 174 files / 2658 tests passed.
2. `styles.css` diff against `main` is empty.
3. `route-move.test.ts` assertions byte-identical.
4. One home per primitive: one `Absent`, one `Crumb`, one `SnapshotEmptyNotice`, one
   `SnapshotStaleNotice`, one `Card`.
5. Every domain card's external prop API is unchanged; no route file JSX moved.
6. No `selectedIndex`/`onSelectIndex` pair survives, and nothing outside the provider
   converts a rung key to an index or back.
7. T7 asserted, closed, not downgraded.
8. `useEffect` still appears exactly once in `apps/web/src`, in `lib/binance-spot.ts`.
9. Every `*.test.tsx` under `apps/web/src` opens with the jsdom docblock.
10. No regex retired without its replacement, vacuously — none retired.

## What this closes

This single PR is what closes #404, and what closes #405, and what closes #406, and what
closes #407, and what closes #408, and what closes #409, and what closes #410, and what
closes #403.


AmetAlvirde (commented):
## Review: PR #411, spec #403's seven slices

**Verified independently, not taken from the PR body.** Checked out `2757e0c` in a
scratch worktree against the repo's real `node_modules`:

- `tsc --noEmit -p apps/web/tsconfig.json` clean.
- Full `vitest run`: 2655 passed, 4 skipped files, 0 failures. (One extra failure,
  `ops/testkit/gitignored-path-globs.test.ts`, is an artifact of reviewing from a
  worktree; it passes in the real checkout. Not a defect in this PR.)
- Success signals 2, 4, 8 confirmed by hand: `styles.css` diff empty, exactly one
  `Absent` / `Crumb` / `SnapshotEmptyNotice` / `SnapshotStaleNotice` / `Card`,
  `useEffect` still only in `lib/binance-spot.ts`.
- `route-move.test.ts` assertions are byte-identical; only header prose moved.

**Behavior preservation holds.** I walked every converted component for class set,
heading level, rendered string and ARIA attribute. `Card` composes `card` with the
caller's words and emits the same strings the raw sections did; `whyAbsent` reproduces
the old `reason ? REASON_COPY[reason] : "suppressed"` arm exactly, with the default now
living in the primitive; the notices' JSX whitespace collapses to the same sentences the
three routes rendered. The one expression the Seam E conversion rewrote in the selection
path, `view.rungs[selectedIndex]?.key` becoming `selected?.key`, is provably equivalent:
`selectedIndex` is `findIndex` of `selected.key`, and both spellings yield `undefined` on
an empty ladder.

**The render tests are not vacuous.** I mutation-tested them rather than reading them.
Red where it should be red: deleting the `.sr-only` class, dropping `aria-hidden` off
`.fp-chart`, demoting either `level={1}`, making `Card` interpolate a trailing space,
removing the rung rows' `onFocus`, pinning `aria-current` to every row, changing
`Absent`'s default word, and stripping a jsdom docblock (the guard names the offending
file). The chart-a11y file's anti-vacuity guard, asserting the adapter's own
`[tabindex="-1"]` is present, does its job.

**What the mutations found green that should have been red** is below, in the inline
comments. Nothing here contradicts an acceptance list: every one of the seven lists is
satisfied as written. The gap is between what the tests hold and what the spec's own
problem statement promised, and it clusters on one control.

**Both named deviations hold.** Judgments inline at
`glance/row-view.test.ts` and at the `fp-selected` section.

Nothing I found blocks merge.

PR #419, opened 2026-08-23
The component package and its workbench: @numisma/components, the --nms-* token spec, and react-cosmos

AmetAlvirde:
Turns the tailwind-package spike's PROVEN verdict into a package two consumers build against, and converts its two silent failure modes from findings into standing tests.

One PR for the whole spec, per #412 §6. Nothing merged per-slice.

Closes #412
Closes #413
Closes #414
Closes #415
Closes #416
Closes #417
Closes #418

## What lands

`@numisma/components` ships unbuilt TSX with no `dist`, no build step and no CSS of its own. It declares twelve `--nms-*` tokens and their grayscale defaults; every consumer supplies the values. `apps/web` mounts Tailwind 4 as a second linked stylesheet beside `styles.css`, adopts the package `Button` on its sign-in form, and mints an accent. `apps/workbench` is a standalone react-cosmos app with three theme modes and no SSR.

## The two silent failures, and what now catches them

Both were invisible in the spike, and they demand opposite diagnostics.

**A missing token emits no rule at all.** Tailwind 4 emits a utility only when its theme variable exists. `apps/web/src/tailwind-scan.test.ts` builds the stylesheet and asserts it carries a utility only the package can produce. Removing the package `@source` line drops the built file from 20,606 to 7,216 bytes, takes the asserted class to zero, and the build still exits 0. Demonstrated red, then restored.

**A missing bare var emits a correct-looking rule that computes to nothing.** `ops/components/nms-namespace.test.ts` fails on any `var(--…)` read in package source outside the `--nms-` namespace, naming the offending file. Demonstrated red with a planted `var(--muted)`, then reverted.

Each test states in its own comment what it does **not** prove. A passing build is not evidence of scanning; a namespaced read is not evidence the token has a value. Conflating those two channels is the mistake the pair exists to prevent.

## Verification

`pnpm typecheck` clean across nine projects. `pnpm test` at **191 files passed / 3 skipped (194)**, **2850 passed / 52 skipped (2902)** — up from 177 files / 2672 tests at `f15a812`.

`apps/web/src/styles.css` differs from `f15a812` by exactly one additive 70-line `:root` block, and its five structure tests assert the new content.

From a wiped `.vercel/output`, the production build emits both stylesheets and every CSS filename referenced from `functions/` resolves in `static/assets/`. tailwindlabs#19888 did not reproduce.

### Verified in a browser, because no text search can prove it

The `@theme` indirection resolves. `text-destructive`, `bg-destructive/10` and a bare `color: var(--nms-destructive)` all compute to the same base color with the modifier at the expected alpha rather than transparent, so the spec's literal-values fallback was not needed. `rounded-md` computes to 6px from the non-color `--nms-radius-md`.

Preflight did not land: a bare `<p>` still computes the UA default margin, and the built stylesheet carries no preflight markers.

The sign-in button renders through the package `Button` in the app's palette. Setting `--accent` at runtime turned the focus ring green while the fill stayed blue, which separates the two channels: the ring travels `--accent → --nms-ring → --color-ring`, and the fill is won by `.auth-card button`'s unlayered literal hex.

All three workbench modes change computed colors, measured in Chrome:

| | `bg-primary` | `bg-secondary` | `text-destructive` | `rounded-md` |
| --- | --- | --- | --- | --- |
| grayscale | `oklch(0.205 0 0)` | `oklch(0.97 0 0)` | `oklch(0.45 0 0)` | 6px |
| themed | `rgb(122,62,161)` | `rgb(44,140,120)` | `rgb(198,40,40)` | 14px |
| app | `rgb(59,108,240)` | `rgb(38,42,51)` | `rgb(240,115,106)` | 8px |

## Findings worth reading before review

**The token spec cannot be derived by scanning `var()` reads.** Only 3 of Button's 12 tokens appear as a `var()` anywhere. The other nine are reached through theme utilities (`bg-primary`, `border-ring`), where no `var()` is written and the name is still required. `pnpm components:add` therefore reads class names against a closed list of shadcn roles, and refuses rather than guessing when a token has no known default. A placeholder would satisfy the test and ship a wrong color.

**The scanning guard nearly disarmed itself.** `@source "./"` scans `apps/web/src` including test files, so spelling the asserted class as one literal in the test fed the instrument to the extractor and the guard went green with the package unscanned. The class is now split across a concatenation, and a third case asserts no file under `apps/web/src` contains it while `button.tsx` does.

**The generated tokens file is imported with `layer(theme)`** although it is not Tailwind's own. `__root.tsx` links `styles.css` first, so in document order the generated defaults would beat it. Inside a layer they lose to unlayered CSS instead, which is what lets Slice 5 override a token from `styles.css`'s own `:root`.

**The adopted button's fill does not travel the token path.** `.auth-card button { background: #3b6cf0 }` is unlayered and beats `layer(utilities)`, exactly as Seam C designs. The tokens govern everything that rule does not set. Rewriting it belongs to the `styles.css` migration, which #417 places out of scope.

**react-cosmos 7.4.0 cannot serve its own UI under pnpm.** Its dev server calls express 5's `res.sendFile` with no options, and send@1's `dotfiles: "ignore"` default 404s any path containing `node_modules/.pnpm/`. The playground loads blank with `mountPlayground is not defined` as the only symptom. Fixed by `patches/react-cosmos@7.4.0.patch` through `patchedDependencies`; the rejected alternative, `virtualStoreDir`, costs a full relink for every developer and CI run. On a react-cosmos bump the patch must be re-cut, and `pnpm install` fails loudly if it stops applying.

**The spec's own `--muted` count was wrong.** §10 Q1 says 27 call sites; the real number is 25 rules plus the new alias, and `styles.css`'s own comment already said 25. ADR-023 records the correct figure.

## Docs and ADRs

`docs/component-package.md` carries what the package ships, what a consumer owes it, the two failure modes with their two diagnostics, and the manual theming procedure. That procedure is manual because jsdom will not resolve `color-mix` through a cascade; Playwright is deferred to its own increment rather than bought as a side effect of this one.

**ADR-023** records unbuilt TSX plus the consumer-supplied namespaced token spec. **ADR-024** records react-cosmos standalone, with the pnpm patch as a named cost. Per §8, the preflight omission and the `layer(utilities)` asymmetry deliberately get **no** ADR: both are coexistence tactics with a designed expiry, and an ADR for a decision scheduled to be unmade reads as still binding.

## Out of scope, and staying out

None of `styles.css`'s 1,396 lines move. `apps/web/src/components/ui/*` stays where #411 put it. The package ships `Button` and nothing speculative.


AmetAlvirde (commented):
Five findings: two high, two low, one medium. I checked the claims this increment says are invisible by building stylesheets, not by reading assertions.

## What held

**The scanning guard is genuinely armed.** I reproduced its negative control outside the repo with the pinned tailwindcss 4.3.3. Scanning `apps/web/src` alone emits 6,882 bytes and no `.rounded-[min(var(--nms-radius-md),8px)]`; adding the package source emits 13,511 bytes and the selector appears.

**The concatenation defence works.** A scanned file containing `"rounded-[min(var(--nms-radius" + "-md),8px)]"` produces no candidate at all, not even the left half; the same literal spelled whole emits the rule. The split is load-bearing and correct, and the third exclusivity case covers the whole surface `@source "./"` reaches.

Also verified: `styles.css` differs from f15a812 by one hunk, 70 added lines, zero removed, no selector touched. `packages/components/tsconfig.json` declares no `paths`, and `tsconfig-paths.test.ts:41` holds that against the real file rather than a fixture. Every `var()` read in `packages/components/src` is `--nms-`. The workbench imports nothing from `apps/web`, and writing the seam test against import specifiers is what keeps the drift test's disk read from tripping it. `--nms-nms-*` cannot be produced and a re-add cannot duplicate a token entry. The `#3b6cf0` contrast claim of 4.6:1 against white computes correctly.

**The fourteen-versus-twelve asymmetry holds**, and the second reason is the one that carries it. It also survives the obvious objection: the day `Card` enters the package, `nms-tokens.test.ts`'s "maps %s" case goes red until `@theme` gains `--color-card`, so the two carried names cannot rot into a silent hole.

One observation, not a finding. `@source "./"` now makes Tailwind emit utilities for any hand-written class name that happens to be a utility. I diffed the emitted selectors against all 102 class names `styles.css` writes: there is exactly one collision, `.sr-only`, and Tailwind's version is a strict subset of the hand-written rule with identical values. The unlayered file wins regardless. Seam C behaves as designed.

## What did not hold

Both high findings sit in one place. The namespacing rewrite and the standing namespace guard match the literal string `var(`, and Tailwind 4 has a second syntax for a custom-property read that never writes it. The guard reports zero offences on a package that reads the consumer's `--muted` bare.

## Merge

Neither high finding breaks anything that ships today. `button.tsx` uses neither form, so the package is clean right now and every green result on this branch is honest. They are a hole in the instrument, not a defect in the product, and the hole opens on the next `pnpm components:add` of a component that uses the shorthand, which includes two of shadcn's own. My call: this does not block merge, and the regex widening should land before the next component enters the package.

PR #431, opened 2026-08-24
refactor(web): tear styles.css down to its token blocks

AmetAlvirde:
Tears `apps/web/src/styles.css` down from 1,466 lines to 103: two `:root` blocks,
the palette and the twelve `--nms-*` aliases, and no rule at all. Every
presentational rule is now a Tailwind utility on the element that carries it.
Succeeds `component-package` (#419, `1b5bdca`), which built the ground this
runs on.

Ten slices, delete-as-you-go, in the order the spec set: locks, auth, shell,
summary, table, glance, DCA, Fill Path, ladder, chart.

## What holds it up

Four failure modes here are silent and all of them are colour-shaped: a utility
that emits nothing, one that computes to transparent, one that is correct in the
text and still loses the cascade, and one that compiles and paints the wrong
grey. Grep catches none of them, so the migration is held by four channels
instead:

- **Scan guards** prove the rule is emitted. Two now: the package's, and the
  app-side `min-w-[320px]` sentinel on `<body>` added in slice 0.
- **Census successors** prove the element references it. Six `*-structure`
  files, each converted in the same commit as its deletion, per-class
  `toContain`, never full-string equality.
- **The terminal assertion** proves nothing survives with a carrier: rendered
  class names intersected with the class selectors left in `styles.css` is
  empty, on all six surfaces.
- **A Chrome computed-style check per slice** proves the utility won the
  cascade and computed to the right value. Colours asserted by computed value,
  at 320px and desktop, recorded in each slice's final commit body.

The standard evidence format from slice 3 onward: stash the slice, probe the
pre-slice tree with the identical script, diff computed properties and bounding
rects, then break each measurement deliberately to show it can fail. Slice 8 ran
11,088 comparisons over 313 element/width pairs with zero differences and seven
negative controls. Slice 9 diffed fourteen sections against a second dev server
running the pre-slice tree.

Suite goes from 192 files / 2,868 tests to 204 / 3,011. `pnpm typecheck` green.
The nitro production build still pairs its CSS hashes.

## What actually changed on screen

One repaint, accepted and named in the spec: deleting `.auth-card button` lets
the package `Button`'s own geometry through, which is why that rule was the
first switch of the migration. Colour is unchanged by construction.

One defect found and fixed, invisible to the entire suite. `border-solid` and
`border-dashed` set `border-style` on four sides, and with preflight off nothing
had zeroed the UA's `border-width: medium`, so each legend swatch rendered as an
18x5 box with a 3px grey edge down three sides instead of an 18x2 rule. Only the
pre/post computed diff caught it. Fixed with the per-edge arbitrary property in
`6f918a8`.

Everything else is byte-identical, or differs only where the difference is
provably inert: `flex: 1 1 0px` against `1 1 0%`, and a `border-bottom-style` on
an edge whose width is `0px` on both sides.

## Decisions the build had to make

Five places where the written record disagreed with the shipped code. Shipped
behaviour was preserved in every one, and each is recorded in its commit body:

- The `Button`'s real geometry is `h-9 px-2.5`, not the `px-4` the spec's S1
  paragraph names.
- `.verdict-yes` (the `needs you` arm) paints `--neg`, not `--pos`. The
  alarming answer gets the alarming colour, which is what the component's own
  docblock argues for. If the issue text is right and the app has been wrong,
  that is a product change and belongs in its own increment.
- `.dca-alert a:focus-visible` declared no outline, only an underline. The ring
  is the user agent's, alive because preflight is off.
- `.dca-alert` declared neither a background nor a border.
- There are six structure test files, not the five the spec counts. The
  terminal assertion went on all six.

Spec §10's open question is settled the other way round from its guess:
Tailwind 4.3.3's `sr-only` carries `clip-path: inset(50%)` and omits the legacy
`clip`, so no `@utility` extension is owed. ADR-025 records it.

Two seams the spec did not have to cut, both handled by keeping a bare class
name alive until its last descendant rule died: `error` through slice 2, and
`absent` through slices 3, 7 and 8. Slice 8 deleted the hook.

## Docs

ADR-025 records the end-state cascade contract: `styles.css` is the permanent,
unlayered home of the palette and the aliases and holds no rule; Tailwind runs
without preflight indefinitely with the house base rules in `@layer base`; theme
colour utilities are package-only in app code and house colours are read as
`[var(--x)]`. `docs/component-package.md` §7 is rewritten for the end state, §4's
guard table now carries the four standing guards with their negative controls,
and a new §8 documents the Chrome checklist procedure as the app-side twin of
§5's manual theming procedure, caveats included.

## Known gaps, stated rather than smoothed

- Keyboard dispatch through the browser extension failed in two of three
  attempts (slices 6 and 8). Where it failed, the focus ring was read under a
  real `:focus-visible` requested in script, and the commit body says so instead
  of claiming a Tab that never arrived. The keyboard assertions in
  `fill-path-selection.test.tsx` are unmodified and green.
- `-webkit-overflow-scrolling: touch` is emitted in the built stylesheet, but
  desktop Chrome exposes no computed value for it, before or after. Text channel
  only.
- Three ladder arms and two `.absent` positions have no fixture or live state
  that renders them. Each was measured on both trees through an identical
  patch or an injected carrier using the exact class string the component
  emits, and each is recorded as a gap in its commit body.
- Slice 9's live ladder route was measured on the converted tree only; the
  pre-slice server had no session. The fixture route carries the parity claim,
  which is what the spec names it for.

This closes #421, closes #422, closes #423, closes #424, closes #425,
closes #426, closes #427, closes #428, closes #429 and closes #430. It also
closes #420, the spec they slice.


AmetAlvirde (commented):
## Review — parity migration, ten slices

Read against spec #420 §3's per-slice gate, §4's seams and §8's success signal. I
ran the suite in a clean worktree at `596a7b1`: **203 files passed / 4 skipped,
3,009 tests passed / 54 skipped, exit 0**; `pnpm typecheck` exit 0. I also
compiled every `className` candidate in the app's non-test `.tsx` against the
repo's own `tailwindcss@4.3.3` and read the emitted order, so the cascade claims
below are measured rather than argued.

One repaint is out. `text-[13px]` staying `text-[13px]`, arbitrary-value
verbosity, and anything of the shape "this could be styled better" are §2.2/D5
decisions and are not reported.

Ranked, most severe first.

---

### 1. The torn-act banner's `<code>` chip lost `.notice code` — a live repaint on a production route

`apps/web/src/components/FillPath.tsx:493` — `<code>pnpm orders:fill</code>`
carries no class.

On `main`, `TornActBanner` rendered `<div className="card notice fp-torn">`
(`main:apps/web/src/components/FillPath.tsx:297`), and `notice` was the hook that
brought `.notice code { background: #000; padding: 2px 6px; border-radius: 6px }`
(`styles.css:340-344` on `main`) onto the `<code>` at `:303`. `TORN`
(`FillPath.tsx:468-469`) spells the surface out by hand with the `--neg` border,
correctly, and drops `notice` — and nothing replaces the chip.

Four `<code>` elements on `main` sat inside a `.notice` box. Three were
converted and import `NOTICE_CODE`: `ui/SnapshotNotice.tsx:66`,
`routes/ladder-fixture.$state.tsx:92` and `:135`. This is the fourth.

**Failure scenario.** `/ladder/$planId` with `view.tornActs.status ===
"outstanding"`. The command the operator is meant to read out of the one red
block that claims a hard stop renders as bare UA monospace on the card ground,
instead of a black chip with 2px/6px padding and a 6px radius. Every viewport.
Nothing in the suite goes red — `fill-path-cards-structure.test.tsx` asserts the
banner's own utilities, not its descendant's, and the terminal assertion is
vacuous (finding 2).

Slice 2's commit body claims "every carrier of every one of them, wherever it
renders, now carries the utilities instead" and names this banner among the
twelve `card` carriers. The `<code>` inside it is the one that slipped.

Fix is one import and one attribute: `<code className={NOTICE_CODE}>`.

---

### 2. The terminal assertion cannot fail, and is subsumed by the guard beside it

Six sites, identical body: `summary-card-structure.test.tsx:278`,
`section-table-structure.test.tsx:267`, `glance-card-structure.test.tsx:259`,
`dca-card-structure.test.tsx:389`, `fill-path-cards-structure.test.tsx:1283`,
`shell-structure.test.tsx:92`.

`styleSheetClassSelectors()` (`render.testkit.tsx:450`) strips comments, strips
declaration bodies, and scans what is left for `.name`. Run against the
end-state `styles.css` it returns **zero selectors** (measured). So `styled` is
empty, `survivors` is `[]` for any render at all, and all six assertions are
`expect([]).toEqual([])`.

The parser itself is sound — it handles compounds like `.notice.error h1`,
survives at-rule nesting, and cannot read `.5s` as a class — and
`render.testkit.test.tsx:97` pins it against a sample so a reader that stopped
reading is caught. The problem is the end state, not the instrument.

Its only failure mode requires a rule to return **and** a component to render its
class. `styles-css-end-state.test.ts:150` already reds on **any** rule returning,
carrier or not:

```js
expect(openers).toEqual([":root", ":root"]);
```

plus a per-declaration check that every property starts with `--`. The terminal
assertion's failure set is a strict subset. Its docblock's claim — "this says
something the file cannot know" — was true while the file still held rules and
is not true now.

Not a weakening: nothing was removed to make room for it, and Seam E asked for
it. But six tests that cannot fail are worth either deleting or rewriting to
assert their own non-vacuity.

---

### 3. Three cards lost their only witness that `Absent` still renders

The old censuses pinned `"absent"` and `"muted absent-why"` as **present**, which
incidentally proved the `Absent` primitive mounted on each card's suppression
path. Slice 8 deleted the name, so the successor asserts the inverse — and
`not.toContain("absent")` is satisfied both by a correct conversion and by the
element being gone.

- `glance-card-structure.test.tsx:233` (was `main:63-64`). This is the only test
  in the repo that renders `GlanceCard`, and its fixture sets `change: { rendered:
  false, suppressedBy: "reference-withheld" }`, which drives `GlanceCard.tsx:149`.
- `summary-card-structure.test.tsx:252` (was `main:81,86`).
- `dca-card-structure.test.tsx` — `main:97,103` carried them; the branch file
  asserts nothing about `Absent` at all, while `DcaCard.tsx:257/264/271` render
  three.

**Failure scenario.** Change `GlanceCard.tsx:149` to `return null`. On `main` the
census goes red. On this branch every assertion stays green: the deleted-name
loop still holds, no query reaches the suppressed slot, and the terminal
assertion is vacuous. A card ships printing nothing where an em dash and a stated
cause belong — the invariant `ui/Absent.tsx:4` calls load-bearing.

`section-table-structure.test.tsx` is the same shape but partially mitigated:
`glance/row-view.test.ts:273-292` scans the source for `rowView.rendered` and the
`<Absent>` call, so deleting the branch breaks something.

The cheap fix is the move `fill-path-cards-structure.test.tsx:396` already makes
— find the primitive by its em dash rather than by a class name.

---

### 4. `base-layer.test.ts:25` spells the sentinel class in prose, which the design forbids

`tailwind.css:80-88` states the invariant: the sentinel class is written nowhere
but `__root.tsx`, "here included", because Tailwind extracts candidates from
comment text as readily as from JSX. `base-layer.test.ts:25` writes it out in a
docblock.

I confirmed the mechanism rather than assuming it: a `.test.ts` file whose only
content is a docblock naming the class emits `.min-w-\[320px\]` into the built
stylesheet under `@source "./"`.

Bounding it honestly, because it is smaller than it looks. I removed `@source
"./"` from `tailwind.css` in a scratch worktree and ran the guard: **red at
`app-scan-sentinel.test.ts:127`**, so the negative control still holds — deleting
that line stops the whole directory being scanned, test files included. And
`:133` still pins exactly one non-test writer, so removing the class from
`__root.tsx` is caught.

What is lost is narrower: `:127` no longer proves that `__root.tsx`'s class is
the one being scanned, only that some file in `src` supplies the candidate. An
eroded claim, not a live hole — and the file it erodes is the one whose whole
subject is not letting a guard supply its own evidence.

---

### 5. `row-suppressed` is rendered and asserted nowhere

`apps/web/src/components/SectionTable.tsx:160`. `main` pinned it present at
`section-table-structure.test.tsx:115` and `:133`; the branch mentions it in no
test file. It never had a rule on either ref, so this is drift rather than
breakage — but deleting the class from the `<tr>` now goes unnoticed, and
anything later keyed on it misses silently.

---

### 6. `expectClasses`'s null guard passes on `undefined`

`fill-path-cards-structure.test.tsx:88-92`. `expect(element).not.toBeNull()`
passes on `undefined`, and callers do pass possibly-undefined values
(`absentIn(container).at(-1)` at `:414`, a `.find(...)` around `:1041`). The test
still fails, because `classTokens` then throws on `undefined.getAttribute` — but
it fails as a TypeError rather than as "the element is missing". `toBeTruthy()`.

---

### 7. `grid-cols-1` is not `1fr` — noting, not claiming

`FillPath.tsx:185-187`. `.fp-tiles` set `grid-template-columns: 1fr`, which is
`minmax(auto, 1fr)`; `grid-cols-1` compiles to `repeat(1, minmax(0, 1fr))`, which
has no min-content floor. Observable only if a tile's min-content exceeds the
card interior, and I could not construct that at 320px against the figures this
renders. Low confidence; recorded so it is not rediscovered as new.

---

## Checked and found clean

- **Cascade order.** I compiled every candidate set and read emitted position.
  `.m-0` before every `.mt-*`/`.mb-*`; `.border` before `.border-l-4`;
  `.border-[color]` before `.border-l-[color]`; every arbitrary property after
  the shorthand it must beat; every container variant after every unvariant
  utility. The base-plus-override cascades a class string cannot express were
  converted to total maps — `BADGE_TONE`, `PILL_TONE`, `ROW_TINT`/`ROW_EDGE`,
  the three swatch constants — and each map's priority matches the deleted rules'
  source order.
- **Preflight-off traps.** No siblings of `6f918a8`. Every `border-style`
  utility in the app has a width on the edge it styles: `PILL`'s `border-dashed`
  rides `border`, `rungRowClasses` rides `ROW`'s `border`, the swatches use
  `[border-top-style:…]` + `border-t-2`, `WARN_INFERRED` uses
  `[border-left-style:dashed]` + `border-l-4`. Tailwind 4's
  `--tw-border-style` variable makes `border` + `border-dashed` order-independent
  in any case — verified in the emitted CSS. UA margins are reproduced wherever
  the deleted rule was holding them off, `<dd>`'s 40px indent included.
- **Container names.** `GlanceCard.tsx:46` splits `@container` from
  `[container-name:glance_metrics-card]` on purpose: Tailwind's named
  `@container/glance` emits the `container` shorthand, which would have dropped
  `metrics-card` and stranded the metrics list at its narrow layout with nothing
  in a diff to see. Every `@[380px]/name:` variant in the app has a matching
  `@container/name` ancestor.
- **Slice 0's two-sided edit.** `nms-tokens.test.ts`'s surplus arm and the
  workbench's `toBe(NMS_TOKEN_NAMES.length)` now mirror in both directions;
  dropping an alias from one side alone reds the other.
- **Login form.** `main`'s `.auth-card input` never set `font-family`, only
  `font-size`, so `text-[1rem]` is parity and not a dropped inherit.
  `p-2.5`/`rounded-lg`/`border` are exact against 10px/8px/1px.
- **Table.** `th, td` and `th.num, td.num` were selector lists on `main`, so the
  header cells' padding, hairline and alignment are reproductions, not additions.

## Checked and dismissed

`.notice.error h1` looked at first like a descendant colour widened onto its
container in `SnapshotNotice.tsx:95` and `ladder-fixture.$state.tsx:119`. It is
not: `main`'s div was `className="card notice error"` and `.error { color:
var(--neg); margin: 0 }` painted the whole box, so `text-[var(--neg)] m-0` on the
container is exact and the `<h1>`'s own copy is the redundancy its comment says
it is.

## Read

Merge-blocking: finding 1 only, and it is a one-line fix. Findings 2 to 6 are
follow-ups that do not change what ships. The parity evidence in the commit
bodies held up everywhere I re-derived it, including the two places I expected to
catch it out.

PR #438, opened 2026-08-24
The generics migration: four components into @numisma/components, three token mints

### LINKED ISSUE #432: Spec — the generics migration: four components into @numisma/components, three token mints, five slices
Succeeds increment `component-package` (spec #412) and the Tailwind migration
(spec #420, PR #431, `e38f569`), which between them built everything this runs
on: `@numisma/components`, the `--nms-*` token specification, `apps/workbench`
with its three theme modes and four guards, and an `apps/web` with no
hand-written presentational CSS left. Opens from grill
`notes/numisma/2026-08-23-component-package-migration-grill.md` (D1–D5) and the
repo review in
`notes/numisma/2026-08-23-generics-migration-spec-handoff.md`.

This is **wave 1 of three**. Wave 2 is the domain migration, wave 3 is polish.
They do not share a PR with this one and do not inherit this spec.

## 1. Problem

`@numisma/components` holds one component. Eleven live in `apps/web/src/components`,
and `apps/workbench` — the diagnostic surface built to review components against
three palettes — can see exactly none of them.

That is not a gap the workbench can close from its side. `@numisma/workbench`
depends on `@numisma/components` and React and nothing else, `cosmos.config.json`
watches only its own `src`, and `seam-isolation.test.ts` fails on any import
specifier reaching `apps/web`. `fixture-coverage.test.ts` stays green throughout,
because it demands fixtures for package exports and the package exports one
component. The guard is sound; it measures a set of one.

The components move. Generics first, and that is this spec.

Moving a file across that boundary is not a `git mv`, because **the token
vocabulary inverts on the way in**. Package source may read colour only inside
the `--nms-` namespace (`ops/components/nms-namespace.test.ts`); app source must
read house colours bare, as `[var(--muted)]` (`apps/web/src/theme-color-utilities.test.ts`).
Which rule applies is decided by the directory the file sits in. So every colour
read in all four components is rewritten as the file moves, and three of the
names it needs to rewrite to do not exist yet.

Two of the five reads are the wrong-by-collision trap. `--muted` in the app is
**secondary text**; `--nms-muted` in the package is the **recessed surface**.
A mechanical `--muted` → `--nms-muted` rewrite compiles, emits a rule, paints
the app's most-used grey wrong, and leaves every guard green — because the file
just left the tree `theme-color-utilities.test.ts` scans, and the package-side
guard only checks that the read is inside the namespace, never that it is the
right name in it. That failure is invisible to the entire test suite and this
spec's naming section is the whole defence against it.

### 2.1 In

- Four component files into `packages/components/src/ui/`, kebab-cased:
  `Absent.tsx` → `absent.tsx`, `Card.tsx` → `card.tsx`, `Crumb.tsx` → `crumb.tsx`,
  `SnapshotNotice.tsx` → `snapshot-notice.tsx`.
- Eight names onto the package's public surface: six components
  (`Absent`, `Card`, `CardTitle`, `Crumb`, `SnapshotEmptyNotice`,
  `SnapshotStaleNotice`) and two data exports (`CARD_SURFACE`, `NOTICE_CODE`).
- Three new `--nms-*` names, minted on all three sides that must agree.
- Six cosmos fixtures, one per exported component, in `apps/workbench/src/ui/`.
- The `renderLink` slot on `Crumb`, and its three route call sites.
- Two guard edits: the `route-move.test.ts` runtime allow-list, and each
  `theme-color-utilities` scan-tree consequence.
- Every `apps/web` import of the four moved components rewritten to
  `@numisma/components`.
- `docs/component-package.md` and the two prose docblocks that name
  `ui/Absent.tsx` by path.

### 2.2 Out, and staying out

- **The seven domain components.** Wave 2. They keep importing `CARD_SURFACE`
  and `NOTICE_CODE` from the package the moment those move, and nothing else
  about them changes.
- **The `glance/` and `ladder/` view modules.** Every domain component imports
  app-local view types from them; re-homing is wave 2's real unknown and earns
  its own diagnosis.
- **`/ladder-fixture/$state`.** It cannot die before the view modules resolve:
  `started-ladder.fixtures.test.ts:188` reads its source for a DCE assertion,
  and a `FillPathCards` fixture needs `composeFillPathPage`, which
  `seam-isolation.test.ts` forbids the workbench from importing. Wave 2.
- **The seven names this spec declines to mint** (§4.2). Named here, minted
  when read.
- **Any visible change.** This migration is behaviour-preserving. Polish is
  wave 3, and the review gates differ: one PR doing both tangles parity
  evidence with intentional-change evidence, and you lose the ability to say
  "nothing moved" about the half where it is true.
- **The login a11y fix (map §5.1).** Independent of all three waves, still
  shippable on its own.
- **Moving the contract tests into the package** (§4.5).

## 3. The per-slice gate

Every slice ships green on the full suite. Beyond that, each component slice
proves the move did not repaint anything:

1. **The read rewrite is enumerated, not grepped.** The slice brief lists every
   `var(--x)` occurrence in the file by line, with the `--nms-*` name it becomes.
   A count that does not match the brief fails the slice.
2. **App mode renders the component identically.** The cosmos fixture in app
   mode is the parity check: the same values `apps/web` paints, reached through
   the new names. A colour that changes is either a naming error or a missing
   alias, and both are §4.1 failures.
3. **Themed mode repaints every read.** A token that does not visibly change
   when the fixture switches into themed mode is a token the fixture does not
   exercise, which means the fixture is incomplete, not that the token is fine.
4. **`fixture-coverage.test.ts` goes from silent to demanding** on that
   component the moment it is exported. The slice that exports is the slice that
   fixtures.

### 4.1 The token vocabulary — the package owns names, the consumer owns values

`packages/components/src/tokens.ts` is the contract. Slice 0 widens it by three
and does not touch a component.

The four generics read exactly five house names — verified by grep over the four
files, with counts: `--muted` ×4, `--card` ×2, `--line` ×2, `--neg` ×2,
`--text` ×1. Two of the five already have a package name at the same role:

| house read | becomes | status |
|---|---|---|
| `--line` | `--nms-border` | exists |
| `--text` | `--nms-foreground` | exists |
| `--card` | `--nms-card` | **mint** |
| `--muted` | `--nms-muted-foreground` | **mint** |
| `--neg` | `--nms-neg` | **mint** |

**This does not reverse spec #420 S0, and the distinction is the whole point.**
S0 deleted `--nms-card` and `--nms-muted-foreground` from `styles.css` on the
rule `tokens.ts` still states verbatim: *"Add a name here when a component
starts reading it, not before."* Those two aliases had been minted by #412 §4.2
for components that had not arrived, and until they arrived app mode carried two
values no fixture could render, under a label saying it shows what the app
shows. Slice 0 is the day they arrive. The argument `app-token-drift.test.ts`
records as *"the argument that lost"* — carry an unexercised alias now, so the
day `Card` enters the package app mode is already correct — is still lost. We
are not reviving it; we are satisfying its precondition. Both docblocks that
narrate the deletion get a sentence saying so, so the next reader does not read
the mint as a quiet reversal.

**`--nms-neg` is minted distinct from `--nms-destructive`, deliberately.**
`styles.css:94` already writes `--nms-destructive: var(--neg)`, so in app mode
the two names resolve to the same colour today, and reusing `--nms-destructive`
for `SnapshotNotice` would cost nothing this week. It is refused because the two
are different kinds of thing: **`--neg` is data — the sign of a number — and
`--nms-destructive` is intent — the affordance of a button that destroys
something.** Welding them means the day the money-red wants to soften, or wants
a colourblind-safe pair with `--nms-pos`, you cannot move one without moving the
other, and the rename lands in a wave where eleven components read it. Two names
at one value now is free; one name at two meanings later is not. It also keeps
`--nms-pos` available as `--nms-neg`'s symmetric partner in wave 2 without the
pair reading as `pos`/`destructive`.

**`--muted` cannot become `--nms-muted`** and this is the §1 trap. `--nms-muted`
is the recessed surface (`styles.css:84` aliases it to `--recess`); the app's
`--muted` is secondary text. `--nms-muted-foreground` is the shadcn-shaped name
for it, and it is the name #420 S0 deleted — reusing that exact spelling is
intentional, so the vocabulary stays one vocabulary.

**Minting is a three-sided edit.** Every new name lands in all three or the
guards red:

- `packages/components/src/tokens.ts` — the name, a grayscale default, and a
  note on what it is for.
- `apps/web/src/styles.css` — the alias line (`--nms-card: var(--card)`,
  `--nms-muted-foreground: var(--muted)`, `--nms-neg: var(--neg)`), keeping that
  file the one place the app defines colour.
- `apps/workbench/src/theme-modes.ts` — `APP_TOKENS` gains the three **resolved**
  literals (the drift test walks the app's alias chain to meet them), and
  `THEMED_TOKENS` gains three values **pairwise distinct from all fifteen**.

Held by `apps/web/src/nms-tokens.test.ts` (styles.css declares no `--nms-*` the
package never declared), `apps/workbench/src/app-token-drift.test.ts` (exact
name-for-name equality plus themed distinctness), and
`packages/components/src/tokens.test.ts`.

**Grayscale defaults.** `--nms-neg`'s package default is grey, like every other,
and that means grayscale mode cannot review sign — a negative number and a
positive one are two greys. That is correct and stated in the token's note:
grayscale mode reviews hierarchy, spacing and state; sign is reviewed in themed
and app mode, which is what those modes are for. The same note pre-answers the
question for `--nms-pos` in wave 2.

### 4.2 The other seven names, declared and not minted

Grill D2 ruled "mint all ten in slice 0", on #412's argument that minting the
full namespace kills the collision surface for all ten at once. That is
overruled, because #420 S0 overturned its premise after the grill was written
and `tokens.ts`'s own rule now says the opposite. Minting ten would mean
rewriting that rule, both drift docblocks, and seven grayscale defaults and
seven distinct themed values for names nothing renders — the exact unverifiable
state S0 deleted.

D2's real concern is answered instead by **naming all ten here, now, and minting
the five the generics read.** Wave 2 mints from a written vocabulary under no
pressure, rather than making `--muted`'s split a second time:

| house name | package name | wave |
|---|---|---|
| `--bg` | `--nms-background` | exists |
| `--text` | `--nms-foreground` | exists |
| `--line` | `--nms-border` | exists |
| `--card` | `--nms-card` | **1** |
| `--muted` | `--nms-muted-foreground` | **1** |
| `--neg` | `--nms-neg` | **1** |
| `--pos` | `--nms-pos` | 2 |
| `--ok` | `--nms-ok` | 2 |
| `--warn` | `--nms-warn` | 2 |
| `--now` | `--nms-now` | 2 |

This table is the deliverable of the decision. It goes into `tokens.ts`'s header
as a comment, so the four unminted names are a written commitment in the file
that owns them rather than a note in a spec nobody re-reads.

### 4.3 How a migrated component reads colour: bare `var()`, not theme utilities

House components read `text-[var(--nms-muted-foreground)]`. They do not read
`text-muted-foreground`.

`Button` reads through Tailwind theme utilities (`bg-primary`) mapped in
`apps/web/src/tailwind.css` `@theme`, and that stays — it is shadcn-derived and
the utilities are how shadcn writes components. The house four do not follow it,
for cost: the utility form makes every mint a five-sided edit instead of three
(`@theme`, plus `ops/components/consumers.ts` which generates the workbench's
CSS), and it grows `theme-color-utilities.test.ts`'s `THEME_COLOR_NAMES` — a
**literal** list of nine — or else `text-neg` in app code slips past the guard
that exists to catch exactly that.

The bare form also makes the migration mechanical and reviewable: `[var(--x)]`
becomes `[var(--nms-x)]`, one substitution per line, which is what §3.1 asks the
slice brief to enumerate.

Stated as a rule in the package: **shadcn-derived components read theme
utilities; house components read bare `var(--nms-*)` inside arbitrary values.**
Both resolve to the same value by the `@theme` mapping `tokens.ts` already
documents, so this is a source-form convention, not two palettes.

### 4.4 `renderLink` — the router leaves the package

`Crumb` takes a slot typed:

```ts
renderLink: (props: { className: string; children: ReactNode }) => ReactElement;
```

The package owns the anchor's classes and hands them to the slot. `apps/web`
supplies a fully-typed `<Link>`; the workbench fixture supplies `<a href="#">`.
`to` leaves the package entirely.

Routerlessness is load-bearing for the workbench — the vite config's header
calls the absence of a router and SSR "the whole point of the second consumer" —
so `@tanstack/react-router` is not available to the package. Chosen over a
link-adapter context because TanStack checks `to` against the generated route
tree; an adapter normalizing to `to: string` throws that check away silently. A
fixture whose link navigates is a fixture that just left the state you were
looking at, so the `<a href="#">` stub is correct, not a compromise.

Consequences in wave 1: three call sites (`routes/index.tsx`,
`routes/big-picture.tsx`, `routes/ladder.$planId.tsx`) carrying five `<Crumb>` usages, and
`crumb-contract.test.tsx:50`'s `renderCrumb(to, label)` helper changes shape.
`Crumb`'s hover becomes `hover:text-[var(--nms-foreground)]`.

### 4.5 The app→package boundary, and what it costs the guards

- **`route-move.test.ts:213-244`** walks the ladder route's relative import
  closure and allows exactly two runtime `@numisma/*` specifiers
  (`@numisma/engine/format`, `@numisma/engine/calendar`). `ladder.$planId.tsx`
  and `FillPath.tsx` import `@numisma/components` at runtime from wave 1 on.
  `@numisma/components` joins `allowed`, with a docblock sentence saying why —
  the closure guard is a fifth guard on this migration and `Card.tsx`'s own
  docblock warns of it.
- **`theme-color-utilities.test.ts`** needs more than ten non-test `.tsx` files
  under `apps/web/src` to be a meaningful scan. Nineteen exist; fifteen survive
  wave 1. No edit needed, and the number is recorded here because wave 2 takes
  it to eight and the guard has to be re-homed then, not now.
- **`packages/components/tsconfig.json` lacks `allowImportingTsExtensions`**,
  while the app files import `./Card.tsx`. Imports are rewritten extensionless
  on the way in, the same rewrite `components:add` performs for `@/`. The
  tsconfig is not changed to accommodate the old form.
- **File names are kebab-case in the package**, for every component, generic and
  domain alike — matching `ui/button.tsx`, which the shadcn CLI wrote and which
  is the convention `components:add` will keep writing. This is the house rule
  from here on, so wave 2 renames the seven the same way and no file arrives
  needing a second rename.

### 4.6 The contract tests stay in `apps/web`

`absent-contract`, `card-composition`, `crumb-contract` and
`snapshot-notice-copy` import `../../render.testkit.tsx`, whose header states RTL
is imported at exactly one path in the repo; the package has no RTL or jsdom
devDependency. They stay where they are and import the components from
`@numisma/components`, which keeps them testing the shipped surface rather than
a local file.

Moving them is a real question with a real answer — a package's contract tests
belong to the package — but it needs an RTL harness in the package and a second
decision about `render.testkit.tsx`'s one-path rule. It is out of scope here and
should be revisited when wave 2 doubles the number of tests in this position.

## 5. Slices

Five, each independently reviewable. S1–S4 are the same shape, run four times,
which is the point: the first one that goes wrong goes wrong across 93 lines.

### S0 — Mint the three names

`tokens.ts` gains `--nms-card`, `--nms-muted-foreground`, `--nms-neg` with
grayscale defaults and notes; `styles.css` gains three alias lines;
`theme-modes.ts` gains three `APP_TOKENS` literals and three distinct
`THEMED_TOKENS` values; the ten-name table lands in `tokens.ts`'s header; the
two docblocks that narrate #420 S0's deletion gain the sentence from §4.1. No
component moves.

Done when: `nms-tokens.test.ts`, `app-token-drift.test.ts` and `tokens.test.ts`
are green with fifteen names on both sides, and themed mode has fifteen pairwise
distinct values.

### S1 — `Absent`

The smallest complete pipe, carrying the highest-risk read alone: `--muted` ×3,
and nothing else. `absent.tsx` into the package, exported, one fixture,
`SectionTable`/`GlanceCard`/`SummaryCard`/`FillPath` imports rewritten, the two
prose docblocks that name `ui/Absent.tsx` by path corrected
(`render.testkit.tsx:172`, `SummaryCard.tsx:68`).

If `--nms-muted-foreground` is going to be wrong, it is wrong here, in one file,
against a fixture in app mode.

### S2 — `Card` and `CardTitle`

`--card` ×2, `--line` ×2. Two component exports plus `CARD_SURFACE`, which fifteen
app files import as data and which moves with its file. Two fixtures, or one
naming both — `Card.Title` usage does **not** satisfy `fixture-coverage` for
`CardTitle`; the name must appear in a named import from `"@numisma/components"`.

### S3 — `SnapshotEmptyNotice` and `SnapshotStaleNotice`

`--neg` ×2, and the slice that proves `--nms-neg` renders. `NOTICE_CODE` moves
with the file. Two exports, two fixtures.

### S4 — `Crumb` and the `renderLink` slot

`--muted` ×1, `--text` ×1, plus the only API change in the wave. Three route call
sites build the `<Link>`; the workbench fixture supplies `<a href="#">`;
`crumb-contract.test.tsx` changes helper shape. Held to its own PR-visible slice
because it is the one place a reviewer must check a type, not a colour.

## 6. Test needs

| need | where |
|---|---|
| fifteen names agree across `tokens.ts`, `styles.css`, `APP_TOKENS` | existing, S0 |
| fifteen pairwise-distinct themed values | existing, S0 |
| every exported component has a fixture | existing, flips live per slice |
| package source reads only inside `--nms-` | existing (`nms-namespace.test.ts`) |
| no app-code house colour via theme utility | existing, scan tree shrinks |
| ladder route's runtime `@numisma/*` closure | **edited S1** (allow-list + docblock) |
| `Crumb` renders through the slot, not a router | **new**, S4, in `crumb-contract` |
| the four contract tests pass against package imports | existing, re-pointed per slice |

No new test file is expected. If a slice wants one, that is a signal the move
changed behaviour, and the slice stops.

## 7. Docs and ADRs

- `docs/component-package.md`: the export surface, the kebab-case rule, and the
  shadcn-utilities-vs-house-`var()` rule from §4.3.
- `tokens.ts` header: the ten-name table, and the note that four are named and
  unminted.
- **No ADR.** The three decisions that could earn one — bare `var()` over theme
  utilities, `--nms-neg` distinct from `--nms-destructive`, mint-when-read over
  mint-all-ten — are each already recorded in the file that enforces them, which
  is where a reader hits them. Revisit if wave 2 finds any of the three fighting
  back.

## 8. Success signal

- `@numisma/components` exports eight names; six have fixtures; the workbench
  renders all six in three modes.
- Zero occurrences of `var(--card)`, `var(--line)`, `var(--muted)`,
  `var(--neg)`, `var(--text)` in `packages/components/src`.
- Zero imports of `./ui/Absent.tsx`, `./ui/Card.tsx`, `./ui/Crumb.tsx`,
  `./ui/SnapshotNotice.tsx` anywhere in `apps/web`.
- `apps/web/src/components/ui/` holds only the four contract tests.
- Full suite green, and app mode in the workbench paints the same values
  `apps/web` paints, checked per slice rather than at the end.
- Wave 2 opens against a written ten-name vocabulary and a proven move recipe.

## 9. Decisions, ruled

1. **The package is the home for every component** (grill D1). Generics this
   wave, domain next. The workbench becomes the viewing surface it was built to
   be.
2. **Mint what the generics read; name all ten** (grill D2, reopened and
   overruled). §4.1, §4.2.
3. **`--nms-neg` is distinct from `--nms-destructive`.** Sign is data,
   destructive is intent. §4.1.
4. **House components read bare `var(--nms-*)`; shadcn-derived read theme
   utilities.** §4.3.
5. **Links leave through `renderLink`** (grill D3), over a link-adapter context,
   for route type safety. §4.4.
6. **Three increments, migration never sharing a PR with polish** (grill D4).
7. **Kebab-case component filenames throughout the package**, generic and domain
   alike. §4.5.
8. **Contract tests stay in `apps/web`** for this wave. §4.6.
9. **No skeletons** (grill D5) — unchanged by this spec, recorded because it is
   what makes wave 3 small.

## 10. Left undecided, on purpose

- **Where the contract tests finally live.** §4.6 — revisit at wave 2.
- **Whether `theme-color-utilities.test.ts` survives wave 2**, when the scan tree
  drops to eight files. It holds at fifteen; it does not obviously hold at eight.
- **What polish changes beyond the login a11y fix.** It depends on states nobody
  can see yet, which is what this wave is for.
- **Re-homing `glance/` and `ladder/`.** Wave 2's real unknown; earns its own
  diagnosis, not a guess here.

### LINKED ISSUE #433: Slice 0 — records the ten-name token vocabulary and the package's two source conventions
Spec: #432 (wave 1 of three, the generics migration).

Foundational slice. It moves no component and declares no token. Its job is to
put the vocabulary and the two source-form conventions into the files that
enforce them, before the first file crosses the boundary, so that wave 2 mints
from a written table and no reader mistakes the coming mints for a reversal of
spec #420's slice 0.

## Why this slice is not §5's S0

Spec #432 §5 proposed a slice 0 that mints `--nms-card`,
`--nms-muted-foreground` and `--nms-neg` while no component moves. That slice
cannot ship green.

`packages/components/src/tokens.test.ts` carries a standing assertion, "declares
nothing the components do not read". It joins every package source file except
`tokens.ts` and fails any declared token name that appears in none of them,
either literally or through a Tailwind theme utility naming the same role.
Verified against the current package source: the roles `card`,
`muted-foreground` and `neg` are all unread. Minting three names with no reader
reds that test, which is the same rule §4.2 quotes when it declines to mint all
ten.

Ruled: **mint each name in the slice whose component first reads it.**
`--nms-muted-foreground` lands with `Absent`, `--nms-card` with `Card`,
`--nms-neg` with the snapshot notices. This is what `tokens.ts`'s own rule says
verbatim, and it makes each token edit part of a tracer bullet instead of a
separate hop. The cost, accepted: the four component slices serialise rather
than fanning out from a single mint.

## What lands

**The ten-name table, in `tokens.ts`'s header.** Every house colour name paired
with its `--nms-*` counterpart, each marked as existing, minted in wave 1, or
named-and-unminted for wave 2. The four wave-2 names are a written commitment in
the file that owns them, so the wave-2 author is not deciding `--muted`'s split
a second time under pressure. The table is prose in the file that already states
the mint-when-read rule, not a new export and not a new test.

**The two source conventions, in `docs/component-package.md`.**

1. *Colour reads.* shadcn-derived components read Tailwind theme utilities
   (`bg-primary`); house components read bare `var(--nms-*)` inside arbitrary
   values (`text-[var(--nms-muted-foreground)]`). Both resolve to the same value
   through the `@theme` mapping the token spec already documents, so this is a
   source-form rule, not a second palette. Stated with its cost: the utility
   form would make every mint a five-sided edit and would grow
   `theme-color-utilities.test.ts`'s literal name list, which is the list that
   catches an app-side `text-neg`.
2. *File names.* Every component file in the package is kebab-case, generic and
   domain alike, matching the name the shadcn CLI writes and the name the add
   script keeps writing. Wave 2 renames its seven the same way, and no file
   arrives needing a second rename.

**The narration correction, at three sites.** Spec #420's slice 0 deleted
`--nms-card` and `--nms-muted-foreground` from the app's stylesheet on the rule
that a token nothing reads is a token nothing can verify. Wave 1 re-mints both
spellings. That is not a reversal, it is the day the precondition was satisfied,
and every place that narrates the deletion has to say so or the next reader
reads a quiet undo:

- the `APP_TOKENS` docblock in the workbench's theme-mode data, which explains
  why the table carries twelve names and once carried fourteen;
- the header of the workbench's app-token drift test, which names both deleted
  aliases;
- the comment on the drift-adjacent case in the app's `nms-tokens` test that
  currently asserts the app defines **no** `--nms-muted-foreground`. This third
  site is not prose. It is a live negative assertion, and the slice that mints
  the alias inverts it. Recording that here means the mint slice finds a
  sentence already written for it rather than deleting an assertion that looks
  deliberate.

## Acceptance criteria

- `tokens.ts`'s header carries a table naming all ten house colours with their
  `--nms-*` counterparts, and marks which are declared but not yet minted.
- `docs/component-package.md` states the theme-utility-versus-bare-`var()` rule
  and the kebab-case filename rule, each with the reason it exists.
- All three sites that narrate spec #420's deletion of `--nms-card` and
  `--nms-muted-foreground` say that wave 1 re-mints them, and why that satisfies
  the deletion's rule rather than reversing it.
- `NMS_TOKENS` still declares twelve names. The app's stylesheet still declares
  twelve `--nms-*` aliases. `APP_TOKENS` and `THEMED_TOKENS` still carry twelve.
- Full suite green.

This slice has no failing test of its own. That is deliberate and it is why it
is named foundational rather than shipped as part of slice 1: its content is
what the wave-2 author reads, and burying it inside a component move makes it
invisible.

## Out of scope

No token declared, no stylesheet alias written, no `APP_TOKENS` or
`THEMED_TOKENS` entry added, no component moved, no export added to the package,
no fixture written, no guard edited. Everything in spec #432 §2.2 stays out for
the whole wave: the seven domain components, the `glance/` and `ladder/` view
modules, the `/ladder-fixture/$state` route, the four wave-2 token names, any
visible change, the login accessibility fix, and moving the four contract tests
into the package.

### LINKED ISSUE #434: Slice 1 — moves Absent into the package and mints --nms-muted-foreground
Spec: #432 (wave 1 of three, the generics migration).
Blocked by: #433.

The smallest complete pipe across the package boundary, carrying the wave's
highest-risk token rename alone. `Absent` reads one house colour and nothing
else, so if `--muted` is going to be renamed wrong it is wrong here, in one
file, against a fixture that can be looked at in app mode.

The collision this slice exists to catch: in the app's vocabulary `--muted` is
**secondary text**; in the package's vocabulary `--nms-muted` is the **recessed
surface**. A mechanical `--muted` to `--nms-muted` rewrite compiles, emits a
rule, wins its cascade, paints the app's most-used grey wrong, and leaves every
guard in the repo green, because the file has just left the tree the app-side
colour guard scans and the package-side guard only checks that a read sits
inside the namespace, never that it is the right name in it. The correct target
is `--nms-muted-foreground`, and this slice mints it.

## The per-slice gate (spec §3)

1. **The read rewrite is enumerated, not grepped.** The table below is the
   contract. A count that does not match it fails the slice.
2. **App mode renders the component identically.** The cosmos fixture in app
   mode paints the same value `apps/web` paints today, reached through the new
   name. A colour that changes is a naming error or a missing alias.
3. **Themed mode repaints every read.** A token that does not visibly change
   when the fixture switches into themed mode is a token the fixture does not
   exercise. That is an incomplete fixture, not a fine token.
4. **`fixture-coverage.test.ts` goes from silent to demanding** on `Absent` the
   moment it is exported. The slice that exports is the slice that fixtures.

## The colour reads

`Absent` performs **exactly two** colour reads, both of `--muted`, both becoming
`--nms-muted-foreground`. Anchored by what they sit in rather than by line
number:

| # | where | today | becomes |
|---|---|---|---|
| 1 | the exported surface constant, the colour utility following `inline-flex items-baseline gap-1.5` | `text-[var(--muted)]` | `text-[var(--nms-muted-foreground)]` |
| 2 | the inner reason `<span>`'s class attribute, alongside `text-[0.72rem] font-medium` | `text-[var(--muted)]` | `text-[var(--nms-muted-foreground)]` |

**Two, and the count is the failing condition.** Spec #432 §5 says three, and
§4.1's grep says `--muted` appears four times across the four files. Both counts
include prose. See below.

## The third occurrence is prose, and it must not be namespaced

The file's docblock quotes the deleted `styles.css` rule verbatim, including
`color: var(--muted)`, in the section explaining that the styling moved into the
component. That is history, not a read.

It still matters, because the package's namespace guard scans **raw file text
with no comment stripping**. The moment this file lands in the package, that
docblock occurrence is reported as a bare read outside the namespace and the
guard reds.

Rewriting it to `--nms-muted-foreground` is not the fix: the deleted rule never
read that name, and a docblock that says it did is a false record in the one
place a future reader will trust. **Reword the quoted declaration so it does not
spell `var()`, keeping the historical fact intact.** This is also the only way
spec §8's "zero occurrences of `var(--muted)` in `packages/components/src`"
signal is satisfiable at all.

The same docblock's paragraph about the import closure should be re-pointed
while it is open: it explains the file's dependency-free discipline by naming the
app's ladder-route closure guard, and from this slice on the file is a package
file whose discipline is enforced from the other side.

## The mint, which is a three-sided edit

`--nms-muted-foreground` lands in all three places or the guards red:

- **the package token spec**: the name, a grayscale default, and a note saying
  what reads it. The note must be explicit that this is secondary text and is
  **not** `--nms-muted`, which is the recessed surface, because those two names
  standing next to each other in one list is exactly where the collision gets
  re-made.
- **the app stylesheet**: the alias `--nms-muted-foreground: var(--muted)`,
  keeping that file the one place the app defines colour. The spelling is
  deliberately the one spec #420's slice 0 deleted, so the vocabulary stays one
  vocabulary.
- **the workbench theme-mode data**: `APP_TOKENS` gains the **resolved** literal
  the app's alias chain arrives at, `#9aa1ad`, because the drift test walks that
  chain to meet a literal rather than comparing alias text. `THEMED_TOKENS`
  gains a loud value distinct from all twelve existing themed values.

## Guard edits this slice owes

- **The app's `nms-tokens` test carries a live negative assertion** that the app
  defines no `--nms-muted-foreground` alias. This slice inverts it, and rewrites
  the comment above it, which currently explains the absence by citing spec
  #420's slice 0. Slice #433 pre-wrote the replacement sentence.
- **The same test's guard-the-guard floor** asserts the package declares at
  least twelve tokens. Thirteen exist after this slice; raise the floor to
  thirteen. Left at twelve it stays green forever while meaning nothing, which
  is the failure mode every guard-the-guard in this suite is written against.
- **The same test's alias-form table**, which pins five aliases to their exact
  `var(--x)` text, gains `--nms-muted-foreground` paired with `var(--muted)`.
- **The ladder route's import-closure guard** allows exactly two runtime
  `@numisma/*` specifiers. From this slice on the fill path imports
  `@numisma/components` at runtime and sits in that closure, so
  `@numisma/components` joins the allow-list, with a docblock sentence saying
  why it is safe under ADR-007's client-bundle invariant: the package declares
  only `@base-ui/react`, `class-variance-authority`, `clsx` and `tailwind-merge`
  and reaches no `node:` builtin, unlike the engine root that guard exists to
  keep out of the browser bundle.

## What moves, and every consumer

`Absent.tsx` becomes `absent.tsx` under the package's `ui/` directory, with its
imports written extensionless. The package tsconfig is **not** given
`allowImportingTsExtensions` to accommodate the old form.

`Absent` joins the package's named public surface.

**Five runtime importers**, not the four spec §5 lists. Spec §5 omits the DCA
card:

- the section table
- the glance card
- the summary card
- the fill path
- **the DCA card**

Plus the `Absent` contract test, which stays in `apps/web` for this wave and
imports from `@numisma/components` so it tests the shipped surface rather than a
local file.

**Two prose docblocks name the old path** and go stale on the move: one in the
render test kit, explaining which marker witnesses an `Absent`, and one in the
summary card, explaining a substitution. Both are corrected.

## The fixture

One cosmos fixture under the workbench's `ui/` directory, importing `Absent` by
name from `"@numisma/components"`. A mention in a comment does not satisfy the
coverage guard, and neither does a re-export: the guard matches a named import
from that exact specifier. The fixture must render both the bare form and the
form carrying a reason string, since both paint through read 2.

## Acceptance criteria

- `packages/components` exports `Absent`, and its source file is kebab-cased
  under `ui/`.
- No file anywhere in `apps/web` imports `Absent` from a relative path. Five
  components and one contract test import it from `@numisma/components`.
- `packages/components/src` contains zero occurrences of `var(--muted)`, in
  code or in prose.
- The package declares thirteen tokens. The app's stylesheet declares thirteen
  `--nms-*` aliases. `APP_TOKENS` and `THEMED_TOKENS` each carry thirteen, and
  the thirteen themed values are pairwise distinct.
- A workbench fixture imports `Absent` by name from `"@numisma/components"`, and
  the fixture-coverage guard demands it.
- In the workbench's app mode the component paints `#9aa1ad` on both reads,
  which is what `apps/web` paints today. In themed mode both reads visibly
  change.
- The ladder route's closure guard allows three runtime `@numisma/*` specifiers
  and its docblock says why the third is safe.
- Full suite green, including the package namespace guard, the app's colour
  utility guard, the app-token drift test, the app's `nms-tokens` test and the
  package's own token test.

## Out of scope

`Card`, `CardTitle`, `CARD_SURFACE`, the two snapshot notices, `NOTICE_CODE` and
`Crumb` all stay where they are. No token other than `--nms-muted-foreground` is
minted. `Absent`'s props, markup, class strings other than the two reads above,
and its rendered output do not change: this is behaviour-preserving, and a
visible difference is a defect, not polish. The contract test stays in
`apps/web`. No new test file. Everything in spec #432 §2.2 stays out.

### LINKED ISSUE #435: Slice 2 — moves Card, CardTitle and CARD_SURFACE into the package and mints --nms-card
Spec: #432 (wave 1 of three, the generics migration).
Blocked by: #433, #434.

The card surface crosses the boundary. This is the widest-reach slice in the
wave, not because the component is complicated but because `CARD_SURFACE` is
imported as data by fourteen files, and the snapshot notices in the next slice
cannot move until it has: a package file cannot reach back into `apps/web` for
the class string it composes.

## The per-slice gate (spec §3)

1. **The read rewrite is enumerated, not grepped.** The table below is the
   contract. A count that does not match it fails the slice.
2. **App mode renders the component identically.** The cosmos fixture in app
   mode paints the same surface `apps/web` paints today.
3. **Themed mode repaints every read.** Both the fill and the border must
   visibly change when the fixture switches into themed mode. A card whose
   border does not move in themed mode is a fixture that is not showing you the
   border.
4. **`fixture-coverage.test.ts` goes from silent to demanding** on both `Card`
   and `CardTitle` the moment they are exported.

## The colour reads

The two components perform **exactly two** colour reads between them, both
inside the exported surface constant. `CardTitle` reads no colour at all.

| # | where | today | becomes | status |
|---|---|---|---|---|
| 1 | the exported surface constant, the border colour utility | `border-[var(--line)]` | `border-[var(--nms-border)]` | name exists |
| 2 | the exported surface constant, the background utility | `bg-[var(--card)]` | `bg-[var(--nms-card)]` | **mint** |

**Two, and the count is the failing condition.** Spec #432 §5 and §4.1 say
`--card` twice and `--line` twice. Both extra occurrences are prose.

`rounded-xl` and `p-4` are untouched. `rounded-xl` is the one default-scale
class in the constant and it is an exact match for the deleted 12px radius;
`rounded-md` would not be, because the app's `@theme` remaps the medium radius
onto the package's 8px control radius. That is the same shape of trap as the
colour utilities, in the radius namespace, and it is why the house colours are
read as arbitrary values in the first place.

## Two prose occurrences, and they must not be namespaced

The surface constant's docblock quotes the deleted `.card` rule verbatim,
including `background: var(--card)` and `border: 1px solid var(--line)` on one
line, to explain what the four utilities replaced.

The package's namespace guard scans **raw file text with no comment stripping**,
so both are reported as bare reads outside the namespace the moment the file
lands in the package. Rewriting them to `--nms-card` and `--nms-border` would
record a rule that never existed. **Reword the quoted declaration so it does not
spell `var()`, keeping the historical fact.** This is also what makes spec §8's
"zero occurrences of `var(--card)` and `var(--line)` in
`packages/components/src`" true.

The same docblock's paragraph on the import closure, which explains the file's
dependency-free discipline by naming the app's ladder-route closure guard,
should be re-pointed at the same time. Spec §4.5 notes this file's docblock
warns about that guard specifically.

## The mint, which is a three-sided edit

`--nms-card` lands in all three places or the guards red:

- **the package token spec**: the name, a grayscale default, and a note saying
  what reads it, namely the card surface that eight elements in the app carry.
- **the app stylesheet**: the alias `--nms-card: var(--card)`. The spelling is
  deliberately the one spec #420's slice 0 deleted.
- **the workbench theme-mode data**: `APP_TOKENS` gains the resolved literal
  `#181b22`; `THEMED_TOKENS` gains a loud value distinct from all thirteen
  existing themed values.

`--nms-border` already exists and is already aliased. Nothing about it changes.

## Guard edits this slice owes

- **The app's `nms-tokens` test floor** rises from thirteen to fourteen.
- **Its alias-form table** gains `--nms-card` paired with `var(--card)`.
- The ladder route's closure allow-list already carries `@numisma/components`
  from #434. No further edit.

## What moves, and every consumer

`Card.tsx` becomes `card.tsx` under the package's `ui/` directory, imports
written extensionless. `Card`, `CardTitle` and `CARD_SURFACE` join the package's
named public surface.

`Card.Title` stays attached as a plain property and stays asserted equal to the
named `CardTitle` export. `Card` stays a `<section>` and only a `<section>`: no
polymorphic `as` prop, because four elements in the app carry the surface
without being cards and they import the string, not the component.

**Five importers of the `Card` component**: the section table, the glance card,
the summary card, the fill path, the DCA card.

**Fourteen importers of `CARD_SURFACE`.** Spec §5 says fifteen; fifteen files
contain the identifier, but one of those is the file that defines it. The
fourteen are the four card-bearing components' structure tests, the fill path
and its two tests, the glance/summary/section/DCA structure tests, the card
composition test, the snapshot notice source and its copy test, the auth card
utilities test, the login route, the ladder route, and the ladder fixture route.

**The ladder fixture route is one of them, and spec §2.2 lists that route as out
of scope.** Both are true and the brief says so on purpose: the route is not
moved, deleted or migrated into cosmos this wave, but its import of
`CARD_SURFACE` is re-pointed at `@numisma/components` like every other, because
the file it imported from no longer exists. Leaving a dangling relative import
is not "out of scope".

## The fixtures

The coverage guard demands a named import from `"@numisma/components"` for
**both** `Card` and `CardTitle`. `Card.Title` usage does not satisfy it: the
guard matches the name inside an import brace, and property access is invisible
to it. One fixture naming both in its import, or two fixtures, either is fine.

The fixture must show the heading at both levels the component supports, since
the level is an accessibility fact the call site decides and the component is now
the one place in the layer that can get it wrong for every card at once.

## Acceptance criteria

- `packages/components` exports `Card`, `CardTitle` and `CARD_SURFACE`, from a
  kebab-cased file under `ui/`. `Card.Title` is the same function as `CardTitle`
  and it is asserted.
- No file anywhere in `apps/web` imports `Card`, `CardTitle` or `CARD_SURFACE`
  from a relative path.
- `packages/components/src` contains zero occurrences of `var(--card)` and
  `var(--line)`, in code or in prose.
- The package declares fourteen tokens. The app's stylesheet declares fourteen
  `--nms-*` aliases. `APP_TOKENS` and `THEMED_TOKENS` each carry fourteen, and
  the fourteen themed values are pairwise distinct.
- Workbench fixtures import both `Card` and `CardTitle` by name from
  `"@numisma/components"`, and the fixture-coverage guard demands both.
- In app mode the card paints `#181b22` behind a `#262a33` hairline, which is
  what `apps/web` paints today. In themed mode both change.
- Full suite green.

## Out of scope

The two snapshot notices, `NOTICE_CODE` and `Crumb` stay where they are. No
token other than `--nms-card` is minted. No `variant` enum, no polymorphic `as`
prop, no context provider: the parts share nothing at runtime and the open
composition shape is settled. `rounded-xl` and `p-4` do not move to another
scale. The card composition contract test stays in `apps/web`. No new test file.
Behaviour-preserving: a visible difference is a defect. Everything in spec #432
§2.2 stays out, and the ladder fixture route keeps its four states and its
DCE assertion.

### LINKED ISSUE #436: Slice 3 — moves the two snapshot notices into the package and mints --nms-neg
Spec: #432 (wave 1 of three, the generics migration).
Blocked by: #433, #435.

The two snapshot guard notices cross the boundary, and this is the slice that
proves `--nms-neg` renders. It also pins the wave's one deliberate duplication:
`--nms-neg` is minted as a name distinct from `--nms-destructive` even though
both resolve to the same colour in the app today.

**The block on #435 is inherent, not an ordering preference.** The notices
compose the card surface constant. Once they are package files they must import
it from within the package, and a package file cannot reach into `apps/web`.
The surface has to arrive first.

## The per-slice gate (spec §3)

1. **The read rewrite is enumerated, not grepped.** The table below is the
   contract. A count that does not match it fails the slice.
2. **App mode renders both notices identically.** The empty notice paints the
   card surface; the stale notice paints the surface plus the negative colour on
   the wrapper and the heading.
3. **Themed mode repaints every read.** The stale notice must change colour in
   themed mode. If it does not, the fixture is not exercising the token.
4. **`fixture-coverage.test.ts` goes from silent to demanding** on both notices
   the moment they are exported.

## The colour reads

The file performs **exactly two** colour reads, both of `--neg`, both in the
stale notice. The empty notice and the inline-code constant read no colour.

| # | where | today | becomes |
|---|---|---|---|
| 1 | the stale notice's wrapper `<div>` class attribute, after the composed card surface and `m-0` | `text-[var(--neg)]` | `text-[var(--nms-neg)]` |
| 2 | the stale notice's `<h1>` class attribute | `text-[var(--neg)]` | `text-[var(--nms-neg)]` |

**Two, and the count is the failing condition.** This file's counts in spec
§4.1 and §5 are exact: there are no prose occurrences of `var()` in it, unlike
the two files before it.

The heading's read is redundant by inheritance from the wrapper and is written
anyway. That is deliberate and stays: a converted element relying on its
parent's colour reads as an omission the next time someone recolours the box.
Deleting it during the move would be a behaviour-preserving change that is not
behaviour-preserving in review terms, and it is out of scope here.

The inline-code constant keeps `rounded-[6px]` and `bg-black` exactly as
written. `rounded-md` compiles, resolves and paints 8px because the app's
`@theme` remaps the medium radius onto the package's control radius, and
`bg-black` is a Tailwind default rather than a mapped house token, so it is not
the same collision.

## The mint, which is a three-sided edit

`--nms-neg` lands in all three places or the guards red:

- **the package token spec**: the name, a grayscale default, and a note that
  carries two facts. First, that this is the **sign of a number**, data, and is
  deliberately not `--nms-destructive`, which is the affordance of a button that
  destroys something. Welding them means the day the money red wants to soften,
  or wants a colourblind-safe pairing with a positive counterpart, you cannot
  move one without moving the other, and by then eleven components read it.
  Second, that the grayscale default makes sign **unreviewable in grayscale
  mode**, which is correct: grayscale reviews hierarchy, spacing and state, and
  sign is reviewed in themed and app mode. That second sentence pre-answers the
  same question for the positive counterpart in wave 2.
- **the app stylesheet**: the alias `--nms-neg: var(--neg)`. It sits alongside
  the existing `--nms-destructive: var(--neg)`. Two aliases pointing at one app
  token is the intended end state, and the app palette token itself is still
  declared exactly once, which the stylesheet's additive-edit guard checks.
- **the workbench theme-mode data**: `APP_TOKENS` gains the resolved literal
  `#f0736a`. This will **equal** the app value already carried for
  `--nms-destructive`, and no guard objects: only `THEMED_TOKENS` asserts
  pairwise distinctness. `THEMED_TOKENS` gains a loud value that must differ
  from the existing destructive themed value, so the two roles separate visibly
  in the one mode built to separate them.

## Guard edits this slice owes

- **The app's `nms-tokens` test floor** rises from fourteen to fifteen, its
  final wave-1 value.
- **Its alias-form table** gains `--nms-neg` paired with `var(--neg)`.
- Nothing else. The closure allow-list is already widened, and the app-side
  colour utility guard needs no edit: it requires more than ten non-test `.tsx`
  files under the app's source tree to be a meaningful scan, nineteen exist
  today and fifteen survive the whole wave. Wave 2 takes it to eight and the
  guard has to be re-homed then, not now.

## What moves, and every consumer

`SnapshotNotice.tsx` becomes `snapshot-notice.tsx` under the package's `ui/`
directory, imports written extensionless, and its import of the card surface
becomes an intra-package relative import. `SnapshotEmptyNotice`,
`SnapshotStaleNotice` and `NOTICE_CODE` join the package's named public surface.

Both notices keep the element they render. They are `<div>`s on the card
surface, not `Card`s: converting them would change the landmark structure of
three pages, which is a visible and accessibility change and belongs to no wave.
There is no generic `Notice` with a severity prop, for the reason the file
already records: it would have exactly these two consumers.

The control flow stays in the routes. Each route keeps its own status branch and
its own shell. These components are the words, not the guard.

**Importers re-pointed:**

- the glance route and the big-picture route, each importing both notices
- the ladder route, importing both notices
- the fill path and its cards structure test, importing `NOTICE_CODE`
- the ladder fixture route, importing `NOTICE_CODE`
- the snapshot notice copy test, importing both notices and the card surface

**The ladder fixture route is again one of them**, and again spec §2.2 lists
that route as out of scope for the wave. The route is not moved or deleted; its
import specifier is re-pointed because the file it imported from no longer
exists.

## The fixtures

Two fixtures, or one naming both in its import brace. The coverage guard demands
a named import from `"@numisma/components"` for each of `SnapshotEmptyNotice`
and `SnapshotStaleNotice`.

The stale notice's fixture must pass a stored version and a supported minimum
and maximum, because all three are rendered: refusing to render without the
numbers leaves the operator with nothing to act on.

## Acceptance criteria

- `packages/components` exports `SnapshotEmptyNotice`, `SnapshotStaleNotice` and
  `NOTICE_CODE`, from a kebab-cased file under `ui/`.
- No file anywhere in `apps/web` imports either notice or `NOTICE_CODE` from a
  relative path.
- `packages/components/src` contains zero occurrences of `var(--neg)`.
- The package token spec declares `--nms-neg` and `--nms-destructive` as two
  names with two notes, and the `--nms-neg` note states both the data-versus-
  intent distinction and the grayscale consequence.
- The package declares fifteen tokens. The app's stylesheet declares fifteen
  `--nms-*` aliases. `APP_TOKENS` and `THEMED_TOKENS` each carry fifteen, and
  the fifteen themed values are pairwise distinct, including `--nms-neg` against
  `--nms-destructive`.
- Workbench fixtures import both notices by name from `"@numisma/components"`.
- In app mode the stale notice paints `#f0736a` on both reads, which is what
  `apps/web` paints today. In themed mode both change, to a value different from
  the destructive role's.
- Both notices still render the elements they render today, with the same copy
  and the same heading levels.
- Full suite green.

## Out of scope

`Crumb` stays where it is. No token other than `--nms-neg` is minted. The
notices do not become `Card`s, do not merge into a parameterised `Notice`, and
do not absorb the ladder route's own not-found or no-price-axis notices, which
have different copy, one call site each, and a different class set. The status
branching stays in the three routes. The redundant heading colour stays. The
copy contract test stays in `apps/web`. No new test file. Behaviour-preserving.
Everything in spec #432 §2.2 stays out.

### LINKED ISSUE #437: Slice 4 — moves Crumb into the package behind a renderLink slot
Spec: #432 (wave 1 of three, the generics migration).
Blocked by: #433, #434.

The last generic crosses the boundary, and it is the only one whose API changes.
`Crumb` currently imports TanStack Router's `Link` and takes a `to` string. The
package cannot import a router, so the link leaves through a slot.

It gets its own reviewable slice because it is the one place in the wave a
reviewer must check a **type** rather than a colour. The other three slices are
the same move run three times; this one changes the shape of five call sites and
a test helper.

**The block on #434 is a token dependency, not an ordering preference.** `Crumb`
reads the secondary-text grey, which #434 mints. It mints nothing itself.

## The per-slice gate (spec §3)

1. **The read rewrite is enumerated, not grepped.** The table below is the
   contract. A count that does not match it fails the slice.
2. **App mode renders the crumb identically**, resting and hovered. The hover
   state is half of this component and a fixture that cannot be hovered is not
   showing you read 2.
3. **Themed mode repaints every read.** Both the resting colour and the hover
   colour must visibly change.
4. **`fixture-coverage.test.ts` goes from silent to demanding** on `Crumb` the
   moment it is exported.

## The colour reads

`Crumb` performs **exactly two** colour reads, both in the anchor's class
attribute, and both names already exist in the package after #434.

| # | where | today | becomes | status |
|---|---|---|---|---|
| 1 | the anchor's class attribute, resting colour | `text-[var(--muted)]` | `text-[var(--nms-muted-foreground)]` | minted in #434 |
| 2 | the same class attribute, hover colour | `hover:text-[var(--text)]` | `hover:text-[var(--nms-foreground)]` | name exists |

**Two, and the count is the failing condition.** Spec §5 says `--muted` once and
`--text` once, which is exact. There are no prose occurrences of `var()` in this
file, so no docblock reword is needed here.

The wrapping `<p>`'s `m-0` and its size utility are untouched. `m-0` is not
decoration: preflight is off, the user agent's paragraph margin is live, and
dropping it pushes every crumb down the page with nothing in the diff to say
why.

**No mint.** The package still declares fifteen tokens after this slice, the
same fifteen #436 left.

## The API change

`to` leaves the package entirely. The component takes a render slot:

```ts
renderLink: (props: { className: string; children: ReactNode }) => ReactElement;
```

The package owns the anchor's classes and hands them to the slot. It does not
own the anchor.

**Why a slot and not a link-adapter context.** TanStack checks `to` against the
generated route tree. An adapter that normalised every destination to
`to: string` would throw that check away silently, and the first typo in a route
path would ship. The slot keeps each call site building a fully typed `<Link>`
with its own literal destination, so the route-tree check still runs where the
destination is written.

**Why the package cannot just import the router.** Routerlessness is what makes
the workbench a diagnostic instrument rather than a second view of the app. The
workbench's vite config states the absence of a router and of SSR is the point
of the second consumer, so `@tanstack/react-router` is not available to the
package and must not appear in its manifest.

The arrow glyph stays the caller's. Two of the five usages point in opposite
directions, and a component deciding the glyph from the destination would be
guessing at page geography from a string.

## What moves, and every consumer

`Crumb.tsx` becomes `crumb.tsx` under the package's `ui/` directory, imports
written extensionless, and the router import is deleted rather than re-pointed.
`Crumb` joins the package's named public surface, bringing it to six components
and two data exports.

**Five usages across three route files.** The glance route has one pointing down
to the big picture; the big-picture route has one pointing back to the glance;
the ladder route has three, all pointing back to the glance, in three different
status branches. Each builds its own `<Link>` inside `renderLink`, passing the
`className` the slot receives straight through and keeping the destination it
renders today. Nothing on any page moves.

**The crumb contract test's helper changes shape.** It currently mounts a crumb
from a destination and a label. It now supplies a `renderLink` and asserts
through it.

## The fixture

One cosmos fixture importing `Crumb` by name from `"@numisma/components"`,
supplying `<a href="#">` as its `renderLink`.

The stub is correct, not a compromise. A fixture whose link navigates is a
fixture that has just left the state you were looking at. The fixture must also
not import anything from `apps/web`, which the workbench's seam-isolation guard
enforces on every specifier.

## The new test case

One case in the crumb contract test, and it is the only new test the wave adds.
It proves the anchor renders **through the slot**, carrying the classes the
package hands it, with no router present. Spec §6's rule that a slice wanting a
new test file has changed behaviour and should stop still holds: this is a case
in an existing file, and the behaviour it covers is the API change this slice
exists to make.

## Acceptance criteria

- `packages/components` exports `Crumb`, from a kebab-cased file under `ui/`,
  and neither that file nor the package manifest mentions
  `@tanstack/react-router`.
- `Crumb`'s props are `renderLink` and `children`. No `to`.
- `packages/components/src` contains zero occurrences of `var(--muted)`,
  `var(--card)`, `var(--line)`, `var(--neg)` and `var(--text)`, in code or in
  prose. This is spec §8's signal, and this slice is where it comes true.
- No file anywhere in `apps/web` imports `Crumb` from a relative path. Five
  usages across three routes build a typed `<Link>` inside `renderLink`, each
  keeping the destination and the arrow it renders today.
- `apps/web/src/components/ui/` holds only the four contract tests.
- `packages/components` exports eight names: six components and two data
  exports. Six workbench fixtures exist, one per component, each importing its
  component by name from `"@numisma/components"`, and the workbench renders all
  six in all three modes.
- A crumb contract case proves the anchor renders through the slot with the
  package's classes and no router.
- The ladder route's import-closure guard allows exactly three runtime
  `@numisma/*` specifiers.
- In app mode the crumb rests at `#9aa1ad` and hovers to `#e7e9ee`, which is
  what `apps/web` paints today. In themed mode both change.
- Full suite green.

## Out of scope

No token minted. No link-adapter context. The arrow glyphs, the destinations,
the heading of any page and the crumb's own size and margin do not change. The
component keeps its wrapping `<p>`. `params` and `search` do not get slots of
their own: the slot hands out a class name and children, and everything about
navigation stays at the call site. The crumb contract test stays in `apps/web`.
Behaviour-preserving: a visible difference is a defect.

Everything in spec #432 §2.2 stays out and stays out for wave 1's whole life:
the seven domain components, the `glance/` and `ladder/` view modules, the
`/ladder-fixture/$state` route, the four wave-2 token names, any visible change,
the login accessibility fix, and moving the four contract tests into the
package. Wave 2 opens against a written ten-name vocabulary and a proven move
recipe, which is what this wave was for.

AmetAlvirde:
Wave 1 of three. Four generic components leave `apps/web/src/components/ui/` for
`packages/components/src/ui/`, three `--nms-*` names are minted, and `Crumb`
gives up its router. The workbench can now see six components instead of one.

Behaviour-preserving throughout. A visible difference is a defect, not polish;
polish is wave 3 and does not share this PR.

## The slices

| commit | slice | what |
|---|---|---|
| `6337ef8` | S0 | the ten-name vocabulary into `tokens.ts`'s header, the two source conventions into `docs/component-package.md`, the narration correction at three sites |
| `f513b0d` | S1 | `Absent`, and the `--nms-muted-foreground` mint |
| `421594b` | S2 | `Card`, `CardTitle`, `CARD_SURFACE`, and the `--nms-card` mint |
| `90a5b2a` | S3 | both snapshot notices, `NOTICE_CODE`, and the `--nms-neg` mint |
| `e134217` | S4 | `Crumb` behind a `renderLink` slot, and the wave's one new test case |
| `924e645` | — | docs reconciled against what the five slices actually landed |

## The trap this wave was shaped around

In the app's vocabulary `--muted` is secondary text. In the package's,
`--nms-muted` is the recessed surface. A mechanical `--muted` → `--nms-muted`
rewrite compiles, emits a rule, wins its cascade, repaints the app's most-used
grey, and leaves every guard in the repo green — because the file has just left
the tree the app-side colour guard scans, and the package-side guard only checks
that a read sits inside the namespace, never that it is the right name in it.

The defence was enumeration, not grep: every slice brief listed its colour reads
by what they sit in, and a count that did not match failed the slice. Every
count matched. `Absent` 2, `Card` 2, the notices 2, `Crumb` 2.

## Where the briefs were wrong, and what happened

**The mint is a four-sided edit, not three.** Every brief describes it as
landing in `tokens.ts`, `styles.css` and `theme-modes.ts`. But
`nms-tokens.test.ts` iterates `NMS_TOKEN_NAMES` demanding a Tailwind `@theme`
mapping, so `apps/web/src/tailwind.css` is a fourth side. S1 found it by going
red; S2, S3 and S4 were briefed with it.

**S0 could not ship as the spec wrote it.** Spec §5 proposed minting all three
names while no component moved. `tokens.test.ts` fails any declared token no
component reads, which is the same rule §4.2 quotes when it declines to mint all
ten. Ruled in #433: each name is minted in the slice whose component first reads
it. The cost, accepted, is that the four component slices serialise.

**Prose is scanned.** `nms-namespace.test.ts` reads raw file text with no
comment stripping, so a docblock quoting a deleted CSS rule verbatim reds the
guard the moment its file lands in the package. Namespacing the quote would file
a false record of a rule that never existed, so `absent.tsx` and `card.tsx` had
their quoted declarations reworded to stop spelling `var()`, keeping the
historical fact. This is also the only way spec §8's zero-occurrence signal is
satisfiable at all.

## Spec §8, checked rather than claimed

- `@numisma/components` exports eight names; six components have fixtures; the
  workbench renders all six in three modes.
- Zero occurrences of `var(--card)`, `var(--line)`, `var(--muted)`, `var(--neg)`
  and `var(--text)` in `packages/components/src`, code and prose.
- Zero relative imports of the four moved components anywhere in `apps/web`.
- `apps/web/src/components/ui/` holds only the four contract tests.
- Fifteen tokens declared, fifteen `--nms-*` aliases, fifteen in each mode
  table, fifteen pairwise-distinct themed values.
- Full suite green: 204 files, 3045 tests, 52 skipped. Typecheck clean.

## One thing a reviewer should look at that no slice owned

`apps/web/src/theme-color-utilities.test.ts`'s `THEME_COLOR_NAMES` is a literal
list of nine and gained neither `card` (S2) nor `neg` (S3). Both now exist as
Tailwind theme utilities in the app's build, so an app-side `bg-card` or
`text-neg` would compile, resolve through the package's namespace, and pass the
guard written to catch exactly that. Spec §4.3 predicted this cost as an
argument against the utility form; nobody was scoped to pay it. It is a hole
this wave widened rather than one it opened, and it wants a decision rather than
a quiet fix inside a behaviour-preserving PR.

Also left deliberately: `card.tsx:5,25` still narrates spec #420 in the present
tense. That is pre-existing drift from a different increment, untouched by these
six commits, and correcting it means re-deriving what #420 deleted with its diff
in hand.

## One deliberate repaint, found in the workbench

`NOTICE_CODE` set `bg-black` and inherited its foreground. That only ever
rendered correctly because the single consumer was a dark app whose ambient
foreground is light; against the workbench's other two palettes the chip painted
near-black on black. Every test stayed green throughout, because the class list
was pinned to exactly what the app shipped, and the app is where it worked.

`dcfe0ea` pairs the background with a literal `text-white`. Literal for the same
reason `bg-black` is: no house token stays light in every mode, so a `var()`
read would re-acquire the dependence on ambient palette that caused the bug. The
chip does not theme, by construction.

**This moves app-mode pixels**: the chip's text goes from `#e7e9ee` to
`#ffffff`. It is the one exception to the behaviour-preserving claim above,
stated here rather than absorbed quietly. Nothing else in the app changes.

Worth noting what caught it. The workbench fixture already gave the chip its own
row and already stated the rule — but stated it about the background, which is
the half that was never at risk. The mode the wave built is what surfaced it,
which is the argument for the wave.

## What this leaves for wave 2

A written ten-name vocabulary in the file that owns it, four names declared and
deliberately unminted, and a move recipe run four times. The seven domain
components, the `glance/` and `ladder/` view modules and `/ladder-fixture/$state`
all stay exactly where they are.

---

Closes #433
Closes #434
Closes #435
Closes #436
Closes #437
Closes #432


AmetAlvirde:
## Review: wave 1 of the generics migration

Verified locally against `origin/feature/generics-migration` (924e645) in a clean
worktree: `pnpm typecheck` clean, `pnpm test` 203 passed / 4 skipped files,
3043 passed / 54 skipped tests. I also built `@numisma/web` on this branch and on
`origin/main` and diffed the two emitted stylesheets rule by rule, which is where
finding 2 comes from.

**The behaviour-preserving claim holds.** Every rule in the built `tailwind.css`
that existed on `main` still exists on this branch with the same declarations;
the branch adds five rules and removes none. The four moved component bodies are
character-identical to their `main` versions apart from the colour reads, so the
elements, the copy, the heading levels, the em dash and the class strings are
unchanged. Each moved read resolves through the alias chain to the literal
`apps/web` painted before:

- `--nms-muted-foreground` → `--muted` → `#9aa1ad` (`absent.tsx` ×2, `crumb.tsx` resting)
- `--nms-foreground` → `--text` → `#e7e9ee` (`crumb.tsx` hover)
- `--nms-card` → `--card` → `#181b22`, `--nms-border` → `--line` → `#262a33` (`CARD_SURFACE`)
- `--nms-neg` → `--neg` → `#f0736a` (`snapshot-notice.tsx` ×2)

`--nms-muted` stays pointed at `--recess`, so the rename this wave was shaped
around did not happen. The counts check out as asserted: eight new exported
names, fifteen declared tokens, fifteen `--nms-*` aliases in `styles.css`,
fifteen `@theme` mappings, fifteen entries in `APP_TOKENS` and `THEMED_TOKENS`,
and zero `var(--card|line|muted|neg|text)` in `packages/components/src`.
The five `Crumb` call sites keep their destinations (`/` ×4, `/big-picture` ×1)
and their arrows, and the slot moved route typing from a bare `to: string` to a
literal `to` the route tree checks — a gain, not a loss. `Card.Title` is still
`CardTitle`, exercised in `card.fixture.tsx` and asserted in the contract test.

Five findings, none of which contradict that.

---

### 1. The closure guard now trusts a whole package that nothing pins

`apps/web/src/routes/route-move.test.ts:243`

`@numisma/components` joins the allow-list, and the walker's own docblock says
the walk "follows relative specifiers only — a package specifier is what is being
judged, not traversed". So the guard never looks inside the package. The safety
argument in the new comment rests entirely on the package declaring four runtime
dependencies, and no test in the repo asserts that set: `ops/components/guards.test.ts`
only rejects radix, and only on the `components:add` path, which is not the path
house components arrive by. Add `@numisma/engine` to
`packages/components/package.json` and re-export anything from it, and the
engine's `node:os`/`node:path` reach lands in the ladder route's browser bundle
with this test green and `client-bundle.integration.test.ts` still `skipIf(!hasBuild)`.

The two existing entries are safe by construction — leaf modules with no imports
— which is exactly what the third one is not. Either walk into the package's
manifest here, or add a standing assertion on its dependency list, so the prose
argument has something holding it.

### 2. The reworded docblock still compiles a house colour into the app's stylesheet

`packages/components/src/tokens.test.ts:42`

Both namespace guards skip `*.test.ts`. Tailwind's scanner does not:
`apps/web/src/tailwind.css:57` points `@source` at the whole package `src`, test
files included. The shorthand written in this docblock as an example is scanned
as a candidate and emitted as a real rule.

Diffing the built stylesheets:

```
main:   .bg-\(--muted\) { background-color: var(--muted) }
branch: .bg-\(--bg\)    { background-color: var(--bg)    }
```

Nothing in the app wears either class, so no pixel moves, and the file is not the
one spec §8's zero-occurrence signal names. But the substitution moved the
example from one house name to another rather than out of the house vocabulary,
and the result is a bare `var(--house-name)` declaration shipping in `apps/web`'s
stylesheet from package source — the failure this wave is written against,
surviving in the one file every guard excludes. `--nms-muted` or `--nms-bg` in
the example costs nothing and closes it. Worth knowing separately that the
guards' test-file exclusion is not shared by the Tailwind scanner.

### 3. `--nms-card` is still the example of a token nothing renders

`packages/components/src/tokens.test.ts:107`

> A token pasted in from an upstream theme — `--nms-card`, `--nms-popover` — is a
> name every consumer is asked to define and nothing ever renders

`--nms-card` is now declared, aliased, mirrored in both mode tables and read by
`CARD_SURFACE`. S2's message says "the two docblocks that staged `--nms-card` as
pending lose that clause"; this is a third one that kept it, and it is the one
sitting directly above the assertion that would fail if the sentence were true.
`--nms-popover` alone still makes the point.

### 4. Duplicate import statements from the same specifier

`apps/web/src/components/SectionTable.tsx:1-2`, `SummaryCard.tsx:3-4`,
`DcaCard.tsx:3,5`, `GlanceCard.tsx:1,11`, `apps/web/src/routes/login.tsx:4,7`

Five files import from `"@numisma/components"` twice, two of them on adjacent
lines. The relative paths they replaced were genuinely different modules; the
package specifier is one module, so this is a merge artifact of the mechanical
re-point rather than a choice. One import each.

### 5. Docblock reflow artifact

`packages/components/src/ui/card.tsx:60-62`

The `var()`-removing rewrite left the sentence wrapping mid-phrase:
"…exported because EIGHT ELEMENTS CARRY / THE SURFACE AND THREE OF THEM / ARE NOT
CARDS". Cosmetic. (The count disagreeing with the file's own opening paragraph —
"a dozen elements … four of them are not cards" — is pre-existing on `main` and
not this PR's to answer.)

---

On the two disclosed items: I re-derived both and neither is worse than stated.
`THEME_COLOR_NAMES`'s regex appends an optional `-foreground`, so `muted` already
covers `text-muted-foreground` and the hole is exactly `card` and `neg` as
described; no `.tsx` in `apps/web`, `apps/workbench` or `packages/components`
writes either utility today, so it is latent. `card.tsx:5,25` is untouched by
these six commits.

One number to correct in the body: I measured 207 test files (203 passed,
4 skipped) and 3097 tests (3043 passed, 54 skipped), not 204 / 3045 / 52.

AmetAlvirde:
All five findings are fixed in `42fc29f`.

1. `route-move.test.ts` — the allow-list docblock argued `@numisma/components` was safe because of what it depends on. That argument is now an assertion: a new case pins the whole key set of both `dependencies` and `peerDependencies`, so an addition reds the guard instead of slipping past a walker that only follows relative specifiers. Peers are covered too, since that is where the same dependency would go to dodge a `dependencies`-only check.
2. `tokens.test.ts:42` — the shorthand examples now carry a space inside the parens (`bg-( --nms-bg )`), which stops Tailwind's candidate scan while leaving the syntax readable, and the `var()` example moved onto an `--nms-` name. Rotating which house colour the example named is what the previous two slices each did, and it moved the problem rather than removing it. Verified against a real build: the emitted stylesheet now carries no `.bg-\(…)` rule at all. A paragraph in the docblock says why the spaces are load-bearing, so nobody closes them back up.
3. `tokens.test.ts:107` — the example of a token nothing renders moved off `--nms-card`, which this PR declares, onto `--nms-popover`.
4. The five duplicate `@numisma/components` imports are merged.
5. `card.tsx` docblock reflowed.

`pnpm test` 3046 passed / 52 skipped, typecheck clean. The only stylesheet delta from these fixes is the stray rule finding 2 removed; no component output changed.

On the count discrepancy: the review measured 203/3043/54 against the body's 204/3045/52. That is the DB-gated integration file skipping without `NUMISMA_TEST_DATABASE_URL` set, not a difference in the work.

